PR #3, opened 2017-08-06
enable tls and escape paths for powershell

musm:


musm:
Closing since I don't want to go fixing the errors right now even though they are basic, feel free to use this to fix underlying issue

musm:
cc @staticfloat  heads up

PR #11, opened 2017-10-22
Start working on Wizard

Keno:
Not nearly complete, but something to start from.

Demo: https://asciinema.org/a/7gDelUUX0wYtmgO6V2Behkmec


src/Wizard.jl
```diff
@@ -0,0 +1,207 @@
...
+
+    println()
+
+    build_path = tempname()
+    mkpath(build_path)
+    local history
+    cd(build_path) do
+        temp_prefix() do prefix
+            histfile = joinpath(build_path, ".bash_history")
+            dr = DockerRunner(prefix = prefix, platform = :linux64,
```
ararslan: Platform arguments should now be using `Platform` objects. In this case it should be `platform = Linux(:x86_64)`.

Keno: True, there's a bunch of other code like this in this package also.

ararslan: Ah yeah you're right. This is Binary**Builder**, not Binary**Provider**. Perhaps I'll go through and update things here.

Keno: I already have those changes locally, so if you haven't yet, don't start. Otherwise feel free to put up a PR.

ararslan: Okay, I just did. Didn't see your message until just now.

src/Wizard.jl
```diff
@@ -0,0 +1,371 @@
...
+            print_with_color(:bold, "\n\t\t\tBuild complete. Analyzing...\n\n")
+
+            audit(prefix; platform=Windows(:x86_64), verbose=true, autofix=false)
+
+            match_files(prefix, Windows(:x86_64), state.files)
+        end
+    end
+
+    println("")
+    println("You have successfully built for Win64. Congratualations!")
```
ararslan: > Congratulations

You've got an excess a in there

vchuravy:
While playing around with this branch I encountered the following issues.

- `tar.bz2` support missing
-  `deps/Makefile` isn't run automatically
-  `deps/sandbox_root` is missing on a fresh checkout
- Can't write to `/tmp`
-  `bash: /root/.bashrc: Permission denied` when starting the wizard

Keno:
FWIW, https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/34 should address the last of @vchuravy issues (the sandbox_root one).

PR #19, opened 2017-10-22
Use `Platform`s instead of `Symbol`s to represent platforms

ararslan:
Since BinaryProvider recently made the same change.

### DIFF
--- a/src/BinaryBuilder.jl
+++ b/src/BinaryBuilder.jl
@@ -10,4 +10,24 @@ include("Auditor.jl")
 include("DockerRunner.jl")
 include("Dependency.jl")
 
+export shortname
+
+"""
+    shortname(p::Platform)
+
+Get a convenient symbol representation of the given platform.
+
+# Examples
+```jldoctest
+julia> shortname(Linux(:i686))
+:linux32
+
+julia> shortname(MacOS())
+:osx64
+```
+"""
+shortname(p::Linux) = Symbol("linux", wordsize(p))
+shortname(p::Windows) = Symbol("win", wordsize(p))
+shortname(p::MacOS) = :osx64
+
 end # module
--- a/src/BinaryBuilder.jl
+++ b/src/BinaryBuilder.jl
@@ -26,6 +26,14 @@ julia> shortname(MacOS())
 :osx64
 ```
 """
+function shortname(p::Linux)
+    a = arch(p)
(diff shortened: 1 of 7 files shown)


src/BinaryBuilder.jl
```diff
@@ -10,4 +10,24 @@ include("Auditor.jl")
...
+# Examples
+```jldoctest
+julia> shortname(Linux(:i686))
+:linux32
+
+julia> shortname(MacOS())
+:osx64
+```
+"""
+shortname(p::Linux) = Symbol("linux", wordsize(p))
```
staticfloat: I don’t think this will work for aarch64 or ppc64le, right? We need to encode the arch or something.

src/DockerRunner.jl
```diff
@@ -80,11 +80,15 @@ within the crossbuild environment.
...
+    platform::Platform
 end
 
 function show(io::IO, x::DockerRunner)
-    write(io, "$(x.platform) DockerRunner")
+    p = x.platform
+    # Displays as, e.g., Linux x86_64 (glibc) DockerRunner
+    write(io, typeof(p), " ", arch(p), " ",
+          Compat.Sys.islinux(p) ? "($(p.libc)) " : "",
+          "DockerRunner")
```
staticfloat: Nice.

src/BinaryBuilder.jl
```diff
@@ -10,4 +10,24 @@ include("Auditor.jl")
...
+Get a convenient symbol representation of the given platform.
+
+# Examples
+```jldoctest
+julia> shortname(Linux(:i686))
+:linux32
+
+julia> shortname(MacOS())
+:osx64
+```
```
staticfloat: I like doc tests. I forgot they exist. How do I run the doctests?

src/BinaryBuilder.jl
ararslan: Oh yes you're right

src/BinaryBuilder.jl
ararslan: Well, first you need Documenter. Is that set up here?

ararslan:
There's something up with Windows. Any idea whether that's related?

staticfloat:
Not related; Appveyor hasn't ever worked, unfortunately

PR #45, opened 2017-11-09
Provide hints for CMake

Keno:
Depends on https://github.com/staticfloat/julia-docker/pull/6 for correct operation.


staticfloat:
Now that #46 is merged, this looks good to go.  Nice work.

PR #57, opened 2017-11-25
Use squashfs on Travis

### LINKED ISSUE #42: Don't download stuff in __init__()
For example, the VS Code extension will load the package if another package just has the line ``using BinaryProvider`` in any source file, and that certainly shouldn't trigger a large download.

Shouldn't the download just happen in the ``deps/build.jl`` file?

### LINKED ISSUE #55: Look into distributing rootfs as a squashfs image
https://en.wikipedia.org/wiki/SquashFS

Would help with #54 by getting rid of one copy of the data and would let us save some time on every build by allowing is to forgo the decompression step.

Keno:
This fixes #55 and should hopefully resolve all the instances of travis
running out of disk space when using BinaryBuilder. While we're at it,
move all the downloading to deps/build.jl, which fixes #42.

Depends on https://github.com/staticfloat/julia-docker/pull/8


staticfloat:
I'm going to integrate my sharding code into this, initially by just downloading all shards at the beginning, but from here we can move toward downloading shards on-demand, or even downloading new shards in the background while building with the shards we have, for even faster builds.

staticfloat:
Superseded by #59

PR #58, opened 2017-11-26
Sharding the build image

staticfloat:
This augments the `kf/squashfs` branch to include build image sharding.  The basic approach is:

* Use docker to build the whole shebang, as usual
* Package that giant build tree up into multiple `.tar.gz` and `.squashfs` files
* Download those `.tar.gz` and `.squashfs` files on-demand on the builder's machine as they need them.

Implementation-wise, this is done by just downloading each tarball and extracting it into a giant rootfs directory.  This rootfs is then mounted in, just like old times.

This also incorporates an easy way to download Apple's macOS SDK, prompting the user to read the macOS and Xcode SDK license terms.  A `.tar.xz` file is then downloaded from the github releases of [this repository](https://github.com/phracker/MacOSX-SDKs) and extracted into the rootfs directory.

This doesn't work for `squashfs` images though; although I could issue multiple `mount` commands to build up the rootfs directory, I can't unpack the macOS SDK file into these directories because the filesystems are read-only.  (Actually, I just tried it, and I can't even do the "multiple mounts" thing because I can't create the mountpoints within the `/opt` folder of the base image)

I think the simplest way to do this is to not try to assemble the whole image as directories on the outside of the build image and them mount them in, but to instead have a way to provide "volume mappings" to `sandbox` a la Docker: `-v <path_to_directory_or_squashfs>:<path_within_sandbox_to_mount_at>`.  I realize we can't directly mount `.squashfs` files within `sandbox` yet, but someday hopefully we can.  Until them, being able to provide flexible directory mappings would be really helpful, as I could them pass something like `sandbox -v ./shards/base:/ -v ./shards/x86_64-apple-darwin14:/opt/x86_64-apple-darwin14 -v ./shards/MacOSX10.10.sdk:/opt/x86_64-appledarwin14/MacOSX10.10.sdk`.


Keno:
Yeah, we can mount all the squashfs images outside and then perform bind mounts inside the sandbox before we chroot. I'll whip that up tomorrow. What does sharding to to total file size? I imagine there's quire a bit of duplication between the various gcc toochains, which squashfs would get rid of in the same image.

staticfloat:
I have previously attempted to combine GCC toolchains without much luck; even trying to use a "super" bindeps to build a cross compiler toolchain doesn't work, unfortunately.

A single squashfs file containing the whole build image is smaller, but not by much: it's `2468024320` split across multiple files versus `2391367680` files, so we're losing 73 MB.  `shasum`'ing all the libraries in the image, there are no libraries that are _precisely_ identical across shards, there are of course copies of the same library within a single shard, but I don't see any big easy savings across the entire thing.

staticfloat:
I'm moving this to a new PR that still incorporates your work in the other PR, but rebases this all on top of the changes to the wizard.

PR #60, opened 2017-11-27
Rework sandbox for a shattered world

Keno:
1. Simplify the way the overlay setup works. Instead of requiring the outside
   world to setup the upper/workdir, we now do so internally by mounting a new
   tempfs on /tmp and creating the directories ourselves.

2. Add --map option to the sandbox to allow mapping arbitrary outside directories
   into the sandbox.


Keno:
@staticfloat the squashfs version fails because apparently the travis user's uid is 2000 not 1000, so we'll have to update the packing script. Once you do that, this should pass.

staticfloat:
@Keno are you saying that `squashfs` will only work if the UID of the currently running user matches the UID of the files within the image?  That's a huge limitation, is it not?  There's no guarantee that Travis' UID will remain constant.  Why would that be a problem in any case?

Keno:
Yes, that's what I'm saying. I'm fairly certain travis' UID won't change. If we ever get mounting inside the sandbox it's a non-issue, but for now there's nothing we can do.

src/RootFS.jl
```diff
@@ -124,15 +179,33 @@ function update_rootfs(triplets::Vector{S}; automatic::Bool = automatic_apple,
...
+            
+            # If squashfs, verify/redownload the squashfs image
+            if !(file_existed && download_verify(
                 url,
                 hash,
                 squashfs_path;
                 verbose = verbose,
-            )
-
+                force = true
```
staticfloat: To answer your question on slack, yes, we do want `force = true`

src/RootFS.jl
```diff
@@ -124,15 +179,33 @@ function update_rootfs(triplets::Vector{S}; automatic::Bool = automatic_apple,
...
+            ))
+                # Unmount the mountpoint. It may point to a previous version
+                # of the file. Also, we're about to mess with it
+                if success(`mountpoint $(dest_dir)`)
```
staticfloat: We should probably unmount it before running `download_verify`, right?  Since we may attempt to overwrite the file.

Keno: Overwriting the file is fine, the kernel keeps a reference to it however, so the mountpoint would reflect the old version.

PR #61, opened 2017-11-28
Bug fixes

staticfloat:
Quash various bugs I've encountered during testing.

@Keno somewhere in the past few days, the overlayfs stopped working.  :(  If I run `touch /opt/<target>/blah` inside the sandbox, that file appears on the outside of the sandbox in my filesystem.  I tried switching the ordering of how we're mounting, thinking perhaps we need to mount the `overlayfs` _after_ we do all our bindmounts, but that obviously doesn't work since the `overlayfs` contains the parents of our mountpoints.

### DIFF
--- a/deps/.gitignore
+++ b/deps/.gitignore
@@ -1,6 +1,7 @@
 # Ignore rootfs artifacts
 root/
 downloads/
+shards/
 
 # Ignore sandbox executable
 sandbox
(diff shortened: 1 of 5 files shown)


Keno:
Right so anything we mount in now is not part of the overlay. We should me it read only to avoid modification, but is there any reason we need to modify the toolchain charts?

staticfloat:
Nope, but I thought the whole reason to use the overlayfs was to provide a plausible world for buildsystens that do “weird things” so that if they try to modify the world they can, but the modifications disappear later?

Keno:
Well modifying the toolchain shards is weirder than what I had in mind (writing to /tmp or /root or even trying to install into /). We should still prevent modifications of course as I said.

Keno:
I'll put up a PR later to make sure to mount those things ro.

PR #69, opened 2017-11-28
Add the ability to import binary dependencies

### LINKED ISSUE #6: Ensure that dependent packages can be built and then installed
From slack discussion:

> **staticfloat**: tshort, in regards to your concerns about lacking libraries/headers; my plan would be to have, as part of the build process for `libbar`, (which depends on `libfoo`), it would download and install the tarball for `libfoo` and put that into the same `Prefix` that we will install `libbar` into.  In effect, acting similarly to `/usr/local`.  `libbar` would then install into that prefix happily enough, `libfoo` can be removed from the `Prefix`, and we just package up the `Prefix` as the installed version of `libbar`.  Then, when the user wants to install `libbar`, we force `libfoo` to be installed first, and everything should "just work".  So in essence, anything that you want to build would require you to first install the tarballs for your dependencies, so that you have the libraries/headers available, but that doesn't seem unreasonable.
> **tshort**: I'm confused on the interplay between Pkg3 and BinDeps2. If I have a library to compile that wants to link to libraries provided by `libopenblas.jl`, how can I compile with BinDeps2? It doesn't have a Julia installation. And, who's in charge of versioning there, Pkg3 or BinDeps2?      ---    Wait a minute, I see you've answered my question before I even asked it :smile_cat: .
> **tshort**: Overall, that makes sense. Thanks, Elliot. But, that capability isn't in BinDeps2, yet, right?
> **staticfloat**: It's close; the big pieces are there, (e.g. installing something to a prefix, removing it again, blah blah blah) I just need to write up some examples to see how it all flows together and to reduce pain points.

Test this with a package that needs to build against, for instance, `libopenblas`, and ensure that it can properly find headers/link against it, and ensure that the libraries that link against it are properly relocatable, etc...

Keno:
Demo: https://asciinema.org/a/1NkfandjvQJYxCbSGxcX1aEFB

Provides the basic functionality as discussed. I'm sure we'll have to evolve it
as we move along, but this seems to work fine.

Fixes #6


staticfloat:
This is going to need a rebase after my merge this morning.  Thanks!

staticfloat:
I'm done with my `ar` adventures, and `OpenBLAS` is building (at least locally for me, travis [is crunching](https://travis-ci.org/staticfloat/OpenBLASBuilder/builds/310392697), hooray for shards!) so I'm going to rebase this and massage it.

staticfloat:
> Demo: https://asciinema.org/a/1NkfandjvQJYxCbSGxcX1aEFB

`make -j40`  😂

src/Auditor.jl
```diff
@@ -214,15 +214,13 @@ end
 Find all files that satisfy `predicate()` when the full path to that file is
 passed in, returning the list of file paths.
 """
-function collect_files(prefix::Prefix, predicate::Function = f -> true)
+function collect_files(prefix::Prefix, predicate::Function = f -> true; exculuded_files=Set{String}())
```
vchuravy: `exculuded_files`xD

PR #78, opened 2017-12-03
Support for qemu-based runner

Keno:
Tested and works on OS X. Steps left for seamless operation:

1. Bake the sandbox executable into the base squashfs image for use as init
2. Set up a builder to build a linux kernel according to our .config
3. Set up a builder for qemu.

The same code should work for windows support as soon as qemu gains a 9p driver
for windows (there's one in the works on the mailing list). On that front, we'll run into
trouble because of symlinks though. We may want to consider some alternate way of
extracting files from the VM (e.g. we could produce the tar files inside and then take
the tar file as input for the auditor).


src/QemuRunner.jl
```diff
@@ -0,0 +1,170 @@
...
+          "QemuRunner")
+end
+
+function generate_cmd(ur::QemuRunner, cmd)
+    cmd = `$(ur.sandbox_cmd) $(cmd)`
+    cmd_string = join(map(cmd.exec) do arg
+        # Could encode as human readable with shell escaping, but we'd just
+        # have to do the decoding on the client side and it's a very, very
+        # complicated encoding. Base64 is much simpler.
+        base64encode(arg)
```
staticfloat: This makes me a little sad that it has to be so complex, but I understand why you're doing it.

src/QemuRunner.jl
```diff
@@ -0,0 +1,170 @@
...
+    end, ' ')
+    env_string = join(map(ur.sandbox_cmd.env) do x
+        base64encode(x)
```
staticfloat: This doesn't make sense to me; why must we `clearenv()` then generate a new one like we do in `sandbox.c` when in `init` mode?  Why can't we just inherit our environment like we do in non-init mode?

src/QemuRunner.jl
staticfloat: Faced with writing my own escaping parser, I would probably do the same.

src/QemuRunner.jl
```diff
@@ -0,0 +1,170 @@
...
+    end
+
+    c = 'b'
+    for (mount,dir) in mounts
+        qemu_cmd = `$qemu_cmd -drive if=virtio,file=$mount,format=raw`
+        sandbox_cmd = `$sandbox_cmd --map /dev/vd$c:$dir`
+        c = Char(Int(c) + 1)
+    end
+
+    sandbox_cmd = setenv(sandbox_cmd, merge(target_envs(triplet(platform)), extra_env))
```
staticfloat: See my recent modifications to `UserNSRunner`; this `setenv()` is more or less useless.

src/QemuRunner.jl
```diff
@@ -0,0 +1,170 @@
...
+    end
+
+    # If we're trying to run macOS and we have an SDK directory, mount that!
+    #=
+    if platform == MacOS()
+        sdk_version = "MacOSX10.10.sdk"
+        sdk_shard_path = joinpath(shards_cache, sdk_version)
+        push!(mapping, sdk_shard_path => joinpath("/opt", tp, sdk_version))
+    end
+    =#
```
staticfloat: Does this not work?

src/QemuRunner.jl
```diff
@@ -0,0 +1,170 @@
...
+"""
+type QemuRunner <: Runner
+    qemu_cmd::Cmd
+    append_line::String
+    sandbox_cmd::Cmd
+    platform::Platform
+end
+
+qemu_path() = "/Users/kfischer/Projects/qemu-hvf/x86_64-softmmu/qemu-system-x86_64"
+kernel_path() = "/Users/kfischer/Projects/qemu-hvf/x86_64-softmmu/bzImage"
```
staticfloat: So ideally both of these will be sourced via `BinaryProvider`, right?  I guess it would be fun to make a `Builder` repository for the kernel itself.....

src/RootFS.jl
```diff
@@ -189,22 +191,24 @@ function update_rootfs(triplets::Vector{S}; automatic::Bool = automatic_apple,
...
-                touch(string(squashfs_path,".sha256"))
             end
-            
+
             # Then mount it, if it hasn't already been mounted:
-            if !success(`mountpoint $(dest_dir)`)
+            if Compat.Sys.islinux() && !success(`mountpoint $(dest_dir)`)
                 mkpath(dest_dir)
                 run(`sudo mount $(squashfs_path) $(dest_dir) -o ro,loop`)
             end
```
staticfloat: Should the `sudo umount` a few lines below this (on line 217) also be gated on `islinux()`?

src/QemuRunner.jl
Keno: Yes and yes

src/QemuRunner.jl
Keno: Ah sorry forgot about this just need to add support to also serve this over 9p

src/QemuRunner.jl
Keno: The init process' environment is not inherited. Instead it's the kernel's startup environment, which contains a bunch of variables we don't want. So we clear that and put in the env we do want.

src/RootFS.jl
Keno: yes

src/QemuRunner.jl
staticfloat: Ah, that makes a lot of sense, thanks.

PR #115, opened 2018-01-03
Bento Sandbox: OSX tests passing

staticfloat:
With these changes, the sandbox is now bundled within the rootfs, and it is used automagically.  I completely rewrote argument passing, instead of parsing out base64-encoded chunks from `/proc/cmdline`, we now create a second serial interface within the qemu environment that is bound to a socket on the outside, and we just write the args and environment over in a simplistic protocol.

This passes tests on OSX now, which is thrilling.  Great work on the initial Qemu setup @keno!


.travis.yml
```diff
@@ -2,8 +2,7 @@
 language: julia
 os:
   - linux
-  # We don't have docker on OSX on travis at the moment...
-  #- osx
+  - osx
```
Keno: I suspect this won't work due to the lack of nested virtualization in the travis environment.

staticfloat: Well it didn't work, but it seems to be because of a segfault from another package?  That's weird.

staticfloat: It would also be really too bad if this didn't work on Travis, as that would eliminate a lot of the usefulness of this work, with regards to licensing.

Keno: Yeah, I know, I didn't find out that Travis disabled this until I was already most of the way through :/

PR #127, opened 2018-01-30
Use rm instead of rmdir (because it exists)

iamed2:
Stacktrace: https://www.dropbox.com/s/4hsozxml0gmvti5/Screenshot%202018-01-30%2016.54.38.png


Keno:
Ideally, we'd have a test case here to make sure this doesn't regress. It's not super straightforward to write to, so let me know if you just want to merge this and we'll get around to writing tests at some point.

iamed2:
TBH I don't have the time to figure out how to make a test case for this, sorry. Now if there was macOS support it could be work-related... 😉

PR #132, opened 2018-02-03
Add "privileged container" mode to `sandbox.c`

staticfloat:
On some Linux kernels, mounting overlay filesystems within user
namespaces is broken, see [0] for an Ubuntu issue tracking this and
their patch to fix it. To work around this, we allow the user to enable
running the sandbox through `sudo`, which adds another path through the
already convoluted `sandbox.c`.  After a thorough reorganization of
`sandbox.c`, we now have three well-defined execution modes (init,
unprivileged and privileged) and a `sandbox` that can either mount
overlay filesystems as root within a user namespace, or can mount
overlay filesystems as root outside of a user namespace, then enter the
user namespace.

[0] https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1531747


Keno:
If you rebase on current master, you'll get a fix to the Travis build config that'll at least test the sandbox from the repo.

docs/src/FAQ.md
```diff
@@ -22,4 +22,8 @@ Absolutely!  There's nothing Julia-specific about the binaries generated by the
 
 ### What platforms are supported?
 
-At the time of writing, we support Linux (x86_64, i686, armv7l, aarch64, ppc64le), Windows (x86_64, i686) and macOS (x86_64).
\ No newline at end of file
+At the time of writing, we support Linux (x86_64, i686, armv7l, aarch64, ppc64le), Windows (x86_64, i686) and macOS (x86_64).
+
+### At line XXX, ABORTED (Operation not permitted)!
+
+Some linux distributions have a bug in their `overlayfs` implementation that prevents us from mounting overlay filesystems within user namespaces.  See [this Ubuntu kernel bug report](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1531747) for a description of the situation and how Ubuntu has patched it in their kernels.  To work around this, you can launch `BinaryBuilder.jl` in "privileged container" mode.  Unfortunately, this involves running `sudo` every time you launch into a BinaryBuilder session, but on the other hand, this successfully works around the issue on distributions such as Arch linux.  To set "privileged container" mode, set the `BINARYBUILDER_RUNNER` environment variable to `privileged`.
```
Keno: Can we just auto detect this, warn and then just use sudo?

staticfloat: I would like to, but I poked around and I'm not yet sure of a good way to preemptively check this.  We could return a special error code from `sandbox` and auto-retry I suppose, but that means that we'd do it wrong every time, so I imagine the warning message would probably just say "if you don't want it to auto-retry every time, set this environment variable".

staticfloat:
Alright, quashed the last bugs, this is working on OSX, Arch and Ubuntu now.  Let's see what Travis says.

@Keno, I've noticed that the wizard is not completing a test due to the builtin HTTP server catching an exception, however it doesn't cause the tests to fail.  Is that intentional?  I would think we should be `@test`'ing something there so that the errors are fatal.

staticfloat:
Travis says OK on 0.6, so I'll merge once Keno gets a chance to give more feedback.

docs/src/FAQ.md
Keno: So I was imagining we'd do something like `sanbox --probe` that runs all the calls we need and then just returns success or error depending on whether that worked or not. We'd run that the first time we try the sandbox and remember it for later.

Keno:
> Is that intentional?

Probably not

docs/src/FAQ.md
staticfloat: In all the julia packages I’ve ever written, I’ve never had “persistent configuration”  by using anything other than environment variables. What do you suggest?

In any case, we should definitely have a mechanism for overriding the “preferred runner”, so this is fine, but I don’t have a great way to store configuration for a package.

Keno: I don't want it to be persistent, just check for every julia process once we try to run the sandbox for the first time in that process. It's a quick check.

PR #137, opened 2018-02-06
tweak prose

KristofferC:
Could also put a `sir/madame` instead.


ararslan:
If you want to keep the silly vibe, you could say, "Please specify how you would like to proceed, captain."

staticfloat:
I enjoy the silly vibe, but whatever.  If we get rid of this, we should get rid of the comment above it as it's no longer true.

KristofferC:
I don't care to be honest. I'm fine with leaving it. I don't think many people will get the reference though.

davidanthoff:
I think you should merge this PR. Very few people will get the reference, and if you don't get it, the current language is not ok.

PR #142, opened 2018-02-07
Fix a call to canonicalize_source_url

davidanthoff:
Right now this errors. I think this is meant to call ``canonicalize_source_url``, but someone better double check.

### DIFF
--- a/src/wizard/obtain_source.jl
+++ b/src/wizard/obtain_source.jl
@@ -241,7 +241,7 @@ function obtain_binary_deps(state::WizardState)
                 print(state.outs, "> ")
                 url = readline(state.ins)
                 println(state.outs)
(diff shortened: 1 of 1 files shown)


Keno:
One of them canonicalizes GitHub repositories, the other canonicalizes references to files within a GitHub repository.

staticfloat:
So by that logic, this should be the correct fix.

Keno:
I think this should probably be `canonicalize_file_url`, since we're asking for a single build.jl file not a directory containing one.

staticfloat:
Oh, I see what you're saying.  I did not understand the difference.  :P

staticfloat:
I've fixed this in #148

PR #146, opened 2018-02-13
Adjust output script style, use printstyled

ararslan:
The first commit here fixes a couple of stylistic oddities: seemingly haphazard use of 2- or 4-space indents, and a named closure instead of a normal function. These were both bothering me. 🙂 

The second commit replaces the deprecated `print_with_color` with calls to `printstyled`, which is available in Compat.


staticfloat:
Thanks Alex!  I saw the `printstyled()` errors and I just groaned to myself.  Thanks for doing my drudgery for me!

PR #149, opened 2018-02-16
Invert installation checks

staticfloat:
Allow binaries installed within the prefix to be used regardless of our currently running platform.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -161,20 +161,20 @@ function print_buildjl(io::IO, product_hashes::Dict; products_str=example_produc
     println(io, ")")
 
     print(io, """
-    if platform_key() in keys(download_info)
-        # First, check to see if we're all satisfied
-        if any(!satisfied(p; verbose=verbose) for p in products)
+    # First, check to see if we're all satisfied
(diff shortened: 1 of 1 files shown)


src/AutoBuild.jl
```diff
@@ -161,20 +161,20 @@ function print_buildjl(io::IO, product_hashes::Dict; products_str=example_produc
     println(io, ")")
 
     print(io, """
-    if platform_key() in keys(download_info)
-        # First, check to see if we're all satisfied
-        if any(!satisfied(p; verbose=verbose) for p in products)
+    # First, check to see if we're all satisfied
+    if any(!satisfied(p; verbose=verbose) for p in products)
+        if platform_key() in keys(download_info)
```
ararslan: `haskey(download_info, platform_key())`?

staticfloat: Why wouldn't I use the `in` keyword?  That's what it's designed for.

ararslan: In this case it's just an efficiency thing; it should be more efficient to check using `haskey` rather than creating the intermediate `KeyIterator` from `keys`.

PR #150, opened 2018-02-17
First try at "Building tips" documentation

tshort:
The intent is to provide tips for common build scenarios. If common items keep appearing, they could be baked into the environment.


docs/src/build_tips.md
```diff
@@ -0,0 +1,135 @@
...
+## Initiating different shell commands based on target
+
+Sometimes, you need to adapt build scripts based on the target platform. This can be done within the shell script. Here is an example:
+
+```sh
+if [[ ${target} == *-apple-* ]]; then
+    export CC=clang
+    export CXX=clang++
+    export LDFLAGS="-mmacosx-version-min=10.8"
+fi
```
staticfloat: While I do want an example like this, this particular example has been made completely defunct.  Let's choose some other example.  Perhaps something from [OpenBLASBuilder](https://github.com/staticfloat/OpenBLASBuilder/blob/master/build_tarballs.jl)?

docs/src/build_tips.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+Here are examples of scripts with target-specific checks:
+
+* [Keno/ReadStatBuilder](https://github.com/Keno/ReadStatBuilder/blob/master/build_tarballs.jl#L35-L39) - windows check
+* [JuliaDiffEq/SundialsBuilder](https://github.com/JuliaDiffEq/SundialsBuilder/blob/cmake-mod/build_tarballs.jl#L125-L131) - 32-bit check
+* [staticfloat/IpoptBuilder](https://github.com/staticfloat/IpoptBuilder/blob/master/build_tarballs.jl#L28-L32) - apple check
+
```
staticfloat: Let's add OpenBLASBuilder to this list.  Also make sure to canonicalize your URLs (press `y` in your browser window) so that when the files change, the line numbers (`#L28-32`, etc...) don't point to random chunks of text.

docs/src/build_tips.md
```diff
@@ -0,0 +1,135 @@
...
+* [staticfloat/NettleBuilder](https://github.com/staticfloat/NettleBuilder/blob/master/build_tarballs.jl)
+
+
+## CMake builds
+
+For CMake, the wizard will suggest a template for running CMake. Typically, this will look like:
+
+```sh
+make -DCMAKE_INSTALL_PREFIX=/ -DCMAKE_TOOLCHAIN_FILE=/opt/$target/$target.toolchain
+```
```
staticfloat: `-DCMAKE_INSTALL_PREFIX=$prefix` is the new hint.  Sorry that changed recently. :)

docs/src/build_tips.md
```diff
@@ -0,0 +1,135 @@
...
+
+The toolchain file sets up several CMake environment variables for better cross -DSUNDIALS_INDEX_TYPE=int32_t-platform support:
```
staticfloat: Perhaps I'm just having a hard time parsing this, but what's up with the `-DSUNDIALS_INDEX_TYPE=int32_t-platform`?

docs/src/build_tips.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+There are plans to handle file changes in the wizard automatically [(#25)](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/125).
+
+
+## Other examples
+
+Examples of other interesting builders include:
+
+* [staticfloat/IpoptBuilder](https://github.com/staticfloat/IpoptBuilder/blob/master/build_tarballs.jl) -- Includes multiple sources, and each of these has different build approaches. Includes custom `CC` and `CXX` definitions for macOS to use Clang and sets `LDFLAGS` to specify a minimum macOS version.
```
staticfloat: The special things here for `IpoptBuilder` are about to go away, sorry!  I promise we'll have more unusual things for you to include here soon.  :)

docs/src/build_tips.md
davidanthoff: For ReadStatBuilder probably better to point to my repo, that is the one I will use for ReadStat.jl.

staticfloat:
This is really, really great.  Thank you so much for bundling this all together.

There seems to be something wrong with the documentation for this repository; I think I got `Documenter.jl` to build it locally once, but my Travis deployment strategy seems quite flawed.  If anyone has experience with Documenter, I'd be happy to hear from them how to fix it.  ;)

docs/src/build_tips.md
tshort: Mistaken paste, I think. (I'm doing this in the web interface.)

docs/src/build_tips.md
tshort: Interesting. I didn't know about `y`. Makes sense!

tshort:
On Documenter builds, it's been a while since I set one up. Do you have a deploy key set up for documenter in this repo? Also, your `deploydocs()` invocation in make.jl has less arguments than some others I've seen, but I don't know if that's it.

You could also try setting `DOCUMENTER_DEBUG=true` in Travis. That should make Travis spit out some debug info that may help. I can't remember exactly how to set this. Using `ENV` will probably work.

davidanthoff:
On docs: https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/156

staticfloat:
Woohoo!  Nice!

PR #168, opened 2018-02-28
0.7 compat

quinnj:


bin/ssh_server.jl
```diff
@@ -11,7 +11,7 @@ sock = listen(port)
 pubkey_to_user_map = Dict{Vector{UInt8}, String}()
 user_to_pubkey_map = Dict{String, Vector{UInt8}}()
 
-const github_token = strip(readstring(joinpath(dirname(@__FILE__),"..","etc","github_tok")))
+const github_token = strip(String(read(joinpath(dirname(@__FILE__),"..","etc","github_tok"))))
```
ararslan: I think the correct replacement is `read(..., String)`

src/Auditor.jl
```diff
@@ -107,36 +107,36 @@ function audit(prefix::Prefix; io=STDERR,
                                 msg = replace("""
                                 Linked library $(libname) has been auto-mapped to
                                 $(new_link)
-                                """, '\n', ' ')
-                                info(io, strip(msg))
+                                """, '\n' => ' ')
+                                println(io, sINFO: trip(msg))
```
ararslan: ?

src/AutoBuild.jl
```diff
@@ -19,7 +19,7 @@ function autobuild(dir::AbstractString, src_name::AbstractString,
         run_travis_busytask = true
         travis_busytask = @async begin
             # Don't let Travis think we're asleep...
-            info("Brewing a pot of coffee for Travis...")
+            Compat.@info("Brewing a pot of coffee for Travis...")
```
ararslan: Why use `Compat.@info` here but change the other `info` uses to `println`s?

quinnj: The other uses were doing `info(io, ...)`, of which there isn't a direct compat for the new logging system (i.e. logging to a specific io stream). I tried to change as little as possible to allow package authors to eventually redesign overall logging as they see fit.

src/Auditor.jl
quinnj: fixed

src/AutoBuild.jl
ararslan: We could make those `printstyled("INFO: ...", color=Base.info_color())` but that's probably overkill.

bin/ssh_server.jl
quinnj: I don't think there's anything wrong w/ `String(read(file))` though?

src/AutoBuild.jl
quinnj: Better to just get 0.7 out already so we can all do proper logging 😉

bin/ssh_server.jl
ararslan: I suppose not, but I believe `read(file, String)` avoids creating the intermediate array.

quinnj: I think it creates a `StringVector` underneath, which is essentially the exact same thing as `String(read(file))`

src/BinaryBuilder.jl
```diff
@@ -2,13 +2,11 @@
 # __precompile__()
 
 module BinaryBuilder
-using Compat
+using Compat, Compat.Libdl, Compat.LibGit2, Compat.Random
 using Reexport
 using ObjectFile
+using Nullables
```
staticfloat: I would like to avoid having to depend on `Nullables`; is that necessary?  I don't see any `Nullable` stuff here; `Compat` should be providing `Nothing` already.

REQUIRE
```diff
@@ -10,3 +10,4 @@ PkgDev
 JSON
 HTTP 0.6.0
 ProgressMeter
+Nullables
```
staticfloat: Same here, would like to avoid `Nullable`s in my `REQUIRE`

src/QemuRunner.jl
```diff
@@ -250,7 +250,7 @@ function run_interactive(qr::QemuRunner, cmd::Cmd; stdin = nothing, stdout = not
     end
     
     # Qemu appears to mess with our terminal
-    Base.reseteof(STDIN)
+    Base.reseteof(stdin)
```
staticfloat: `Base.stdin` is shadowed by the `stdin` parameter being passed into this function; so I think we need to make this explicitly `Compat.stdin`.

src/wizard/obtain_source.jl
```diff
@@ -258,7 +258,7 @@ function obtain_binary_deps(state::WizardState)
                     String(HTTP.get(canon_url).body)))
             elseif bindep_select == 2
                 println(state.outs, "Please provide the build.jl file. Press ^D when you're done")
-                script = String(read(STDIN))
+                script = String(read(stdin))
```
staticfloat: I think this should be `state.ins`.  Feel free to change it.

src/wizard/deploy.jl
```diff
@@ -454,7 +454,7 @@ function github_deploy(state::WizardState)
                 print_travis_deploy(f, gr, secure_key)
             end
         catch e
-            Base.display_error(STDERR, e, catch_backtrace())
+            Base.display_error(Compat.stderr, e, catch_backtrace())
```
staticfloat: Does this need to be `Compat.stderr`?  Can it just be `stderr`?

src/auditor/instruction_set.jl
```diff
@@ -1194,7 +1194,7 @@ if `verbose` is set to `true`.
 Note that this function only really makes sense for x86/x64 binaries.  Don't
 run this on armv7l, aarch64, ppc64le etc... binaries and expect it to work.
 """
-function analyze_instruction_set(oh::ObjectHandle; verbose::Bool = false, io::IO = STDOUT)
+function analyze_instruction_set(oh::ObjectHandle; verbose::Bool = false, io::IO = Compat.stdout)
```
staticfloat: Does this need to be `Compat.stdout`?  Can it just be `stdout`?

src/Wizard.jl
```diff
@@ -55,7 +56,7 @@ function run_wizard(state::WizardState = WizardState())
             printstyled(state.outs, msg, bold=true, color=:red)
         else
             bt = catch_backtrace()
-            Base.showerror(STDERR, err, bt)
+            Base.showerror(Compat.stderr, err, bt)
```
staticfloat: Can this just be `stderr` instead of `Compat.stderr`?

src/UserNSRunner.jl
```diff
@@ -83,9 +83,9 @@ function show(io::IO, x::UserNSRunner)
           "UserNSRunner")
 end
 
-function Base.run(ur::UserNSRunner, cmd, logpath::AbstractString; verbose::Bool = false, tee_stream=STDOUT)
+function Base.run(ur::UserNSRunner, cmd, logpath::AbstractString; verbose::Bool = false, tee_stream=Compat.stdout)
```
staticfloat: Can this just be `stdout`?

src/QemuRunner.jl
```diff
@@ -195,7 +195,7 @@ function qemu_gen_cmd(qr::QemuRunner, cmd::Cmd, comm_socket_path::String)
     return `$long_cmd`
 end
 
-function Base.run(qr::QemuRunner, cmd, logpath::AbstractString; verbose::Bool = false, tee_stream=STDOUT)
+function Base.run(qr::QemuRunner, cmd, logpath::AbstractString; verbose::Bool = false, tee_stream=Compat.stdout)
```
staticfloat: I think this can just be `stdout`.

src/Auditor.jl
```diff
@@ -23,7 +23,7 @@ This method is still a work in progress, only some of the above list is
 actually implemented, be sure to actually inspect `Auditor.jl` to see what is
 and is not currently in the realm of fantasy.
 """
-function audit(prefix::Prefix; io=STDERR,
+function audit(prefix::Prefix; io=Compat.stderr,
```
staticfloat: I think this can just be `stderr`.

src/BinaryBuilder.jl
quinnj: yeah, there are two uses of Nullable in src/wizard/deploy.jl; I'll see if I can get rid of them

src/QemuRunner.jl
quinnj: done

src/wizard/obtain_source.jl
quinnj: done

src/wizard/deploy.jl
quinnj: done

src/BinaryBuilder.jl
quinnj: Hmmmm, looks like 0.6 LibGit2 requires the Nullables for `push`. What do you want to do?

staticfloat: Curses.  Let's include `Nullables` then.  Thanks so much for digging through this.  As a reward for all your hard work, I am about to open a new PR that makes our `build_tarballs.jl` significantly simpler.

quinnj: np. Luckily 0.7 is imminent and then we can all get rid of all this compat junk making our code ugly.

staticfloat:
Thanks Jacob!

Issue #183, opened 2018-03-11
Trying to build libxls crashes wizard


davidanthoff:
I can't copy the error log out of my VM right now, but this is super easy to repo (it happens both on master and the latest tagged versions of BinaryProvider/BinaryBuilder):
1. Start the wizard
2. Get sources from https://github.com/svn2github/libxls ``master``
3. View it crash when it drops into the cross compilation environment

staticfloat:
Good catch David!  This is due to a broken symlink which, while sloppy repository management, definitely shouldn't kill the wizard.  Fix pending in https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/186

PR #186, opened 2018-03-16
Fix Wizard choking on broken symlinks

staticfloat:
Also add a test to make sure this doesn't happen in the future.

This closes https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/183


src/wizard/hints.jl
```diff
@@ -41,8 +41,16 @@ function provide_hints(state::WizardState, path::AbstractString)
     for (root, dirs, files) in walkdir(path)
         for file in files
             file_path = joinpath(root, file)
-            contents = String(read(file_path))
-            if file == "configure" && contains(contents, "Generated by GNU Autoconf")
+
+            # Helper function to try to read the given path's contents, but
+            # returning an empty string on error (for e.g. broken symlinks)
+            read_contents = (path) -> try
```
ararslan: Why not `read_contents(path) = try ...`?

ararslan: You could also move it out of the loop, since it doesn't need to be regenerated with every iteration

staticfloat: Is there a reason to prefer `foo(x) = ...` instead of `foo = x -> ...`?

staticfloat: The reason I have it within the loop is because I like to have it right next to the only place it's being used; reduce mental strain on whoever is trying to read this code.  To really squeeze that last 0.1% performance out of this loop, let's just hope that the compiler realizes this value is a constant and hoists it out of the loop.  ;)

ararslan: Stylistically it's odd to use an anonymous function if you're going to name it.

PR #191, opened 2018-03-20
Add `libstdc++-6.dll` and `libgcc_s_seh-1.dll` to our ignore list

staticfloat:
More libraries we don't care to complain about


vchuravy:
We probably want to ignore  the whole shebang of GCC libraries that we pick up like libatomic, libsspe etc..

PR #192, opened 2018-03-20
Add global github auth so that we don't run into rate-limiting on Travis

staticfloat:


staticfloat:
This is provided via a secret environment variable embedded within the Travis config, stored through the web ui.

PR #197, opened 2018-03-23
Add tar.bz2 file support

visr:
Only format readily available for the project I'm working on,
https://download.osgeo.org/geos/


staticfloat:
Thanks!

Issue #153, opened 2018-02-18
CTRL-C within QemuRunner kills Julia


staticfloat:
We need to do something to ensure that the CTRL-C we try to send to the QemuRunner doesn't get intercepted by Julia and kill everything.

staticfloat:
Keno suggests embedding an SSH server into the rootfs and using that to do all TTY manipulations; we would then tell `init` to start up the SSH server, then connect to it over an emulated network interface built by QEMU.

Keno:
Just to note that all we really need here is the SSH TTY protocol, so we may just be able to take one of  those SSH servers designed for embedded use, strip out all the authentication stuff and connect it over a serial port. The disadvantage of that would be that we couldn't rely on the native ssh client. I do have the necessary client code in [SSH.jl](https://github.com/Keno/SSH.jl/blob/master/src/SSH.jl#L680-L813) to decode the tty protocol. The advantage of doing it that way would be that we could avoid messing with the linux network stack. Of course, we may want to add networking anyway to allow network request from within the sandbox, which I think is not working currently using the Qemu runner.

staticfloat:
I'm closing this as solved by Isaiah in the PR referenced above, as this behavior no longer appears; if we still want to go down the SSH route, we can, but we will need a new motivation for it, as this works and is much simpler.  ;)

PR #202, opened 2018-03-29
Use stty to change interrtupt when running interactive QEMU shell

ihnorton:
ref #153. Maybe not the desired long-term fix, but this makes the macOS experience much less annoying.


staticfloat:
Wow, yes, very yes, thank you!

ihnorton:
Oof `interrTupt` 🙃😪

staticfloat:
Just making a note here that we may want to remove the `isapple()` checks, as this same fix will probably be useful on Linux as well, if we ever use the QEMU runner there.

PR #213, opened 2018-04-05
more libs to ignore

stevengj:
I guess the principle here is to ignore libraries that always ship with the OS?  If so, libdl and librt should qualify.


Keno:
Yes, we should ignore things that we pick up from the OS (libc and related), or that we ship with julia and expect to be picked up from there (libstdc++)

Issue #233, opened 2018-04-27
exceeding Travis build-time limit


stevengj:
There are now so many BinaryBuilder platforms that my builds are exceeding Travis' 50-minute time limit: https://travis-ci.org/JuliaInterop/ZMQBuilder/builds/372036524 (`The job exceeded the maximum time limit for jobs, and has been terminated.`).

What can be done about this?

KristofferC:
Perhaps using multiple jobs like:

https://github.com/staticfloat/LLVMBuilder/blob/98449b944fe0a148812088b1f9d8042955f2849c/.travis.yml#L18-L28

stevengj:
One annoyance with specifying TARGET manually is that then I need to keep my `.travis.yml` file in sync with `supported_platforms()`.

Maybe I can hack something together like:
```sh
# Our build takes too long for one job, so split across 4 jobs:
matrix:
    - PART=1/4
    - PART=2/4
    - PART=3/4
    - PART=4/4
```
and then just divide the `platforms` list roughly equally among 4 jobs.

stevengj:
It looks like this should work to implement a `--part=n/m` argument:
```jl
i = Compat.findlast(x -> startswith(x, "--part="), ARGS)
if i !== nothing
    p = parse.(Int, split(ARGS[i][8:end], '/'))
    (length(p) == 2 && p[2] > 0 && 1 ≤ p[1] ≤ p[2]) || error("invalid argument ", ARGS[i])
    n = (length(platforms) + p[2]-1) ÷ p[2]
    platforms = platforms[n*(p[1]-1)+1:min(end,n*p[1])]
end
filter!(x -> !startswith(x, "--part="), ARGS)
ARGS = filter!(x -> !startswith(x, "--part="), ARGS)
```

stevengj:
Yes, the above code seems to have worked for ZMQBuilder.  I'll make a PR to support `--part` in BinaryBuilder itself.

staticfloat:
The problem here is that the first job generates a `build.jl`, which is then uploaded, and Travis won't overwrite it when your last build stage runs.  The logic within `build_tarballs()` disables `build.jl` generation if you've overridden the platforms (e.g. by passing in a comma-separated list of targets) so what I would do is instead of the `--part` stuff where you have `--part=1/3` and then modify `platforms`, bake the actual platforms you want built into your `.travis.yml`.

[...]

PR #234, opened 2018-04-28
add --part=n/m to build_tarballs

stevengj:
This adds support for a `--part=n/m` flag, discussed in #233, as an alternative giving a manual list of build targets, in order to speed up builds by breaking them into multiple jobs.  Tested on https://github.com/JuliaInterop/ZMQBuilder.

Rationale: giving a manual list of build targets is error prone and tedious because it requires one to keep the `.travis.yml` file in sync with the platforms in the build script — even worse if one is using `supported_platforms()` to build on all supported platforms, in which case you have to keep `.travis.yml` in sync with BinaryBuilder itself.   With `--part=n/m`, you divide the build into `m` jobs regardless of how many platforms you are building for.

Similar to https://github.com/JuliaInterop/ZMQBuilder/blob/master/.travis.yml, your `.travis.yml` file would contain something like
```sh
env:
  global:
  - BINARYBUILDER_DOWNLOADS_CACHE=downloads
  - BINARYBUILDER_AUTOMATIC_APPLE=true
  # Our build takes too long for one job, so split targets across multiple jobs
  matrix:
    - PART=1/3
    - PART=2/3
    - PART=3/3
sudo: required
jobs:
  include:
    - stage: regenerate build.jl
      script: julia build_tarballs.jl --only-buildjl
      if: tag IS present
script:
- julia build_tarballs.jl --part=$PART
```


staticfloat:
Thanks, this is neat!

PR #238, opened 2018-05-08
Provide an iteration workflow in the build tips.

ViralBShah:


ViralBShah:
Any comments on this one?

tshort:
I think something on an iteration workflow is worthwhile. My workflow is mainly to use `ARGS` with `--debug`. That dumps you in the builder shell on error. Then, I don't need to manually find the right folder and run `BinaryBuilder.runshell`.

Issue #241, opened 2018-05-09
instruction_mnemonics() takes a lot of memory


ViralBShah:
The ArpackBuilder's osx build was failing on Travis, which @staticfloat debugged to the `instruction_mnemonics()` being inefficient with its string processing.

It takes 12GB+ of memory during the `Analyzing minimum instruction set for lib/libopenblas64_p-r0.2.20.dylib`

PR #243, opened 2018-05-12
Avoid reading all lines at once by using eachline instead of readlines

andreasnoack:
I think this fixes https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/241. However, the `IOBuffer` might still be quite large. I've confirmed that it fixes https://github.com/ViralBShah/ArpackBuilder/issues/1

cc: @viralbshah


staticfloat:
Oh boy.  That's a simple one.  Nice.

Issue #242, opened 2018-05-09
warnings - load is deprecated, use payload


ViralBShah:
````
WARNING: load(m::Message) is deprecated, use payload(m, String) instead.
Stacktrace:
 [1] depwarn(::String, ::Symbol) at ./deprecated.jl:70
 [2] load(::HTTP.Messages.Response) at ./deprecated.jl:57
 [3] obtain_token(::Base.TTY, ::Base.TTY, ::String, ::String) at /home/viralbshah/.julia/v0.6/BinaryBuilder/src/wizard/deploy.jl:235
 [4] github_deploy(::BinaryBuilder.WizardState) at /home/viralbshah/.julia/v0.6/BinaryBuilder/src/wizard/deploy.jl:415
 [5] step7(::BinaryBuilder.WizardState) at /home/viralbshah/.julia/v0.6/BinaryBuilder/src/wizard/deploy.jl:519
 [6] run_wizard(::BinaryBuilder.WizardState) at /home/viralbshah/.julia/v0.6/BinaryBuilder/src/Wizard.jl:49
 [7] eval(::Module, ::Any) at ./boot.jl:235
 [8] eval_user_input(::Any, ::Base.REPL.REPLBackend) at ./REPL.jl:66
 [9] macro expansion at ./REPL.jl:97 [inlined]
 [10] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73
while loading no file, in expression starting on line 0
````

PR #246, opened 2018-05-18
Use `HTTP.payload()` instead of `HTTP.load()`

staticfloat:
Fixes https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/242

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -8,6 +8,6 @@ MbedTLS
 GitHub
 PkgDev
 JSON
-HTTP 0.6.0
+HTTP 0.6.3
 ProgressMeter
 Nullables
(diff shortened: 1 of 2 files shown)


ararslan:
The minimum version of HTTP in REQUIRE should probably be adjusted to when `payload` was introduced to replace `load`.

staticfloat:
@samoconnor do you have any idea when that would be?  Man, I wish we had an automated way to figure those kinds of things out.

ararslan:
Looks like it's v0.6.3: https://github.com/JuliaWeb/HTTP.jl/commit/b84e6630c22bb954e9b22bc5902b385c7c5f4a33#diff-b683796309866200c2760b3695b2ac45

PR #247, opened 2018-05-20
add option to only turn on verbose output for the auditor

vchuravy:
While building LLVM I found myself wanting verbose output only for the audit 
stage. This might be less important once we move to 0.7 since we can associate
information with a warning message, but currently I see a lot of warnings
and no indication for which binary they are from.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -21,7 +21,7 @@ function build_tarballs(ARGS, src_name, sources, script, platforms, products,
     if "--help" in ARGS
         println(strip("""
         Usage: build_tarballs.jl [target1,target2,...] [--only-buildjl]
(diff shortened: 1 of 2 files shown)


staticfloat:
Sorry, I'm not sure why this is useful?  What messages are you trying to turn off for this?  The sandbox verbosity stuff?

vchuravy:
During the llvm build I mostly care about the auditor output. Once you are
building multiple platforms at once the verbose output from the next build
is hiding the output from the auditor.

For me the actual problem is that the warnings in the auditor in non
verbose mode are not useful, since I have no idea which build product they
belong to.

staticfloat:
Okay great point, the warnings that don't have identifying information in them are fixed in 421ca15

PR #252, opened 2018-05-23
Add `--debug` and cache downloads better

### LINKED ISSUE #237: Add `--debug` to `build_tarballs.jl` to drop into a shell
It shouldn't be too hard to add a `--debug` option to `build_tarballs.jl` such that we can drop into a `runshell()` environment within a platform that just failed.

staticfloat:
Closes #237

### DIFF
--- a/deps/sandbox.c
+++ b/deps/sandbox.c
@@ -495,7 +495,7 @@ static int sandbox_main(const char * root_dir, const char * new_cd, int sandbox_
     sigwait(&waitset, &sig);
 
     pid_t reaped_pid;
-    while ((reaped_pid = waitpid(-1, &status, WNOHANG)) != -1) {
+    while ((reaped_pid = waitpid(-1, &status, NULL)) != -1) {
       if (reaped_pid == main_pid) {
         // If it was the main pid that exited, return as well.
         return WIFEXITED(status) ? WEXITSTATUS(status) : 1;
--- a/deps/sandbox.c
+++ b/deps/sandbox.c
@@ -130,6 +130,12 @@ static int open_proc_file(pid_t pid, const char *file, int mode) {
   return fd;
 }
 
+/* `touch` a file; create it if it doesn't already exist. */
+static void touch(const char * path) {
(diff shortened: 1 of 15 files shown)


deps/sandbox.c
```diff
@@ -480,6 +500,8 @@ static int sandbox_main(const char * root_dir, const char * new_cd, int sandbox_
         // If it was the main pid that exited, return as well.
         return WIFEXITED(status) ? WEXITSTATUS(status) : 1;
       }
+      // Sleep for 100ms so that we're not chewing through CPU cycles
+      usleep(100000);
```
ararslan: :zzz:

vchuravy:
I can confirm that fde03a8 does what it supposed to do.

deps/sandbox.c
Keno: We should just stop passing WNOHANG to waitpid

staticfloat: Yeah, duh.  For some reason I thought we were doing more work within this `while` loop.....

staticfloat:
Quick summary of changes here:

* `--debug` allows entering an interactive session when a `build_tarballs.jl` run fails.  The interactive environment does its best to pretend it's at the last step that was executed within the buildscript (e.g. at the point of script exit, it saves all environment variables, mimics a bash history and saves the current working directory) so that when you get dropped into a shell, you can just `<up arrow> <enter>` and run the command that failed.

* The above is possible because we now mount another workspace at `/meta`, which holds things like `.bash_history` and `.env`, the file that gets sourced by `bash` when the debug shell is opened.

* `autobuild()` now takes in a `downloads_dir` kwarg, that allows `build_tarballs.jl` to download dependencies to a shared directory, which should speed up building multiple platforms at once, as the dependencies will be downloaded only once now, instead of once per platform.  This relies on https://github.com/JuliaPackaging/BinaryProvider.jl/commit/4ec2ce0fcc9f7f9cbc5c149f795f74299cefccda, so we'll have to tag a new `BinaryProvider` release before we can tag a release here.  I'll probably do that after I fix `BinaryProvider` tests on 0.7.

PR #255, opened 2018-05-26
Fix typos in 9p mounting code

Keno:
@staticfloat, we'll need a new base rootfs with this fix.


staticfloat:
Thanks, I haven't used the QEMU sandbox in forever; perhaps this summer I'll try to get buildbot-based CI working for this so that we don't bitrot it in the future.

PR #274, opened 2018-06-12
Use fastly to cache rootfs images

staticfloat:
This is possible now that the images are all <2GB, as we have shattered the build image across multiple shards.


staticfloat:
Yep, it's working.  I'm just gonna merge it.

PR #276, opened 2018-06-12
Test GitHub deploy

Keno:
Depends on https://github.com/JuliaWeb/GitHub.jl/pull/112

### DIFF
--- a/src/wizard/obtain_source.jl
+++ b/src/wizard/obtain_source.jl
@@ -51,8 +51,13 @@ end
 function clone(url, source_path)
     p = Progress(0, 1, "Cloning: ")
     @compat_gc_preserve p begin
-        callbacks = LibGit2.RemoteCallbacks(transfer_progress=cfunction(transfer_progress, Cint, Tuple{Ptr{GitTransferProgress}, Any}),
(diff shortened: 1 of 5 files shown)


src/wizard/obtain_source.jl
```diff
@@ -51,8 +51,13 @@ end
 function clone(url, source_path)
     p = Progress(0, 1, "Cloning: ")
     @compat_gc_preserve p begin
-        callbacks = LibGit2.RemoteCallbacks(transfer_progress=cfunction(transfer_progress, Cint, Tuple{Ptr{GitTransferProgress}, Any}),
-            payload = pointer_from_objref(p))
+        @static if VERSION >= v"0.7-"
+            callbacks = LibGit2.RemoteCallbacks(transfer_progress=@cfunction(transfer_progress, Cint, (Ptr{GitTransferProgress}, Any)),
```
ararslan: You can just use `@cfunction` from Compat.

Keno: Femtocleaner will remove the version check automatically, and it's required for the next line anyway.

PR #292, opened 2018-06-21
Eliminate snakes from codebase

staticfloat:
:one: :zero: :package: -> :snake: -> :airplane:

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -455,6 +455,22 @@ end
     end
 end
 
+@testset "Auditor - absolute paths" begin
+    prefix = Prefix(tempname())
+    try
+        sharedir = joinpath(prefix.path, "share")
+        mkpath(sharedir)
+        open(joinpath(sharedir, "foo.conf"), "w") do f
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -455,6 +455,22 @@ end
...
+        sharedir = joinpath(prefix.path, "share")
+        mkpath(sharedir)
+        open(joinpath(sharedir, "foo.conf"), "w") do f
+            write(f, "share_dir = \"$sharedir\"")
+        end
+
+        # Test that `audit()` warns about an absolute path within the prefix
+        Compat.@info("Expecting a warning about share/foo.conf:")
+        BinaryBuilder.audit(prefix)
+    finally
```
ararslan: Didn't you mean `rescue`? :trolleybus:

staticfloat: we're not rich enough to afford that kind of syntax

staticfloat: maybe if we sent more of our coders into the mines, they might come back out with some more gems like those

Issue #294, opened 2018-06-21
We should version build products


staticfloat:
both autogenerated `build.jl` files and `.tar.gz` files should have the version embedded within them.  Right now we generate:

```
build.jl
Foo.x86_64-linux-gnu.tar.gz
Foo.i686-w32-mingw32.tar.gz
...
```

I argue we should generate something more similar to:

```
build_Foo-v1.2.0.jl
Foo.v1.2.0.x86_64-linux-gnu.tar.gz
Foo.v1.2.0.i686-w64-mingw32.tar.gz
```

It is unfortunate that version numbers naturally have `.` within them, which kind of breaks the natural separation between fields that we've been using so far (e.g. `$(name).$(platform).tar.gz`).

davidanthoff:
What about underscores for the separation?

staticfloat:
I figured it was much more likely for users to put underscores in names than dots, but so far I haven't seen any names with underscores within them.  That's a decent proposal; note first that `x86_64` would kinda-sorta screw that up, and also that the only parsing we do on these files so far is to grab `\.(.*)\.tar\.gz` as the target triplet, which doesn't rely on any kind of separation between fields.

PR #296, opened 2018-06-22
Version output products

staticfloat:
This embeds versions into output tarball and build.jl files.  This is a breaking change, as `build_tarballs.jl` files now must pass in a `version` to `build_tarballs()`.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -3,8 +3,8 @@ import GitHub: gh_get_json, DEFAULT_API
 import SHA: sha256
 
 """
-    build_tarballs(ARGS, src_name, sources, script, platforms, products,
-                   dependencies)
+    build_tarballs(ARGS, src_name, src_version, sources, script, platforms,
+                   products, dependencies)
 
 This should be the top-level function called from a `build_tarballs.jl` file.
 It takes in the information baked into a `build_tarballs.jl` file such as the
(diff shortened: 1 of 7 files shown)


staticfloat:
This closes https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/294

ararslan:
I wonder whether it would make more sense to have the source versioning be part of the source specification? That is, currently we specify sources as an array of name, SHA pairs. But since the version is a property of the source, it seems like it might make more sense to specify it as `url => (version, sha)`, i.e. a `Pair{String,Tuple{VersionNumber,String}}`. Then rather than warning when a user doesn't specify a version in the `build_tarballs` signature, you would instead warn if the user specifies a source as `url => sha` instead of `url => (version, sha)`.

ararslan:
Alternatively, we could introduce a `Source` type, then require the sources be specified as a `Vector{Source}`, where the `Source` constructor would take a URL, SHA, and version.

staticfloat:
So what do we do when there are multiple sources, each with their own version?

ararslan:
In the first thing I said:
```julia
sources = [
    "https://great/thing.tar.gz" => (v"4.2.0", "abc123"),
    "https://super/cool/whatever.tar.gz" => (v"0.1.2", "123abc"),
]
```
and in the second,
```julia
sources = [
    Source("https://great/thing.tar.gz", version=v"4.2.0", sha="abc123"),
    Source("https://super/cool/whatever.tar.gz", version=v"0.1.2", sha="123abc"),
]
```

staticfloat:
What I mean is, the `version` that is being used here identifies the build products that we create.  So with those two sources, what version do I give the output of the build?

ararslan:
Ah, I see. In that case, it seems like the version would be a property of the `LibraryProduct` (or `ExecutableProduct`).

staticfloat:
While we could do that, I think it would conflate the idea of the version of a library (each of which can be different) and the version of the overall release itself; while oftentimes it is true, there is a single library that is the output of a Builder repository, in general it's not true.  We don't really have a concept of a `Release` or anything, all that information is just "whatever we pass to `build_tarballs()`.  So I think this is as good as it's going to get.

stevengj:
It looks like you never updated the documentation?

https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/ce39c5590c3ed97b6a7bf780bb42ddc2397e89f5/docs/src/index.md

PR #303, opened 2018-07-03
ignore missing platforms of a binary dependency

SimonDanisch:
Possibly the worst fix possible, but I thought it's slightly more useful than just opening an issue.

### DIFF
--- a/src/wizard/interactive_build.jl
+++ b/src/wizard/interactive_build.jl
@@ -488,11 +488,13 @@ function step5c(state::WizardState)
     println(state.outs)
 
     pred = x -> !(x in state.validated_platforms)
+
     for platform in filter(pred, state.platforms)
         print(state.outs, "Building $platform ")
         build_path = tempname()
         mkpath(build_path)
         local ok = true
+        local skip = false
         cd(build_path) do
             prefix, ur = setup_workspace(
                 build_path,
@@ -503,7 +505,7 @@ function step5c(state::WizardState)
                 verbose=false,
                 tee_stream=state.outs
             )
-
+            prefix == nothing && (skip = false; return)
             run(ur,
                 `/bin/bash -c $(state.history)`,
                 joinpath(build_path,"out.log");
@@ -526,6 +528,10 @@ function step5c(state::WizardState)
                 state.files;
                 silent = true
             ))
+        end
+        if skip
+            warn("Falure with build $platform")
+            continue
         end
         print(state.outs, "[")
         if ok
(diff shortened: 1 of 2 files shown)


src/wizard/interactive_build.jl
```diff
@@ -503,7 +505,7 @@ function step5c(state::WizardState)
                 verbose=false,
                 tee_stream=state.outs
             )
-
+            prefix == nothing && (skip = false; return)
```
Keno: `skip = true`?

SimonDanisch: oh... I guess the return was enough to skip then, since it was actually working for me!

Keno: I think `ok = false` is also appropriate here, rather than adding a new `skip` flag.

SimonDanisch: That'd be great - I was expecting setting ok to false to interrupt the build... If that _officially_ skips the build, then this could almost be a good solution ;)

SimonDanisch: If that actually goes into a mergeable direction, we should also improve the try catch!

src/wizard/interactive_build.jl
```diff
@@ -527,6 +528,10 @@ function step5c(state::WizardState)
                 silent = true
             ))
         end
+        if skip
```
staticfloat: I don't see where `skip` is declared?

src/wizard/interactive_build.jl
```diff
@@ -503,7 +504,7 @@ function step5c(state::WizardState)
                 verbose=false,
                 tee_stream=state.outs
             )
-
+            prefix == nothing && (ok = false; return)
```
staticfloat: I don't really like this syntax; please use the more explicit

```
if prefix == nothing
    ok = false
    return
end
```

It also seems to me that you mean `continue` here, not `return`.

SimonDanisch: I believe you can't continue from inside a closure?

staticfloat: Oh, of course, you're right.  I missed the `cd() do`

src/wizard/interactive_build.jl
```diff
@@ -503,7 +504,10 @@ function step5c(state::WizardState)
                 verbose=false,
                 tee_stream=state.outs
             )
-
+            if prefix == nothing
```
staticfloat: It would also be nice to have a comment here describing what `prefix == nothing` means.

staticfloat:
Closing this as JLL packages have this "silently skip installation" thing down pat.

PR #330, opened 2018-08-20
audit should check all dynamic links

felipenoris:
I was debuging a [builder](https://github.com/felipenoris/mongo-c-driver-builder) for libmongoc library and got stuck on a dyn lib that depends on another dyn lib that is built from the same source tar.

It turns out you already solved it at branch bb8. What I'm suggesting is to merge this to master as a bugfix for current version.


staticfloat:
I just merged the `sf/bb8` branch, so I'm going to close this.  Thanks!

PR #335, opened 2018-08-29
[WIP] adds customizable url prefix to `build_tarballs`

ssfrr:
needs tests etc., but I figured I'd wait until after the bb8 stuff is merged

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -5,7 +5,7 @@ import SHA: sha256
 """
     build_tarballs(ARGS, src_name, src_version, sources, script, platforms,
                    products, dependencies;
-                   deploy_prefix="https://github.com/$(repo)/releases/download/$(tag)")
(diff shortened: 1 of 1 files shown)


src/AutoBuild.jl
```diff
@@ -4,7 +4,8 @@ import SHA: sha256
 
 """
     build_tarballs(ARGS, src_name, src_version, sources, script, platforms,
-                   products, dependencies)
+                   products, dependencies;
+                   deploy_prefix="https://github.com/$(repo)/releases/download/$(tag)")
```
ararslan: Since this is in a docstring, I think you'll need to escape the `$`s.

staticfloat:
Closing this as time has flowed past us all.  If I'm understanding correctly, I think `--deploy=MyOrg/MyPackage_jll` will do what you were trying to do here.

PR #339, opened 2018-08-30
write_deps_file stubs now accept keyword args

ssfrr:
the real function accepts kwargs, but these stubs did not, so if the
user's dependent build.jl used a keyword argument it would crash out
while setting up the workspace


staticfloat:
Very good.

PR #341, opened 2018-08-31
[WIP] adds `url_hash` convenience function

ssfrr:
Talked to @SimonDanisch about this in slack - seems like it's a thing people want to be able to do. I'm not 100% sure whether this belongs more in BinaryBuilder or BinaryProvider, but I figured this makes sense because you usually will need it in the context of writing a `build_tarballs.jl`

### DIFF
--- a/src/BinaryBuilder.jl
+++ b/src/BinaryBuilder.jl
@@ -9,6 +9,7 @@ using Nullables
 using GitHub
 
 @reexport using BinaryProvider
+export url_hash
 
 include("compat.jl")
 include("Auditor.jl")
@@ -206,6 +207,13 @@ function versioninfo()
             platform=Linux(:x86_64),
         )
         run_interactive(runner, `/usr/local/bin/ccache -s`)
+    end
+end
+
+function url_hash(url::AbstractString)
+    path = download(url)
+    open(path) do io
+        bytes2hex(BinaryProvider.sha256(io))
     end
 end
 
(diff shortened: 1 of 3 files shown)


src/BinaryBuilder.jl
```diff
@@ -209,4 +210,11 @@ function versioninfo()
     end
 end
 
+function url_hash(url::AbstractString)
+    path = download(url)
```
KristofferC: Should clean up the downloaded file? Or is the side effect intended?

Another way of writing this is to just have it as `hash(file)` and have users call it with e.g. `hash(download(...))`.

ssfrr: oh yeah, it should clean up the downloaded file.

I think the common use-case is you have a URL you want to add and you need the hash, so I prefer the API as-is.

It might be nice to tie into the other download machinery in BB that shows the nice progress bar though. I just used this on a large-ish tarball and it just sits there for a bit while it downloads, which could be surprising for a user who's not really thinking about what's going on under the hood.

staticfloat: You can have it be auto-cleaned up by downloading within a temporary prefix (which will get auto-deleted after you exit the `do` block).  You can also use `BinaryProvider.download(url, path; verbose=true)` to get the neat download bar.  Putting that together, what I'm suggesting looks like this:

```
function url_hash(url::AbstractString)
    temp_prefix() do p
        local_path = joinpath(p, basename(url))
        BinaryProvider.download(url, local_path; verbose=true)
        return hex2bytes(sha256(open(local_path)))
    end
end
```

staticfloat:
Thanks for the PR, we have an FAQ entry about this now.

Issue #340, opened 2018-08-31
Wizard doesn't find .gitconfig info


ssfrr:
At the end of running the wizard it said it couldn't find my `~/.gitconfig`, but it definitely exists. Possibly a `LibGit2` issue?

PR #344, opened 2018-09-01
`LibGit2.get(cfg, key, default)` doesn't mix types

staticfloat:
Write our own version to do what we want.  Also fix logic problems.  This should fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/340

### DIFF
--- a/src/wizard/deploy.jl
+++ b/src/wizard/deploy.jl
@@ -306,11 +306,20 @@ function common_git_repo_setup(repo_dir, repo, state)
         break
     end
 
+    # Helper function to get around LibGit2's strange `get()` behavior
+    function lg2get(T, cfg, name)
+        try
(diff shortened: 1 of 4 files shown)


src/wizard/deploy.jl
```diff
@@ -306,11 +306,20 @@ function common_git_repo_setup(repo_dir, repo, state)
         break
     end
 
+    # Helper function to get around LibGit2's strange `get()` behavior
+    function lg2get(T, cfg, name)
+        try
+            return LibGit2.get(T, cfg, name)
+        catch
```
KristofferC: Perhaps worth checking the type of the exception or something? In case someone touches this later and introduces an `UndefVarError`.

omus:
Would be good to make an issue against  the LibGit2 stdlib for this

PR #350, opened 2018-09-14
sf/newrootfs: Fix UserNSRunner

Keno:


staticfloat:
Subsumed by other updates

PR #390, opened 2018-11-19
fix libgcc_s issue

iblis17:
cc @staticfloat, @bicycle1885
This can fix https://github.com/bicycle1885/ZlibBuilder/issues/11

with this patch
```
tar tf Zlib.v1.2.11.x86_64-unknown-freebsd11.1.tar.gz                                                                        
./
./include/
./include/zconf.h
./include/zlib.h
./share/
./share/man/
./share/man/man3/
./share/man/man3/zlib.3
./logs/
./logs/Zlib.log
./lib/
./lib/libz.so.1
./lib/libz.so
./lib/libz.so.1.2.11
./lib/libz.a
./lib/pkgconfig/
./lib/pkgconfig/zlib.pc
```

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -131,6 +131,7 @@ function platform_envs(platform::Platform, host_target="x86_64-linux-gnu")
         # the location of libgcc_s.  LE SIGH.
         # https://github.com/llvm-mirror/clang/blob/f3b7928366f63b51ffc97e74f8afcff497c57e8d/lib/Driver/ToolChains/FreeBSD.cpp
         if occursin("-freebsd", target)
+            mapping["LDFLAGS"] *= " -L/opt/$(target)/$(target)/lib"
             mapping["CXX"] *= " -L/opt/$(target)/$(target)/lib"
             mapping["CC"]  *= " -L/opt/$(target)/$(target)/lib"
         end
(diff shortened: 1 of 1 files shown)


iblis17:
maybe related: https://github.com/JuliaWeb/LibCURLBuilder/pull/11

staticfloat:
I'm really not a fan of pushing `-L` flags into compiler names.  We should figure out why `zlib`'s build process isn't using `LDFLAGS` on FreeBSD (but is on other platforms?!) and instead fix the build process.

iblis17:
I'm fine with this style, since libcurl doesn't respect to the `LDFLAGS`, also. (and maybe more projects are in this situation).

staticfloat:
> since libcurl doesn't respect to the LDFLAGS, either.

We should patch the `Makefile` then.  There are many tools that break when `CC` has flags embedded within it.

giordano:
I'm also frequently hit by this issue (latest case: https://github.com/JuliaPackaging/Yggdrasil/pull/38), but I agree with Elliot that it's not good style to add `-L/path` to the compiler name.  The Makefile must properly use compilation flags anyway (and it would be great to push the fix to upstream :wink:)

giordano:
I think this can be closed, now a hack like this is not even needed.

staticfloat:
Agreed. Woohoo!

PR #400, opened 2018-12-18
WebAssembly platform

tshort:
This is to support a new platform with PR [here](https://github.com/JuliaPackaging/Yggdrasil/pull/2).

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -139,9 +139,11 @@ analyze the name and platform of this shard and return a path based on that.
 function map_target(cs::CompilerShard)
     if lowercase(cs.name) == "rootfs"
         return "/"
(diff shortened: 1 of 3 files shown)


aminya:
@tshort What does this PR do?

tshort:
The main work is at https://github.com/JuliaPackaging/Yggdrasil/pull/2. This is an experiment to add WebAssembly output as another binary target. It adds Emscripten to do this. These early experiments were promising. I got one or two C libraries compiled to WebAssembly that way. 

This particular PR is for the extra bits to add necessary shards to BB. It's all way out of date now.

staticfloat:
Thanks for this Tom, but I don't think this is going to happen anytime soon, as the wasm work for Julia itself has all but stopped, from what I understand.

PR #414, opened 2019-04-03
Misc documentation updates to make the docs build.

fredrikekre:
Docs has not been building for quite some time, this should fix that.


.travis.yml
```diff
@@ -49,6 +49,14 @@ jobs:
       env:
         - BINARYBUILDER_RUNNER=docker
 
+    - stage: Documentation
+      julia: 1.0
+      os: linux
+      script:
+        - julia --project=docs -e 'using Pkg; Pkg.instantiate()'
```
ararslan:
```suggestion
        - julia --project=docs -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
```
?

fredrikekre: Nah, https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/414/files#diff-17fe722019cf21306d74393750d3267aR6

ararslan: Ew, checking in a manifest 😦

staticfloat: Checking in a manifest is good practice; it allows users to know a single point in package version space that at least worked on one person's machine, and allows CI to avoid breaking your package due to upstream changes.  Ideally, we would have CI that does both "use `Manifest`"-style and "install it like I just did `]add`"-style, but we're not quite there yet.  :)

fredrikekre: This is `docs/Manifest.toml` though, no reason not to lock doc-build dependencies.

PR #418, opened 2019-05-10
Two small fixes for running Docker on Windows

visr:
With this, and `BINARYBUILDER=docker`, I can successfully run `BinaryBuilder.runshell` and could also run the SQLiteBuilder `build_tarballs.jl`.

More details in the commit messages.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -329,7 +329,7 @@ function autobuild(dir::AbstractString,
                         repo = LibGit2.clone(src_url, src_path; isbare=true)
                     end
                 else
-                    if isfile(src_url)
+                    if BinaryProvider.safe_isfile(src_url)
                         # Immediately abspath() a src_url so we don't lose track of
                         # sources given to us with a relative path
                         src_path = abspath(src_url)
--- a/src/DockerRunner.jl
+++ b/src/DockerRunner.jl
@@ -78,7 +78,7 @@ function DockerRunner(workspace_root::String;
     docker_cmd = `docker run --privileged `#--cap-add SYS_ADMIN`
 
     if cwd != nothing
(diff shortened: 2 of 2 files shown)


visr:
And interestingly I found that the line continuation in the script also does not work on Docker for Windows:
```bash
# ./configure --prefix=$prefix --host=$target \
#     --with-geos=$prefix/bin/geos-config \
#     --with-proj=$prefix \
#     --with-libz=$prefix \
#     --with-sqlite3=$prefix \
#     --with-curl=$prefix/bin/curl-config \
#     --with-python=no \
#     --enable-shared \
#     --disable-static \
#     "CC=$CC" \
#     "CXX=$CXX"

./configure --prefix=$prefix --host=$target --with-geos=$prefix/bin/geos-config --with-proj=$prefix --with-libz=$prefix --with-sqlite3=$prefix --with-curl=$prefix/bin/curl-config --with-python=no --enable-shared --disable-static "CC=$CC" "CXX=$CXX"
```

visr:
> Does this mean that your pull request is not valid anymore?

No, because this snippet comes from the [build_tarballs.jl](https://github.com/JuliaGeo/GDALBuilder/blob/afb7e5fbeb879fd8829a295df31b44df263db865/build_tarballs.jl#L35-L45), which lives outside the package.

This probably should be opened as a separate issue though. Couldn't quickly find a fix.

staticfloat:
This is fascinating @visr!  I'm in the middle of a giant overhaul to BB, but I will add this to [my TODO](https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/441) to merge into that branch before merging my big PR.  Thanks for the POC!

visr:
Closing since the Docker path commit is now on master, and the `isfile` no longer throws on Julia 1.3 due to https://github.com/JuliaLang/julia/pull/31999.

PR #419, opened 2019-05-20
add `:gcc9` to `gcc_versions` (to help OpenspecfunBuilder)

JeffreySarnoff:


JeffreySarnoff:
here is the problem https://julialang.slack.com/archives/C6A044SQH/p1558370479426600

JeffreySarnoff:
I have no idea why this fails.

staticfloat:
This isn't quite the right thing to do; by putting this into `BinaryBuilder`; you are asserting that there exists a `gcc9` shard with GCC 9 compilers within it that can be used to compile things; this is not true and likely will never be true; since GCC 8 and GCC 9 are very compatible, anything compiled with GCC 8 should be runnable with a system based on GCC 9, so this should not be necessary.  I am helping the original user in that slack conversation link work through their issues, which I think are slightly more complex than initially believed.

PR #420, opened 2019-05-27
Improve documentation

giordano:
There are some interesting tricks that I learnt on the Slack, but given its ephemeral nature I'd like to make this information more persistent, and useful for other people too.

### DIFF
--- a/docs/src/FAQ.md
+++ b/docs/src/FAQ.md
@@ -30,7 +30,7 @@ Some linux distributions have a bug in their `overlayfs` implementation that pre
 
 ### I have to build a very small project without a Makefile, what do I have to do?
 
-What BinaryBuilder needs is to find the relevant file (shared libraries, or executables, etc...) organised under the `$prefix` directory: libraries should go to `$prefix/lib`, executables to `$prefix/bin`.  You may need to create those directory.  You are free to choose whether to create a simple Makefile to build the project or to do everything within the `build_tarballs.jl` script.
+What BinaryBuilder needs is to find the relevant file (shared libraries, or executables, etc...) organised under the `$prefix` directory: libraries should go to `$prefix/lib` (`$prefix/bin` on Windows), executables to `$prefix/bin`.  You may need to create those directory.  You are free to choose whether to create a simple Makefile to build the project or to do everything within the `build_tarballs.jl` script.
 When the script completes, BinaryBuilder expects to find at least one artifact _built for the expected architecture_ in either `$prefix/lib` or `$prefix/bin`.
 Remember also that you should use the standard environment variables like `CC`, `CXX`, `CFLAGS`, `LDFLAGS` as appropriate in order to cross compile.  See [the list of variables](environment_variables.md).
 
--- a/docs/src/FAQ.md
+++ b/docs/src/FAQ.md
@@ -32,6 +32,8 @@ Some linux distributions have a bug in their `overlayfs` implementation that pre
 
 What BinaryBuilder needs is to find the relevant file (shared libraries, or executables, etc...) organised under the `$prefix` directory: libraries should go to `$prefix/lib`, executables to `$prefix/bin`.  You may need to create those directory.  You are free to choose whether to create a simple Makefile to build the project or to do everything within the `build_tarballs.jl` script.
 
+Remember also that you should use the standard environment variables like `CC`, `CXX`, `CFLAGS`, `LDFLAGS` as appropriate in order to cross compile.  See [the list of variables](environment_variables.md).
+
 ### Can I run the build script only for one or a few targets?
 
 Yes.  The `build_tarballs.jl` script can be used as a command line utility, it takes a few options and as argument the list of triplets of the targets.  You can find more information about the syntax of the script with
@@ -43,7 +45,6 @@ For example, with
 julia --compile=min --color=yes build_tarballs.jl --debug --verbose aarch64-linux-musl,arm-linux-musleabihf
 ```
 you can run the build script only for `aarch64-linux-musl` and `arm-linux-musleabihf` targets.  The `--debug` option causes a failed build to drop into an interactive bash shell for debugging purposes.
-
 
 ### Can I open a shell in a particular build environment for doing some quick tests?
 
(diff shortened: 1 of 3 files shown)


docs/src/FAQ.md
```diff
@@ -26,4 +26,34 @@ At the time of writing, we support Linux (x86_64, i686, armv7l, aarch64, ppc64le
...
+Some linux distributions have a bug in their `overlayfs` implementation that prevents us from mounting overlay filesystems within user namespaces.  See [this Ubuntu kernel bug report](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1531747) for a description of the situation and how Ubuntu has patched it in their kernels.  To work around this, you can launch `BinaryBuilder.jl` in "privileged container" mode.  Unfortunately, this involves running `sudo` every time you launch into a BinaryBuilder session, but on the other hand, this successfully works around the issue on distributions such as Arch linux.  To set "privileged container" mode, set the `BINARYBUILDER_RUNNER` environment variable to `privileged`.
+
+### I have to build a very small project without a Makefile, what do I have to do?
+
+What BinaryBuilder needs is to find the relevant file (shared libraries, or executables, etc...) organised under the `$prefix` directory: libraries should go to `$prefix/lib`, executables to `$prefix/bin`.  You may need to create those directory.  You are free to choose whether to create a simple Makefile to build the project or to do everything within the `build_tarballs.jl` script.
+
```
rai-nhdaly: Per Stefan's request on Slack, I was surprised when I _did_ put things into the bin directory but it still complained. Could we maybe also add a line about what is expected after running the script?

maybe something like:
```suggestion
When the script completes, BinaryBuilder expects to find at least one artifact _built for the expected architecture_ in either `$prefix/lib` or `$prefix/bin`.
```
?

docs/src/build_tips.md
```diff
@@ -120,6 +120,29 @@ There are plans to handle file changes in the wizard automatically [(#25)](https
...
+```sh
+if [[ "${target}" == *-freebsd* ]] || [[ "${target}" == *-apple-* ]]; then
+    CC=/opt/${target}/bin/${target}-gcc
+    CXX=/opt/${target}/bin/${target}-g++
+    FC=/opt/${target}/bin/${target}-gfortran
+    LD=/opt/${target}/bin/${target}-ld
+    AR=/opt/${target}/bin/${target}-ar
+    AS=/opt/${target}/bin/${target}-as
+    NM=/opt/${target}/bin/${target}-nm
+    OBJDUMP=/opt/${target}/bin/${target}-objdump
```
staticfloat: Now that we have JIT compiler wrappers, we can just tell people to use `${target}-gcc` or even `gcc` if they want.  :)

docs/src/environment_variables.md
```diff
@@ -13,3 +13,27 @@
 * `BINARYBUILDER_ALLOW_ECRYPTFS`: When set to `true`, this allows the mounting of rootfs/shard/workspace directories from within encrypted mounts.  This is disabled by default, as at the time of writing, this triggers kernel bugs.  To avoid these kernel bugs on a system where e.g. the home directory has been encrypted, set the `BINARYBUILDER_ROOTFS_DIR` and `BINARYBUILDER_SHARDS_DIR` environment variables to a path outside of the encrypted home directory.
 
 * `BINARYBUILDER_USE_CCACHE`: When set to `true`, this causes a `/root/.ccache` volume to be mounted within the build environment, and for the `CC`, `CXX` and `FC` environment variables to have `ccache` prepended to them.  This can significantly accelerate rebuilds of the same package on the same host.  Note that `ccache` will, by default, store 5G of cached data.
+
+The next environment variables are automatically set in the build environment and should be used to build the project.  Occasionally, you may need to tweak them (e.g., if you need to use GCC on macOS or FreeBSD, see the [build tips](build_tips.md))
```
staticfloat: Let's put a section heading here, like `# Environment Variables within the build environment`

staticfloat: On second thought, perhaps this whole section belongs within `build_tips.md`.

docs/src/build_tips.md
giordano: So I guess we only need to set
```
CC=gcc
CXX=g++
```
and stop?

docs/src/build_tips.md
```diff
@@ -79,4 +105,52 @@ Examples of builders that depend on other binaries include:
...
+* `nproc`: the number of processors of the host machine, useful for parallel building (e.g., `make -j${nproc}`)
+* `nbits`: number of bits of the target architecture (usually it is either 32 or 64)
+* `proc_family`: target processor family (e.g., "intel", "power", or "arm")
+* `dlext`: extension of the shared library on the target system.  It is "dll" for Windows, "dylib" for macOS, and "so" for the other Unix systems.
+
+## Using GCC on macOS and FreeBSD
+
+For these target systems Clang is the default compiler, however some programs may not be compatible with Clang.
+
+For programs built with CMake (see the [CMake build](#CMake-builds-1) section) you can use the GCC toolchain file that is in `/opt/${target}/${target}_gcc.toolchain`.
```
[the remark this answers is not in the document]
staticfloat: It's now at `/opt/${target}/${target}_gcc.cmake`, so we could change it to say that, or we could tell them to use `${CMAKE_TARGET_TOOLCHAIN%.*}_gcc.cmake`.  :P

docs/src/FAQ.md
```diff
@@ -34,26 +34,47 @@ What BinaryBuilder needs is to find the relevant file (shared libraries, or exec
...
 
 ### Can I install packages in the build environment?
 
 Yes, but it's unlikely that you'll need to.  The build environment is based on Alpine Linux (triplet: `x86_64-linux-musl`) so you can use [`apk`](https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management) to install packages in it.  However, if you need runtime libraries or programs for the target system these packages won't help you.  The package manager is useful only to install utilities, tools or libraries that are needed exclusively at compile time on the build system.
+
+### What are those numbers in the list of sources?  How do I get them?
+
+The list source of sources is a vector of dictinaries, whose key is the URL of the source and whose value is its hash.  What kind of hash depends on what the source actually is:
+
+* If the source is a file to be downloaded, the hash is a 64-character SHA256 checksum.  If you have a copy of that file, you can compute the hash in Julia with
```
staticfloat: Can you add an example of using `curl -L $url | shasum -a 256` as well?

giordano: I use this a lot (actually I use `wget` and `sha256sum`), but I was a bit wary about mentioning this as users may not have those tools actually available (Windows, I'm looking at you), but yeah, let's share all tips to make user's life easier!

staticfloat:
Great stuff Mosè!

PR #421, opened 2019-05-28
Optional excluding of Products from supported_platforms()

ianshmean:
Provides exclusion within supported_platforms() to move to an opt-out style BinaryBuilder approach. 

```julia
>supported_platforms(exclude=[Windows(:i686),Windows(:x86_64)])
11-element Array{Platform,1}:
 Linux(:i686, libc=:glibc)                    
 Linux(:x86_64, libc=:glibc)                  
 Linux(:aarch64, libc=:glibc)                 
 Linux(:armv7l, libc=:glibc, call_abi=:eabihf)
 Linux(:powerpc64le, libc=:glibc)             
 Linux(:i686, libc=:musl)                     
 Linux(:x86_64, libc=:musl)                   
 Linux(:aarch64, libc=:musl)                  
 Linux(:armv7l, libc=:musl, call_abi=:eabihf) 
 MacOS(:x86_64)                               
 FreeBSD(:x86_64)  
```

A couple of reasons why this might be nice:

1. The current approach tends to result in manual lists of platforms, with unsupported being left-off, meaning future additions to BB's supported_platforms require more effort to be added
2. With the _manual list & leave out excluded_ approach, it's not always  easy to figure out which platforms have been excluded (if they're not simply commented out) 

cc. @staticfloat, as discussed over in https://github.com/JuliaIO/OggBuilder/pull/2#issuecomment-496617866

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -400,7 +400,7 @@ represented here, it's probably because that platform is still considered "in be
 Platforms can be excluded from the list by specifying an array of platforms to `exclude`.
 i.e. `supported_platforms(exclude=[Windows(:i686),Windows(:x86_64)])`
 """
-function supported_platforms(;exclude::Union{Vector{<:Platform},Function}=[])
+function supported_platforms(;exclude::Union{Vector{<:Platform},Function}=x->false)
     exclude_platforms!(exclude::Function, platforms) = filter(!exclude, platforms)
     function exclude_platforms!(exclude::Vector{<:Platform}, platforms)
     	for e in exclude
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -406,13 +406,8 @@ supported_platforms(exclude=islin)
 ```
 """
 function supported_platforms(;exclude::Union{Vector{<:Platform},Function}=x->false)
-    exclude_platforms!(exclude::Function, platforms) = filter(!exclude, platforms)
-    function exclude_platforms!(exclude::Vector{<:Platform}, platforms)
-    	for e in exclude
-    		filter!(x -> x != e, platforms)
-    	end
-    	return platforms
-    end
+    exclude_platforms!(platforms, exclude::Function) = filter(!exclude, platforms)
+    exclude_platforms!(platforms, exclude::Vector{<:Platform}) = filter!(!in(exclude), platforms)
     standard_platforms = [
         # glibc Linuces
         Linux(:i686),
@@ -435,7 +430,7 @@ function supported_platforms(;exclude::Union{Vector{<:Platform},Function}=x->fal
         Windows(:i686),
         Windows(:x86_64),
     ]
-    return exclude_platforms!(exclude,standard_platforms)
+    return exclude_platforms!(standard_platforms,exclude)
 end
 
 function replace_gcc_version(p::Platform, gcc_version::Symbol)
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -391,14 +391,19 @@ function choose_shards(p::Platform;
 end
 
 """
-    supported_platforms(;exclude::Vector=[])
+    supported_platforms(;exclude::Union{Vector{<:Platform},Function}=x->false)
 
 Return the list of supported platforms as an array of `Platform`s.  These are the platforms we
 officially support building for, if you see a mapping in `get_shard_hash()` that isn't
 represented here, it's probably because that platform is still considered "in beta".
 
-Platforms can be excluded from the list by specifying an array of platforms to `exclude`.
-i.e. `supported_platforms(exclude=[Windows(:i686),Windows(:x86_64)])`
+Platforms can be excluded from the list by specifying an array of platforms to `exclude` i.e.
+`supported_platforms(exclude=[Windows(:i686),Windows(:x86_64)])`
+or a function that returns true for exclusions i.e.
+```
+islin(x) = typeof(x) == Linux
+supported_platforms(exclude=islin)
+```
 """
 function supported_platforms(;exclude::Union{Vector{<:Platform},Function}=x->false)
     exclude_platforms!(exclude::Function, platforms) = filter(!exclude, platforms)
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -400,10 +400,15 @@ represented here, it's probably because that platform is still considered "in be
 Platforms can be excluded from the list by specifying an array of platforms to `exclude`.
 i.e. `supported_platforms(exclude=[Windows(:i686),Windows(:x86_64)])`
 """
-function supported_platforms(;exclude::Vector=[])
-    platforms = [
-        #Standard Platforms
-
+function supported_platforms(;exclude::Union{Vector{<:Platform},Function}=[])
+    exclude_platforms!(exclude::Function, platforms) = filter(!exclude, platforms)
+    function exclude_platforms!(exclude::Vector{<:Platform}, platforms)
+    	for e in exclude
+    		filter!(x -> x != e, platforms)
+    	end
+    	return platforms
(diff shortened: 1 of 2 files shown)


src/Rootfs.jl
```diff
@@ -391,14 +391,19 @@ function choose_shards(p::Platform;
...
 Return the list of supported platforms as an array of `Platform`s.  These are the platforms we
 officially support building for, if you see a mapping in `get_shard_hash()` that isn't
 represented here, it's probably because that platform is still considered "in beta".
+
+Platforms can be excluded from the list by specifying an array of platforms to `exclude`.
+i.e. `supported_platforms(exclude=[Windows(:i686),Windows(:x86_64)])`
 """
-function supported_platforms()
-    return [
+function supported_platforms(;exclude::Vector=[])
```
staticfloat: Let's have `exclude` be a `Union{Vector{<:Platform},Function}`, then have an inner method definition that does something like:

```
exclude_platforms!(exclude::Function, platforms) = filter(exclude, platforms)
function exclude_platforms!(exclude::Vector{<:Platform}, platforms)
	for e in exclude
		filter!(x -> x != v, platforms)
	end
	return platforms
end
```

staticfloat: That will make it easy to do things like turn off all `MacOS` builds, even if an armv7l one appears in the future.  :)

staticfloat:
Excellent.  Please add two tests, ensuring that this works and then I think we're good to go.  Don't worry about the currently failing tests; they're pretty late in the process, so your tests should run.

src/Rootfs.jl
ianshmean: Yeah, makes sense. Added

ianshmean:
```julia
>supported_platforms(exclude=[Windows(:i686),Windows(:x86_64)])
11-element Array{Platform,1}:
 Linux(:i686, libc=:glibc)                    
 Linux(:x86_64, libc=:glibc)                  
 Linux(:aarch64, libc=:glibc)                 
 Linux(:armv7l, libc=:glibc, call_abi=:eabihf)
 Linux(:powerpc64le, libc=:glibc)             
 Linux(:i686, libc=:musl)                     
 Linux(:x86_64, libc=:musl)                   
 Linux(:aarch64, libc=:musl)                  
 Linux(:armv7l, libc=:musl, call_abi=:eabihf) 
 MacOS(:x86_64)                               
 FreeBSD(:x86_64) 
```

```julia
>iswindows(x) = typeof(x) == Windows
>supported_platforms(exclude=iswindows)
11-element Array{Platform,1}:
 Linux(:i686, libc=:glibc)                    
 Linux(:x86_64, libc=:glibc)                  
 Linux(:aarch64, libc=:glibc)                 
 Linux(:armv7l, libc=:glibc, call_abi=:eabihf)
 Linux(:powerpc64le, libc=:glibc)             
 Linux(:i686, libc=:musl)                     
 Linux(:x86_64, libc=:musl)                   
 Linux(:aarch64, libc=:musl)                  
 Linux(:armv7l, libc=:musl, call_abi=:eabihf) 
 MacOS(:x86_64)                               
 FreeBSD(:x86_64) 
```

src/Rootfs.jl
```diff
@@ -400,10 +400,15 @@ represented here, it's probably because that platform is still considered "in be
...
 i.e. `supported_platforms(exclude=[Windows(:i686),Windows(:x86_64)])`
 """
-function supported_platforms(;exclude::Vector=[])
-    platforms = [
-        #Standard Platforms
-
+function supported_platforms(;exclude::Union{Vector{<:Platform},Function}=[])
+    exclude_platforms!(exclude::Function, platforms) = filter(!exclude, platforms)
+    function exclude_platforms!(exclude::Vector{<:Platform}, platforms)
+    	for e in exclude
```
ararslan: Intendation is messed up, should be 4 space indents, whereas this is 4 spaces and a tab. Same for the lines below.

src/Rootfs.jl
```diff
@@ -420,6 +430,7 @@ function supported_platforms()
         Windows(:i686),
         Windows(:x86_64),
     ]
+    return exclude_platforms!(exclude,standard_platforms)
```
ararslan:
```suggestion
    return exclude_platforms!(exclude, standard_platforms)
```

src/Rootfs.jl
```diff
@@ -391,14 +391,24 @@ function choose_shards(p::Platform;
...
 represented here, it's probably because that platform is still considered "in beta".
+
+Platforms can be excluded from the list by specifying an array of platforms to `exclude`.
+i.e. `supported_platforms(exclude=[Windows(:i686),Windows(:x86_64)])`
 """
-function supported_platforms()
-    return [
+function supported_platforms(;exclude::Union{Vector{<:Platform},Function}=[])
+    exclude_platforms!(exclude::Function, platforms) = filter(!exclude, platforms)
+    function exclude_platforms!(exclude::Vector{<:Platform}, platforms)
```
ararslan: Conventionally, the modified argument goes first.
```suggestion
    function exclude_platforms!(platforms, exclude::Vector{<:Platform})
```

ararslan: Actually, I think this entire thing can just be
```suggestion
    exclude_platforms!(platforms, exclude::Vector{<:Platform}) = filter!(!in(exclude), platforms)
```

ianshmean:
Added tests and adjusted based on review 👍🏻

test/basic.jl
```diff
@@ -19,6 +19,17 @@
...
 
+@testset "Supported Platforms" begin
+    all = supported_platforms()
+    opt_out_specific = supported_platforms(exclude=[Linux(:x86_64, libc=:glibc)])
+    islin(x) = typeof(x) == Linux
+    opt_out_fx = supported_platforms(exclude=islin)
+
+    @test length(all) == length(opt_out_specific)+1
+    @test !any(opt_out_specific .== [Linux(:x86_64, libc=:glibc)])
+    @test !any(opt_out_fx .== [Linux(:x86_64, libc=:glibc)])
```
staticfloat: Just being paranoid here, but please add the following tests, so that it's obvious to readers from the future how `opt_out_fx` and `opt_out_specific` differ:

```
@test any(opt_out_specific .== [Linux(:i686 , libc=:glibc)])
@test !any(opt_out_fx .== [Linux(:i686 , libc=:glibc)])
```

src/Rootfs.jl
```diff
@@ -420,6 +430,7 @@ function supported_platforms()
         Windows(:i686),
         Windows(:x86_64),
     ]
```
staticfloat: There was a part of me that had a knee-jerk reaction to this list of platforms and almost commented "Just use `supported_platforms()` instead of listing these out like this".  lol.

ianshmean: haha! Hopefully with this PR those pains are reduced!

test/basic.jl
ianshmean: I actually inverted the logic on your suggestion, such that if the function returns true, it's excluded. I think it makes more sense given the keyword is `exclude`.

Should've mentioned that, apols.

staticfloat: Right, but my request still stands.  `opt_out_specific` will only exclude `Linux(:x86_64, :glibc)`, but will still include other `Linux` platforms, for example, `Linux(:i686, :glibc)` whereas `opt_out_fx` will have all `Linux` platforms excluded.  I just want tests here showing that explicit difference, so that we don't break it in the future and so that someone reading this sees how the two `opt_out_*` lists differ.

ianshmean: ah! I missed that you moved to `:i686`. Just updated

ianshmean:
Just came across an instance of this being useful! Is it ready to merge? @staticfloat?

PR #423, opened 2019-05-31
Add $prefix/share/pkgconfig to PKG_CONFIG_PATH

giordano:
See https://github.com/bicycle1885/ZlibBuilder/pull/14#issuecomment-497628609


staticfloat:
Yeah, that is a good idea.  I didn't know newer versions had shifted to use `share` instead of `lib`.  Good find!

giordano:
> I didn't know newer versions had shifted to use `share` instead of `lib`.

It seems to be "in addition to", rather than "instead of"

PR #438, opened 2019-08-26
Define OBJC environment variable

### LINKED ISSUE #426: Define OBJC environment variable
Sometimes we need to explicitly set the Objective-C compiler (e.g., see https://github.com/JuliaIO/FFMPEGBuilder/pull/5), it'd be useful to have the `OBJC` environment variable already set.

giordano:
The variable will most probably be used only when building for macOS, but having it defined for consistency on all platforms shouldn't hurt.  What do you think?

Fix #426.


giordano:
This is already in #441.

PR #445, opened 2019-09-18
Add UUID to dependencies of jll packages

giordano:
I couldn't test it thoroughly, but should address https://github.com/JuliaBinaryWrappers/OpenSpecFun_jll.jl/issues/1


fredrikekre:
Should already be fixed by 3215f18ea9bdab5515fde3ebd5a6db70a64e26e0

staticfloat:
Yeah, I already fixed this by having it explicitly import `Base.UUID`.  Is that better? I don't know.  But it works.  I just need to re-build OpenSpecFun_jll, which I have done here: https://github.com/JuliaRegistries/General/pull/3647
Thanks Mosè!

PR #453, opened 2019-10-14
WIP: Allow dlopen flags (e.g. RTLD_GLOBAL) for LibraryProduct

colinxs:
@staticfloat @giordano I haven't tested this extensively, but here's a sketch of what we were talking about earlier. 

Right now the flags are just used when generating the wrapper. Should they also be used in `locate(...)` and `Auditor.jl`?

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -861,7 +861,7 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
                     println(io, """
                         # Manually `dlopen()` this right now so that future invocations
                         # of `ccall` with its `SONAME` will find this path immediately.
-                        global $(vp)_handle = dlopen($(vp)_path)
+                        global $(vp)_handle = $(dlopen_str("$(vp)_path", p.dlopen_flags))
                         push!(LIBPATH_list, dirname($(vp)_path))
                     """)
                 elseif p isa ExecutableProduct
--- a/src/Products.jl
+++ b/src/Products.jl
@@ -78,7 +78,8 @@ struct LibraryProduct <: Product
     libnames::Vector{String}
     variable_name::Symbol
     dir_paths::Vector{String}
-    dont_dlopen::Bool
+    try_dlopen::Bool
+    dlopen_flags::Vector{Symbol}
 
     """
         LibraryProduct(libnames, varname::Symbol)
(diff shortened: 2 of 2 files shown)


staticfloat:
Yes, we should use them in `locate()`.

colinxs:
In reference to the example we discussed on slack where `libA` depends on `libglew` being loaded with RTLD_GLOBAL beforehand, should `locate` and `audit` be aware of this dependency chain?

colinxs:
Hmm yeah it's a bit tricky. Currently dependencies are declared as just strings. But for the case we discussed where `Foo_jll` depends on `GLEW_jll`, and requires that `GLEW_jll` be opened with certain flags (e.g. RTLD_GLOBAL), then there's not much we can do with just `LibraryProduct` other than building a `GLEW_RTLD_GLOBAL_JLL` package lol. 

It seems we would need something like:
```julia
dependencies = [
  Dependency("GLEW_jll", options = ...),
  "GLFW_jll" # shorthand for Dependency("GLFW_jll") so as to not break everyone's `build_tarballs.jl`
]
```

Actually _handling_ those options, without `Foo_jll` knowing about the internals of `GLEW_jll`, is a bit tricky. One option is that instead of `*_jll` packages defining `__init__` they would define `load(args...; kwargs)` allowing `Foo_jll` to call `GLEW_jll.load(RTLD_GLOBAL)`. If we wanted, we could special case dependencies without options to just define `__init__` as currently designed (which seems necesary unless we/you want to rebuild all `jll`'s). 

I don't know enough about the internals of `dlopen`, but is there anything wrong with leaving `__init__` and then reloading everything with `GLEW_jll.load(...)`?

giordano:
Today I found that also GSL needs to open a library with `RTLD_GLOBAL`:
https://github.com/JuliaMath/GSL.jl/blob/436eeccd9d47fc353d678931e0c2ce58365a9a1a/src/GSL.jl#L29-L33

staticfloat:
It would be great to add a test case for this; to create a simple C executable that gets compiled to require `RTLD_GLOBAL` with something else, then build it and load it in for the currently-running platform.  Is that something you could do, @colinxs? I realize that may require a bit of learning, so I don't want to burden you with it, but this seems like a feature with a high probability of bit-rot otherwise.

colinxs:
Hey Elliot, sorry for missing this. That's slightly out of my area of expertise, but should be doable.  I won't be able to get to it for a couple weeks, however. But happy to help, especially since I opened this issue :)

Could you elaborate a bit on what exactly we're testing here?

staticfloat:
> That's slightly out of my area of expertise, but should be doable.

This is how expertise gets spread around.  ;)

> Could you elaborate a bit on what exactly we're testing here?

We want to create a binary that requires `RTLD_GLOBAL` to function properly.  We then want to build that binary as part of the BB test suite, try to use it, and fail if we don't open it with `RTLD_GLOBAL`.  We then want to build it again, and have it succeed when it uses `RTLD_GLOBAL`.

I think there are therefore three parts to this PR:

* Build the abstraction that allows for passing flags to `dlopen()` (done, and quite nicely too!)

* Construct a simple C dependency that requires passing flags to `dlopen()` in order to run properly

* Build and test the resultant JLL as a part of the BB test suite, to ensure that this capability never bitrots.

KristofferC:
Just for reference, another library that requires `RTLD_GLOBAL` is apparently `libpardiso` (which is proprietary..) and requires openmp and blas to be opened with `RTLD_GLOBAL` ( or it complains about not finding `omp_get_wtime_` and `dswap_`, respectively).

barche:
We seem to desperately need this for CxxWrap as well: https://github.com/JuliaInterop/libcxxwrap-julia/issues/39#issuecomment-586763157

barche:
I just did a few additional tests with CxxWrap, and it seems there is no way around adding the `RTLD_GLOBAL` flag there as well. It fails in a very subtle way and only sometimes, but here is an example of the kind of crash we get without `RTLD_GLOBAL`:

https://travis-ci.org/JuliaInterop/libcxxwrap-julia/jobs/656610463#L743

This happens consistently on Travis with Nightly Julia, and is resolved when using a manually-patched jll that adds `RTLD_GLOBAL`. 

To keep this as simple as possible, I would suggest keeping the functionality in this PR as it is, and set the `RTLD_GLOBAL` flag on any package that has dependencies that require it. For testing, maybe adding a test in declarative.jl to check that the RTLD flags are stored as expected is enough? I think not setting `RTLD_GLOBAL` results in undefined behavior, and getting a reliable test for this seems difficult.

barche:
I have made an updated version of this PR at PR #701 and can confirm that a JLL generated with these changes fixes the problems with `libcxxwrap_julia`.

staticfloat:
Closing this since we have a new PR that subsumes it.  :)

PR #459, opened 2019-10-28
Add hint about meson to the wizard

giordano:
I followed the cmake style, but noted that it's slightly inconsistent compared to the autoconf one.  Maybe we should make all of them more consistent?


giordano:
Done!  For example, building FriBidi with the wizard now shows:
```
You have the following contents in your working directory:
  - fribidi
Hints:
  - fribidi/configure.ac

    This file is likely input to GNU Autoconf. 
    The recommended options for GNU Autoconf are:

    followed by `make` and `make install`. Since the prefix environment
    variable is set already, this will automatically perform the installation
    into the correct directory.
  - fribidi/meson.build

    This file is likely input to Meson. The recommended option for Meson is

      meson --cross-file=${MESON_TARGET_TOOLCHAIN}

    followed by `ninja` and `ninja install`. Since the prefix environment
    variable is set already, this will automatically perform the installation
    into the correct directory.
```

PR #470, opened 2019-11-04
Do not echo input when asking for a password in the wizard

### LINKED ISSUE #465: GitHub password entry has echo enabled
The convention for CLI entry of passwords is to disable terminal echo so people looking over your shoulder don't see your password.

giordano:
As [suggested by Keno](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/465#issuecomment-549181497), I've looked into `Base.getpass`.  However, this _always_ appends "`: `" to the prompt, which doesn't play nicely with the custom prompt used by `BinaryBuilder.nonempty_line_prompt`.  Probably on Windows one could use `devnull` as output stream, but this wouldn't work on the other systems.  Thus, I've defined a function based on `Base.getpass` which never writes a prompt, as this is the current need in BB.  Maybe something like this should go into `Base.getpass`?

Fixes #465.

### DIFF
--- a/src/wizard/github.jl
+++ b/src/wizard/github.jl
@@ -44,7 +44,7 @@ function obtain_token(; ins=stdin, outs=stdout, github_api=GitHub.DEFAULT_API)
 
     while true
         user = nonempty_line_prompt("Username", "GitHub username:", ins=ins, outs=outs)
-        password = nonempty_line_prompt("Password", "GitHub password:"; ins=ins, outs=outs)
+        password = nonempty_line_prompt("Password", "GitHub password:"; ins=ins, outs=outs, echo=false)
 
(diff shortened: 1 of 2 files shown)


staticfloat:
This should actually be used somewhere, right?

giordano:
Ooops :sweat_smile:

staticfloat:
Awesome; one more thing; can we make a test on this?  E.g. in `test/basic.jl`, pass in our own `ins` and `outs`, and check that with `echo=true` you can see `"hunter2"` as the text that was sent in, but with `echo=false` you can't?

giordano:
`readline` seems to always write to `stdout`, it's not so easy to test unfortunately.

PR #483, opened 2019-11-11
Clone repository over HTTPS

### LINKED ISSUE #482: Error when deploying if the JLL repo exists, but no registration was ever successful
From https://github.com/JuliaPackaging/Yggdrasil/pull/186#issuecomment-552225829

> If the JLL repo exists, but no registration was ever successful, it tried to create the JLL repo, fails, then errors. We need to ignore the error instead.

giordano:
Move cloning out of `try` for cleaner error messages

This hopefully fixes #482.


staticfloat:
Fixed in #484

PR #490, opened 2019-11-12
Give better error messages when JLL packages do not exist.

staticfloat:
Also update registry more automatically, but only once per session


giordano:
This works beautifully, thanks!

PR #496, opened 2019-11-15
Throw error when authenticating if input is closed and GITHUB_TOKEN is empty

### LINKED ISSUE #476: Infinite loop when deploying the artifacts in non-interactive mode and with `GITHUB_TOKEN` not set
This is due to the infinite loop in `obtain_token` that is not getting any input, forever

giordano:
Ref #476.  See https://github.com/JuliaLang/julia/issues/33628#issuecomment-550030136

### DIFF
--- a/src/wizard/github.jl
+++ b/src/wizard/github.jl
@@ -45,6 +45,12 @@ function obtain_token(; ins=stdin, outs=stdout, github_api=GitHub.DEFAULT_API)
     )
 
     while true
+        if !isopen(ins) || (!Sys.iswindows() && !isopen(stdin))
+            # We have to check also `stdin` because `_getpass` ignores `ins` and
+            # always uses `stdin` on Unices.
+            error("Cannot read from input stream")
+        end
+
         user = nonempty_line_prompt("Username", "GitHub username:", ins=ins, outs=outs)
         password = nonempty_line_prompt("Password", "GitHub password:"; ins=ins, outs=outs, echo=false)
 
--- a/test/wizard.jl
+++ b/test/wizard.jl
@@ -252,5 +252,8 @@ end
 @testset "GitHub - authentication" begin
     withenv("GITHUB_TOKEN" => "") do
         @test BinaryBuilder.github_auth(allow_anonymous=true) isa GitHub.AnonymousAuth
+        input_stream = IOBuffer()
+        close(input_stream)
+        @test_throws ErrorException BinaryBuilder.obtain_token(ins=input_stream)
     end
 end


giordano:
Not sure this is going to help, to be honest:
```
% docker run --rm staticfloat/binarybuilder.jl:master julia -e "println(isopen(stdin))"
true
```
:thinking:

giordano:
When running the docker image without a TTY we have:
```
% docker run --rm staticfloat/binarybuilder.jl:master julia -e "dump(stdin)"
IOStream
  handle: Ptr{Nothing} @0x00007f9ac00f8050
  ios: Array{UInt8}((160,)) UInt8[0x40, 0xfe, 0xb8, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00  …  0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]
  name: String "<fd 9>"
  mark: Int64 -1
  lock: ReentrantLock
    locked_by: Nothing nothing
    cond_wait: Base.GenericCondition{Base.Threads.SpinLock}
      waitq: Base.InvasiveLinkedList{Task}
        head: Nothing nothing
        tail: Nothing nothing
      lock: Base.Threads.SpinLock
        handle: Base.Threads.Atomic{Int64}
          value: Int64 0
    reentrancy_cnt: Int64 0
```

When allocating a TTY, instead:
```
% docker run --rm -t staticfloat/binarybuilder.jl:master julia -e "dump(stdin)"
Base.TTY
  handle: Ptr{Nothing} @0x0000000001ee9390
  status: Int64 3
  buffer: Base.GenericIOBuffer{Array{UInt8,1}}
    data: Array{UInt8}((0,)) UInt8[]
    readable: Bool true
    writable: Bool true
    seekable: Bool false
    append: Bool true
    size: Int64 0
    maxsize: Int64 9223372036854775807
    ptr: Int64 1
    mark: Int64 -1
  cond: Base.GenericCondition{Base.Threads.SpinLock}
    waitq: Base.InvasiveLinkedList{Task}
      head: Nothing nothing
      tail: Nothing nothing
    lock: Base.Threads.SpinLock
      handle: Base.Threads.Atomic{Int64}
        value: Int64 0
  readerror: Nothing nothing
  sendbuf: Nothing nothing
  lock: ReentrantLock
    locked_by: Nothing nothing
    cond_wait: Base.GenericCondition{Base.Threads.SpinLock}
      waitq: Base.InvasiveLinkedList{Task}
        head: Nothing nothing
        tail: Nothing nothing
      lock: Base.Threads.SpinLock
        handle: Base.Threads.Atomic{Int64}
          value: Int64 0
    reentrancy_cnt: Int64 0
  throttle: Int64 10485760
```
In this case,
```
% docker run --rm -t staticfloat/binarybuilder.jl:master julia -e "using BinaryBuilder; BinaryBuilder.obtain_token()"

Creating a github access token
To continue, we need to create a GitHub access token, so that we can do
things like fork Yggdrasil and create pull requests against it.  To create
an access token, you will be asked to enter your GitHub credentials:

GitHub username:
> 
```
`obtain_token` waits patiently for the input, but it can't read it.  At least it doesn't enter an infinite loop.

So maybe we should check if the type of the input stream is an `IOStream` rather then if the buffer is open (which is always true)?

src/wizard/github.jl
```diff
@@ -26,6 +26,12 @@ function github_auth(;allow_anonymous::Bool=true)
 end
 
 function obtain_token(; ins=stdin, outs=stdout, github_api=GitHub.DEFAULT_API)
+    if !isopen(ins) || (!Sys.iswindows() && !isopen(stdin))
```
staticfloat: Can we put this check into the `while` loop so that if someone types `^D` during input, we can exit gracefully instead of infinite looping?

giordano:
> We still want to support STDIN being an IOStream, since it's valid for other code to puppeteer things like that (e.g. when running the tests)

But then this PR would not fix #476

staticfloat:
That's true.  I'm not sure how we can check properly to fix #476.

giordano:
> We still want to support STDIN being an IOStream, since it's valid for other code to puppeteer things like that (e.g. when running the tests)

Is input stream as an `IOStream` useful only for testing?  Wouldn't be better to use an `IOBuffer` for that?

test/wizard.jl
```diff
@@ -252,5 +252,8 @@ end
 @testset "GitHub - authentication" begin
     withenv("GITHUB_TOKEN" => "") do
         @test BinaryBuilder.github_auth(allow_anonymous=true) isa GitHub.AnonymousAuth
+        input_stream = IOBuffer()
+        close(input_stream)
+        @test_throws ErrorException BinaryBuilder.obtain_token(ins=input_stream)
```
staticfloat: Can you also test a successful invocation here?

giordano: I'm not sure this is easy: `_getpass` requires a `Base.TTY` as input stream, and even if I knew how to create a custom `Base.TTY` and how to write there, I'd need real GitHub credentials not to get an error :disappointed:

src/wizard/github.jl
giordano: Done!

PR #504, opened 2019-11-19
Add dependencies when running the wizard

meggart:
On current master, binary dependencies were not added to the build environment when running the wizard, this should be fixed by this PR. I am not very familiar with the codebase, so maybe there are better ways to do this.

### DIFF
--- a/src/auditor/compiler_abi.jl
+++ b/src/auditor/compiler_abi.jl
@@ -28,8 +28,7 @@ function check_libgfortran_version(oh::ObjectHandle, platform::Platform; verbose
         if isa(e, InterruptException)
             rethrow(e)
         end
-        @warn("$(path(oh)) could not be scanned for libgfortran dependency!")
-        @warn(e)
+        @warn "$(path(oh)) could not be scanned for libgfortran dependency!" exception=(e, catch_backtrace())
         return true
     end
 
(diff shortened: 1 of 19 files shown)


staticfloat:
Wow, great fix @meggart! I'm going to push some commits to this branch to refactor/fix the tests a bit, then we'll merge.  :D

staticfloat:
Okay I got a little carried away.

src/auditor/compiler_abi.jl
```diff
@@ -158,8 +158,8 @@ function detect_cxxstring_abi(oh::ObjectHandle, platform::Platform; io::IO = std
         if isa(e, InterruptException)
             rethrow(e)
         end
-        warn(io, "$(path(oh)) could not be scanned for cxx11 ABI!")
-        Base.display_error(io, e, [])
+        @warn("$(path(oh)) could not be scanned for cxx11 ABI!")
```
fredrikekre: You can use `exception` to attach the exception:
```suggestion
        @warn("$(path(oh)) could not be scanned for cxx11 ABI!") exception=e
```
optionally with the backtrace like this:
```suggestion
        @warn("$(path(oh)) could not be scanned for cxx11 ABI!") exception=(e, catch_backtrace())
```

staticfloat: WOW I LOVE IT.

PR #510, opened 2019-11-21
Print compiler versions before compiling

giordano:
This is useful in order to keep the versions in the log file.

Maybe too many commands?  Do you suggest some other commands to log?

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -645,6 +645,8 @@ function autobuild(dir::AbstractString,
 
         dest_prefix = Prefix(joinpath(prefix.path, "destdir"))
         did_succeed = with_logfile(dest_prefix, "$(src_name).log") do io
+            # Let's start the presentations with BinaryBuilder.jl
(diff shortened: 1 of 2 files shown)


giordano:
How do we get the version of `ld` for macOS? :woman_facepalming:

staticfloat:
`ld -v` should work, I think.

I would kind of like these values to be within the log file, but not spat out to the console during build.  Perhaps we can run these, separately, adding them to the log file while quashing the output to `stdout`, then run the main script?

giordano:
> `ld -v` should work, I think.

Yes, it does, thanks.

> Perhaps we can run these, separately, adding them to the log file while quashing the output to `stdout`, then run the main script?

Like as it is now?

staticfloat:
Yes, this is really good.  It's a little annoying that it prints out the commands it's going to run so many times, but meh.

Can you also embed the BB version info?  I want to get [this info](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/e403392dc135493e27e6d502019207569315ca52/src/BinaryBuilder.jl#L125-L134), it's okay to split it out into a separate getter function.

PR #514, opened 2019-11-23
Small fixes

staticfloat:


giordano:
While you are at it, you may also want to fix https://github.com/JuliaPackaging/Yggdrasil/pull/222#issuecomment-557823497.  I think this is a consequence of https://github.com/JuliaRegistries/Registrator.jl/pull/247

staticfloat:
Yes, we should; I think we need to move over to `RegistryTools.jl` :(

ararslan:
As a stopgap to fix automatic registration for JLLs until things are ported over to RegistryTools, we can set an upper bound on Registrator to the breaking version. Right now we have the Registrator dependency fixed to semver-compatible with Registrator 1.0, but a recent 1.x version broke BinaryBuilder.

ararslan:
See #517

PR #515, opened 2019-11-23
Some improvements to documentation

giordano:


docs/src/FAQ.md
```diff
@@ -30,8 +30,8 @@ Some linux distributions have a bug in their `overlayfs` implementation that pre
 
 ### I have to build a very small project without a Makefile, what do I have to do?
 
-What BinaryBuilder needs is to find the relevant file (shared libraries, or executables, etc...) organised under the `$prefix` directory: libraries should go to `$prefix/lib` (`$prefix/bin` on Windows), executables to `$prefix/bin`.  You may need to create those directory.  You are free to choose whether to create a simple Makefile to build the project or to do everything within the `build_tarballs.jl` script.
-When the script completes, BinaryBuilder expects to find at least one artifact _built for the expected architecture_ in either `$prefix/lib` or `$prefix/bin`.
+What BinaryBuilder needs is to find the relevant file (shared libraries, or executables, etc...) organised under the `$prefix` directory: libraries should go to `${libdir}`, executables to `${prefix}/bin`.  You may need to create those directory.  You are free to choose whether to create a simple Makefile to build the project or to do everything within the `build_tarballs.jl` script.
+When the script completes, BinaryBuilder expects to find at least one artifact _built for the expected architecture_ in `${libdir}`.
```
staticfloat: We actually have `$bindir` as well, if we want to say that. :)

docs/src/build_tips.md
```diff
@@ -115,7 +115,6 @@ The following environment variables are automatically set in the build environme
 * `CXX`: the C++ cross compiler
 * `FC`: the Fortran cross compiler
 * `OBJC`: the Objective-C cross compiler
-* `AR`: the archiver
```
staticfloat: Wait, if we don't set `$AR` anymore that's probably a bug; we should, I think.  :P

docs/src/build_tips.md
```diff
@@ -129,11 +128,14 @@ The following environment variables are automatically set in the build environme
 
 The following variables are useful to control the build script over different target systems, but are not intended to be modified by the users:
 
+* `prefix`: the path to the top-directory of where all the products should be installed.  This will be the top-directory of the generated tarball
+* `libdir`: the path to the directory where the shared library should be installed.  This is `${prefix}/bin` when building for Windows, `${prefix}/lib` for all other platforms
```
staticfloat: Add `bindir` to this list.

docs/src/build_tips.md
```diff
@@ -148,3 +150,31 @@ if [[ "${target}" == *-freebsd* ]] || [[ "${target}" == *-apple-* ]]; then
     CXX=g++
 fi
 ```
+
+## Installing the license file
+
+Generated tarballs should come with the license of the library that you want to install.  If at the end of a successful build there is only one directory inside `${WORKSPACE}/srcdir`, BinaryBuilder will look into it for files with typical names for license (like `LICENSE`, `COPYRIGHT`, etc... with some combinations of extensions) and automatically install them to `${prefix}/share/licenses/${SRC_NAME}/`.  If in the final tarball there are no files in this directory a warning will be issued, to recall you to provide a license file.
```
staticfloat: `recall you` -> `remind you`

docs/src/FAQ.md
giordano: Right!

docs/src/build_tips.md
giordano: Well, I thought [this was done on purpose](https://github.com/JuliaPackaging/Yggdrasil/commit/2a9ef87c95b704cb78684c0fa58cea21b3a15db8) :stuck_out_tongue:

staticfloat: Okay; I think in general it doesn't matter _that_ much, because we now have the compiler wrappers that live at the default names.  But if we're going to set all those other environment variables, we should set `AR` as well for consistency's sake, I think.

The real power is in `HOSTCC` and whatnot pointing to `x86_64-linux-musl-gcc`.

giordano: Also `LD` is gone, is that expected? :thinking:

staticfloat: Okay if they're all gone then I guess I decided they weren't needed anymore, so we should remove them all.

docs/src/FAQ.md
```diff
@@ -30,8 +30,8 @@ Some linux distributions have a bug in their `overlayfs` implementation that pre
 
 ### I have to build a very small project without a Makefile, what do I have to do?
 
-What BinaryBuilder needs is to find the relevant file (shared libraries, or executables, etc...) organised under the `$prefix` directory: libraries should go to `$prefix/lib` (`$prefix/bin` on Windows), executables to `$prefix/bin`.  You may need to create those directory.  You are free to choose whether to create a simple Makefile to build the project or to do everything within the `build_tarballs.jl` script.
-When the script completes, BinaryBuilder expects to find at least one artifact _built for the expected architecture_ in either `$prefix/lib` or `$prefix/bin`.
+What BinaryBuilder needs is to find the relevant file (shared libraries, or executables, etc...) organised under the `$prefix` directory: libraries should go to `${libdir}`, executables to `${prefix}/bin`.  You may need to create those directory.  You are free to choose whether to create a simple Makefile to build the project or to do everything within the `build_tarballs.jl` script.
```
staticfloat: There's another `${prefix}/bin` here.

docs/src/build_tips.md
giordano: OK, `{{done}}`!

docs/src/build_tips.md
```diff
@@ -109,31 +109,35 @@ You can include local files like patches very easily by placing them within a `b
...
+* `CPPFLAGS`: options for the C pre-processor
 * `LDFLAGS`: options for the linker
 * `PKG_CONFIG_PATH`: a colon-separated list of directories to search for `.pc` files
 * `PKG_CONFIG_SYSROOT_DIR`: modifies `-I` and `-L` to use the directories located in target sysroot
 
 The following variables are useful to control the build script over different target systems, but are not intended to be modified by the users:
 
+* `prefix`: the path to the top-directory of where all the products should be installed.  This will be the top-directory of the generated tarball
+* `libdir`: the path to the directory where the shared libraries should be installed.  This is `${prefix}/bin` when building for Windows, `${prefix}/lib` for all other platforms
+* `libdir`: the path to the directory where the executables should be installed.  This is equivalent to `${prefix}/bin`
```
staticfloat: copypasta.  ;)

giordano:
Ok, I think I also fixed most of the typos I introduced to fix previous errors

PR #516, opened 2019-11-23
Fix error in wizard when no binary dependencies are used.

fredrikekre:
E.g.
```
MethodError: no method matching (::BinaryBuilder.var"#pkgspecify#432")(::Nothing)
Closest candidates are:
  pkgspecify(::Pkg.Types.PackageSpec) at /home/fredrik/dev/BinaryBuilder/src/wizard/interactive_build.jl:295
  pkgspecify(::String) at /home/fredrik/dev/BinaryBuilder/src/wizard/interactive_build.jl:294
```


giordano:
I think this is in #514

fredrikekre:
Cool. Looks like my other error is also fixed by that PR.

PR #517, opened 2019-11-24
Fix Registrator compatibility to v1.0.0

ararslan:
Registrator v1.0.1 removed RegEdit, which breaks automatic JLL registration until BinaryBuilder is ported to use RegistryTools.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -28,5 +28,5 @@ ghr_jll = "07c12ed4-43bc-5495-8a2a-d5838ef8d533"
 
 [compat]
 JLD2 = ">= 0.1.6"
-Registrator = "1"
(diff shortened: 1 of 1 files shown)


giordano:
I was thinking to do the same a few minutes ago, thanks!

ararslan:
Perhaps we should do this in the interim so we can get some of the Yggdrasil PRs that are piling up deployed while I figure out how to wrestle the docs in #518.

giordano:
Ok, I'm taking the executive decision to merge this now and figure out in #518 how to properly fix everything.

staticfloat:
Awesome, thanks guys.  Super pleased that you two are able to just fix things like this.

PR #518, opened 2019-11-24
Use RegistryTools for interacting with the registry

ararslan:
The registry editing functionality was removed from Registrator in version 1.0.1.

Alternative to #517.


giordano:
As a way forward I think this is better than #517.  Glad it was that easy, thank you so much!

ararslan:
I didn't test it though so hopefully this actually works 😛

ararslan:
Ugh, checked in manifests are very annoying. I'm not sure how to update the docs one appropriately.

staticfloat:
It's easy; you just `julia --project=docs` then `pkg> update`.  That's it.

staticfloat:
I rebased and changed a few things around; My test to ensure that things work [looks good](https://github.com/JuliaRegistries/General/pull/5868), so we're merging.

PR #525, opened 2019-11-27
set the prefix to the prefix

visr:
instead of the _jll module src directory

Since I'm not sure the following behavior is intended (see also slack thread):

```julia
julia> using GEOS_jll; GEOS_jll.prefix
"C:\\Users\\visser_mn\\.julia\\packages\\GEOS_jll\\8ezLq\\src\\"
```

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1036,7 +1036,7 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
             Open all libraries
             \"\"\"
             function __init__()
-                global prefix = abspath(joinpath(@__DIR__, ".."))
+                global artifact_dir = abspath(artifact"$(src_name)")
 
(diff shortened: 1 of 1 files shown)


visr:
If this is right, I guess we may want to reuse the `prefix` variable later in the `__init__` rather than keep calling the `artifact_str` macro for efficiency.

ararslan:
I don't have anything useful to add here, I just want to note that I really enjoy the PR title.

staticfloat:
Hmm, yes, I think you are correct.  Let's call it `artifact_dir` instead of `prefix` in any case; I can't see where this is used (I think it must be vestigial) and use it to invoke `@artifact_str` only once (although that macro should be very fast; most of the work is done at compile-time now).

visr:
Changed it to `artifact_dir`, tests are green. The PR title is no longer correct but I don't want to take Alex's joy away ;)

PR #526, opened 2019-11-27
Initialise LIBPATH_list with the libraries provided by Julia

### LINKED ISSUE #524: JLL wrappers should add LIBDIR to LD_LIBRARY_PATH
See e.g. https://github.com/JuliaIO/FFMPEG.jl/commit/53c52cced0130acc3a3639c00625ed98a164719e.

giordano:
Fix #524

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1040,6 +1040,8 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
 
                 # Initialize PATH and LIBPATH environment variable listings
                 global PATH_list, LIBPATH_list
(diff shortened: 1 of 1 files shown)


giordano:
`LIBPATH` is only used for running executables.  Shouldn't it be used also for libraries? :thinking:

staticfloat:
Thanks @giordano!  Super helpful, as always.

> LIBPATH is only used for running executables. Shouldn't it be used also for libraries by setting LIBPATH_env? 🤔

In principle no, we don't need to, because we manually `dlopen()` them.  We could, but so far we don't need to.

PR #528, opened 2019-11-27
Improve `probe_unprivileged_containers()` to check mounting

staticfloat:
We basically need to check that `--map` and `--workspace` work on the
current kernel.  It might be necessary for us to check squashfs and
unpacked separately in the future.


staticfloat:
🦃 🦃 🦃

PR #531, opened 2019-11-28
Fix initialisation of LIBPATH_list

giordano:
This is declared as `const` in the main file of the JLL package, so we should
append to its Ref instead of redefining it.

This should fix https://github.com/JuliaPackaging/Yggdrasil/issues/244.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1041,7 +1041,8 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
                 # Initialize PATH and LIBPATH environment variable listings
                 global PATH_list, LIBPATH_list
                 # We first need to add to LIBPATH_list the libraries provided by Julia
-                LIBPATH_list = [joinpath(Sys.BINDIR, Base.LIBDIR, "julia"), joinpath(Sys.BINDIR, Base.LIBDIR)]
+                append!.(Ref(LIBPATH_list), ([joinpath(Sys.BINDIR, Base.LIBDIR, "julia")], 
+                                             [joinpath(Sys.BINDIR, Base.LIBDIR)]))
             """)
 
             if !isempty(dependencies)


giordano:
Is a generated JLL package that includes an executable tested anywhere?

src/AutoBuild.jl
```diff
@@ -1041,7 +1041,8 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
                 # Initialize PATH and LIBPATH environment variable listings
                 global PATH_list, LIBPATH_list
                 # We first need to add to LIBPATH_list the libraries provided by Julia
-                LIBPATH_list = [joinpath(Sys.BINDIR, Base.LIBDIR, "julia"), joinpath(Sys.BINDIR, Base.LIBDIR)]
+                append!.(Ref(LIBPATH_list), ([joinpath(Sys.BINDIR, Base.LIBDIR, "julia")],
```
giordano: As an alternative, the initialisation can be done where the variable is defined in the top-level file. This is probably a better option.

staticfloat: I think your wrapping is a little off here; I think you should either not broadcast `append!()` here, or broadcast `push!()`.  Example:

```julia
julia> using OpenSpecFun_jll

julia> append!(OpenSpecFun_jll.LIBPATH_list, [
           joinpath(Sys.BINDIR, Base.LIBDIR, "julia"),
           joinpath(Sys.BINDIR, Base.LIBDIR),
       ])
3-element Array{String,1}:
 "/Users/sabae/.julia/artifacts/59ad924b2398533f8d46638a20c34fff66b54f71/lib"
 "/Users/sabae/local/dist/julia-release-1.3/bin/../lib/julia"
 "/Users/sabae/local/dist/julia-release-1.3/bin/../lib"
```

staticfloat:
Fixed in #532

PR #532, opened 2019-11-30
Don't overwrite PATH_list and LIBPATH_list

ararslan:
Should fix
```
julia> using LuaCall
[ Info: Precompiling LuaCall [b19d909c-f536-5a7e-8424-a4fb7ce1e4d7]
WARNING: redefining constant LIBPATH_list
WARNING: redefining constant LIBPATH_list
```

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1041,13 +1041,13 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
                 # Initialize PATH and LIBPATH environment variable listings
                 global PATH_list, LIBPATH_list
                 # We first need to add to LIBPATH_list the libraries provided by Julia
-                LIBPATH_list = [joinpath(Sys.BINDIR, Base.LIBDIR, "julia"), joinpath(Sys.BINDIR, Base.LIBDIR)]
+                append!(LIBPATH_list, [joinpath(Sys.BINDIR, Base.LIBDIR, "julia"), joinpath(Sys.BINDIR, Base.LIBDIR)])
             """)
 
             if !isempty(dependencies)
                 println(io, """
-                    append!.(Ref(PATH_list), ($(join(["$(dep).PATH_list" for dep in dependencies], ", ")),))
-                    append!.(Ref(LIBPATH_list), ($(join(["$(dep).LIBPATH_list" for dep in dependencies], ", ")),))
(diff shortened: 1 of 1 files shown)


src/AutoBuild.jl
```diff
@@ -1041,13 +1041,13 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
                 # We first need to add to LIBPATH_list the libraries provided by Julia
-                LIBPATH_list = [joinpath(Sys.BINDIR, Base.LIBDIR, "julia"), joinpath(Sys.BINDIR, Base.LIBDIR)]
+                append!(LIBPATH_list, [joinpath(Sys.BINDIR, Base.LIBDIR, "julia"), joinpath(Sys.BINDIR, Base.LIBDIR)])
             """)
 
             if !isempty(dependencies)
                 println(io, """
-                    append!.(Ref(PATH_list), ($(join(["$(dep).PATH_list" for dep in dependencies], ", ")),))
-                    append!.(Ref(LIBPATH_list), ($(join(["$(dep).LIBPATH_list" for dep in dependencies], ", ")),))
+                    foreach(p -> append!(PATH_list, p), ($(join(["$(dep).PATH_list" for dep in dependencies], ", ")),))
```
ararslan: I know this is an unrelated change, but I had to think really hard about what the broadcast was supposed to be doing. This also avoids materializing an array containing the result of the broadcast.

src/AutoBuild.jl
```diff
@@ -1041,13 +1041,13 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
                 # Initialize PATH and LIBPATH environment variable listings
                 global PATH_list, LIBPATH_list
                 # We first need to add to LIBPATH_list the libraries provided by Julia
-                LIBPATH_list = [joinpath(Sys.BINDIR, Base.LIBDIR, "julia"), joinpath(Sys.BINDIR, Base.LIBDIR)]
+                append!(LIBPATH_list, [joinpath(Sys.BINDIR, Base.LIBDIR, "julia"), joinpath(Sys.BINDIR, Base.LIBDIR)])
```
ararslan: Ah, I just saw #531. @giordano, I think this is what that line should be that Elliot noted.

src/AutoBuild.jl
staticfloat: I didn't even know that `foreach()` existed.  And yeah, this line is horrible. I'ma add a comment to make this easier to understand for our poor future selves.

PR #533, opened 2019-11-30
Set JLL packages to be installable but not loadable on 1.0-1.2

staticfloat:
This allows users that really want to jump through hoops to avoid bifurcating their packages across the 1.3 divide to do so.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1097,6 +1097,25 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
     open(joinpath(code_dir, "src", "$(src_name)_jll.jl"), "w") do io
         print(io, """
         module $(src_name)_jll
+
+        if VERSION < v"1.3.0-rc4"
+            # We lie a bit in the registry that JLL packages are usable on Julia 1.0-1.2.
+            # This is to allow packages that might want to support Julia 1.0 to get the
+            # benefits of a JLL package on 1.3 (requiring them to declare a dependence on
+            # this JLL package in their Project.toml) but engage in heroic hacks to do
+            # something other than actually use a JLL package on 1.0-1.2.  By allowing
+            # this package to be installed (but not loaded) on 1.0-1.2, we enable users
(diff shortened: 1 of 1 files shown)


src/AutoBuild.jl
```diff
@@ -1097,6 +1097,20 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
+            # We lie a bit in the registry that JLL packages are usable on Julia 1.0-1.2.
+            # This is to allow packages that might want to support Julia 1.0 to get the
+            # benefits of a JLL package on 1.3 (requiring them to declare a dependence on
+            # this JLL package in their Project.toml) but engage in heroic hacks to do
+            # something other than actually use a JLL package on 1.0-1.2.  By allowing
+            # this package to be installed (but not loaded) on 1.0-1.2, we enable users
+            # to avoid splitting their package versions into pre-1.3 and post-1.3 branches
+            # if they are willing to engage in the kinds of hoop-jumping they might need
+            # to in order to install binaries in a JLL-compatible way on 1.0-1.2.
+            error("Unable to import $(src_name)_jll on Julia versions older than 1.3!")
```
c42f:
```suggestion
            error("""Unable to import $(src_name)_jll on Julia versions older than 1.3.
                     If you must support julia versions < 1.3 and >= 1.3 in the same version of
                     your package, there is a hacky escape hatch available: you can use the
                     deprecated build.jl mechanism for the older versions and put your import
                     of the JLL package inside a version check `if VERSION >= v"1.3" ...`.""")
```

staticfloat:
I made the comment a little bit more explicit as to what kinds of heroics an adventurous dev could embark upon; I don't really want to squeeze all that into an error message for two reasons; the first being that a dev who wants to know why this error is happening will likely look at the source code anyway, and secondly that for the casual user I mostly want this error message to be a "don't do that; use Julia 1.3!".

c42f:
Ok then. Personally I'd have liked a subtle hint that there is an escape hatch, but also no good having casual users shooting themselves in the foot.

Issue #534, opened 2019-12-02
`--deploy=<repo>` isn't as flexible as it seems


ssfrr:
The docstring for the `--deploy` flag is:

            --deploy=<repo>     Deploy the built binaries to a github release of
                                an autogenerated wrapper code repository.  Uses
                                `github.com/JuliaBinaryWrappers/<name>_jll.jl` by
                                default, unless `<repo>` is set, in which case it
                                should be set as `<owner>/<name>_jll`.

It seems there are several places where the repo name is actually ignored and it's assumed that it's `<name>_jll.jl`. It seems like it would make more sense just to have this option be `--deploy=<owner>` and have it default to `JuliaBinaryWrappers`. This would better reflect what is actually configurable in the code.

PR #536, opened 2019-12-02
fixes some bugs with private deployments

ssfrr:
- now putting JLL packages in `Pkg.devdir()` instead of hardcoding path
- consolidates some default values that were in multiple places
- respects given `--deploy=<repo>` (see #534)
- deployments were failing if the given repo's owner was a user instead 
of an organization


staticfloat:
Looks good to me!  Thanks @ssfrr!

PR #537, opened 2019-12-02
build and push the JLL package on `--deploy`

ssfrr:
This allows you to deploy the JLL wrapper package along with the binaries so they can be tested, without needing to register the package in General.

(This PR is built on top of #536, so that should be merged first)

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -196,16 +196,16 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
     end
 
     if deploy
-        if register
-            if verbose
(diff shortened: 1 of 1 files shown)


staticfloat:
Hmmm, actually I think this was always supposed to be the case; I think it was a mistake when I was moving stuff around previously that JLL pushing got put inside that `register` conditional.  Good catch!  Rebase please. :)

ssfrr:
rebased!

ssfrr:
Great, thanks!

giordano:
I think this is making it impossible for us to publish jll packages with azure pipelines

PR #541, opened 2019-12-05
Add Sys.BINDIR to LIBPATH_list in JLL wrappers for Windows

giordano:
Honestly I'm a bit confused about how Julia-provided libraries are scattered around, to be sure I'm adding all possible directories.  Ref https://github.com/JuliaGraphics/Gtk.jl/issues/461.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1023,6 +1023,14 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
                 """)
             end
 
+            # Libraries shipped by Julia can be found in different directories,
+            # depending on the operating system and whether Julia has been built
+            # from source or it's a pre-built binary. For all OSes libraries can
+            # be found in Base.LIBDIR or Base.LIBDIR/julia, on Windows they are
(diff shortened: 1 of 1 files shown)


staticfloat:
Hmmm, unclear why tests are failing.  I'll try running locally.

Also, do we know why we keep on getting that "read-only filesystem" error? Is it because something isn't getting unmounted?

giordano:
> Also, do we know why we keep on getting that "read-only filesystem" error? Is it because something isn't getting unmounted?

I don't know, we're getting this quite often in Yggdrasil as well. But this is a recent thing, like one week or so, I've never seen this error locally

staticfloat:
Try rebasing on top of master once https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/542 is merged.  I think it should fix the issues here.  (The issue being that `~/deps/downloads` isn't being created for some reason)

staticfloat:
bingo bango

staticfloat:
.....darn GH interface jumping underneath my cursor.....

Issue #472, opened 2019-11-05
License of the built binary packages.


sambitdash:
Since, the binary distribution packages do not have explicit mention of how the licensing is affected a statement as below should be provided to clarify:
___________
The README of the wrapper packages like say, OpenSSL_jll, should mention the wrappers do not modify the original licensing of the binary packages in any manner. Mere building of the libraries and making them available for download shall not be construed as redistribution. However, downloading the actually binary shall lead to redistribution. Hence, any redistribution restriction of the original licenses shall apply to the consumer of the wrapper binary. And, the consumers must review the `OpenSSL` original license while using the wrapper packages.

Integration code (Julia code) to be used from the wrapper packages are provided under the same licensing as `BinaryBuilder.jl` (MIT Expat Currently).

giordano:
I'm not sure what you're suggesting here.  All packages built in Yggdrasil since last week ship the license file of the library, see #309.  Next build of OpenSSL will have the license file included.  Would this be sufficient for you?

sambitdash:
That should suffice for the binaries like say `openssl.tar.gz`. There is also Julia code like say `src/OpenSSL_jll.jl` or any wrapper Julia file like `src/wrappers/i686-linux-gnu.jl`. What is the licensing these codes are shipped with. These codes may need a "License.md" file created in each of the repository which may be different from the license of `openssl.tar.gz`. If this has been taken care of I guess my concerns are taken care of.

sambitdash:
Ok. I looked at `MKL_jll.jl` released 4 days back. Now this does not have a `License.md` file. Now the question will be while MKL binaries are shipped with certain Intel Simplified Software Libary, it does not cover the Julia code of the `MKL_jll.jl` package.

giordano:
I was talking about the binary tarballs (you can check that the tarballs in https://github.com/JuliaBinaryWrappers/MKL_jll.jl/releases/tag/MKL-v2019.0.117%2B0 do have licenses attached), you're talking about the Julia wrapper packages, they don't have a license.  I think it's a fair point

staticfloat:
For simplicity, I think all JLL packages should default to MIT licensed.  That's the Julia standard, and it is properly permissive for most uses.

PR #543, opened 2019-12-06
License JLL packages themselves as MIT

staticfloat:
Fixes https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/472


staticfloat:
@giordano we really need a test that builds a JLL package then uses it.  :/

PR #547, opened 2019-12-09
Put upper bounds to compat of dependencies

giordano:
I hope I didn't miss any package.  This should allow auto-merge of https://github.com/JuliaRegistries/General/pull/6355


staticfloat:
Do you know how we can tell codecov to be less strict?  It's annoying to always see it failing because coverage decreased by like `-0.11%` or something ridiculous like that.

giordano:
Some time ago I found the option in coveralls to set a threshold, but not in codecov 😣

staticfloat:
Perhaps this: https://docs.codecov.io/docs/commit-status

giordano:
That looks like it!  I probably searched in the web interface only, but indeed codecov doesn't let you do much in the web interface

PR #549, opened 2019-12-10
Revert "build and push the JLL package on `--deploy`"

giordano:
I think we need to quickly revert #537 because it's making it impossible to publish JLL packages built in parallel (as we do in Yggdrasil) and then figure out how to properly achieve what @ssfrr wanted to do there.

As an alternative, we can remove deploying from `build` job in Yggdrasil and move it to `register`, see https://github.com/JuliaPackaging/Yggdrasil/pull/274


giordano:
Superseded by #551

PR #550, opened 2019-12-10
Add small threshold for codecov success

giordano:
Not that the line
```yaml
comment: false
```
is working particularly well, even though [it does seem correct](https://docs.codecov.io/docs/pull-request-comments#section-disable-comment)...

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,1 +1,10 @@
 comment: false
+
+coverage:
+  status:
+    project:
+      default:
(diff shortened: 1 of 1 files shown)


fredrikekre:
`comment: off` has never worked for me either...

giordano:
I should take this is not working either...

staticfloat:
It's possible that it has to be merged before it takes effect.

PR #551, opened 2019-12-11
Split `--deploy` into `--deploy-bin` and `--deploy-jll`

staticfloat:
We still support `--deploy` as a shorthand for "upload both" but this
allows us to selectively upload only one or the other


staticfloat:
Good point, will add that on in a future commit.

PR #553, opened 2019-12-14
Allow selection of compilers in the wizard

giordano:
I'm not very familiar with wizard code.  The wizard now it's failing for me during audit of the second platform tested, quite possible I did something wrong in `src/wizard/interactive_build.jl`.

Maybe this could be the base for an advanced option to activate on request, that lets the user choose other compiler-related stuff, like the preferred GCC and LLVM.

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -319,8 +319,11 @@ function select_closest_version(preferred::VersionNumber, versions::Vector{Versi
     return versions[closest_idx]
 end
 
+const available_gcc_builds = [v"4.8.5", v"5.2.0", v"6.1.0", v"7.1.0", v"8.1.0", v"9.1.0"]
+const available_llvm_builds = [v"6.0.1", v"7.1.0", v"8.0.1", v"9.0.1"]
+
 function select_gcc_version(p::Platform,
-             GCC_builds::Vector{VersionNumber} = [v"4.8.5", v"5.2.0", v"6.1.0", v"7.1.0", v"8.1.0", v"9.1.0"],
+             GCC_builds::Vector{VersionNumber} = available_gcc_builds,
              preferred_gcc_version::VersionNumber = GCC_builds[1],
          )
     # Determine which GCC build we're going to match with this CompilerABI:
@@ -347,8 +350,8 @@ function choose_shards(p::Platform;
             compilers::Vector{Symbol} = [:c],
             rootfs_build::VersionNumber=v"2020.01.07",
             ps_build::VersionNumber=v"2020.01.15",
-            GCC_builds::Vector{VersionNumber}=[v"4.8.5", v"5.2.0", v"6.1.0", v"7.1.0", v"8.1.0", v"9.1.0"],
-            LLVM_builds::Vector{VersionNumber}=[v"6.0.1", v"7.1.0", v"8.0.1", v"9.0.1"],
+            GCC_builds::Vector{VersionNumber}=available_gcc_builds,
+            LLVM_builds::Vector{VersionNumber}=available_llvm_builds,
             Rust_build::VersionNumber=v"1.18.3",
             Go_build::VersionNumber=v"1.13",
             archive_type::Symbol = (use_squashfs ? :squashfs : :unpacked),
--- a/src/wizard/obtain_source.jl
+++ b/src/wizard/obtain_source.jl
@@ -381,6 +381,18 @@ function get_compilers(state::WizardState)
     end
 end
 
+function get_preferred_version(state::WizardState, compiler::AbstractString,
+                               available_versions=Vector{Integer})
+    terminal = TTYTerminal("xterm", state.ins, state.outs, state.outs)
+    version_selected = request(terminal, "Select the preferred $(compiler) version",
+                               RadioMenu(string.(available_versions)))
+    if compiler == "GCC"
+        state.preferred_gcc_version = available_versions[version_selected]
+    elseif compiler == "LLVM"
+        state.preferred_llvm_version = available_versions[version_selected]
+    end
+end
+
 """
     step2(state::WizardState)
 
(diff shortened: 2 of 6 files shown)


giordano:
Having seen the tests timed out, I take that I broke something vary badly :joy:

src/wizard/obtain_source.jl
```diff
@@ -361,4 +385,5 @@ function step2(state::WizardState)
     obtain_source(state)
     obtain_binary_deps(state)
     get_name_and_version(state)
+    get_compilers(state)
```
staticfloat: Because you have changed this, you will need to go into the tests and add simulated keystrokes to select the desired compilers in all "step 2" tests.

src/wizard/obtain_source.jl
```diff
@@ -352,6 +352,30 @@ function get_name_and_version(state::WizardState)
     end
 end
 
+@enum Compilers C=1 Go Rust
```
staticfloat: I'm not really sure what this `@enum` is adding?  Why not just create a list of `["C/C++/Fortran", "Go", "Rust"]` and use the indices directly?

src/wizard/state.jl
```diff
@@ -35,6 +35,7 @@ necessary.  It also holds all necessary metadata such as input/output streams.
     source_files::Union{Nothing, Vector{String}} = nothing
     source_hashes::Union{Nothing, Vector{String}} = nothing
     dependencies::Union{Nothing, Vector} = nothing
+    compilers::Union{Nothing, Vector} = nothing
```
staticfloat: Can you constrain this to be a `Vector{Symbol}`?  I'm extra paranoid about types here since we have to serialize.

src/wizard/obtain_source.jl
giordano: I'd have two vectors here: vectors of strings with the descriptions, vector of symbols, which is a bit tricky to keep in sync.  I think in this way I don't have to care about the order of the elements of the vectors and it's easier to add new compilers or change the description and the name of a compiler.

src/wizard/obtain_source.jl
```diff
@@ -352,6 +352,30 @@ function get_name_and_version(state::WizardState)
...
+                             )
+            if isempty(result)
+                println("Must select at least one platform")
+            else
+                break
+            end
+        end
+        state.compilers = map(c -> compiler_symbols[c], collect(result))
+    end
+end
```
staticfloat: Is there some way we can add `preferred_gcc_version`/`preferred_llvm_version` here?

giordano: Yes, I was thinking about adding all these settings here. However I'll be a bit busy these days 🙂

giordano:
Ok, I fixed the tests, that are now passing -- apart from the Docker one because of #593.

What else do we want here?  `preferred_{gcc,llvm}_version`, anything else?  I was thinking about asking something like "Do you want to customize the default set of compilers? y/N", and if the users answers yes then these options are shown.

staticfloat:
Yeah, sounds good to me.  Let's start with llvm/gcc version and go from there.

src/Rootfs.jl
```diff
@@ -319,8 +319,11 @@ function select_closest_version(preferred::VersionNumber, versions::Vector{Versi
     return versions[closest_idx]
 end
 
+const available_gcc_builds = [v"4.8.5", v"5.2.0", v"6.1.0", v"7.1.0", v"8.1.0"]
```
staticfloat: I just added `v"9.1.0"` as well. :P

src/wizard/obtain_source.jl
```diff
@@ -357,6 +357,49 @@ function get_name_and_version(state::WizardState)
...
+
+function get_preferred_version(state::WizardState, compiler::AbstractString,
+                               available_versions=Vector{Integer})
+    version = nothing
+    minver, maxver = getfield.(extrema(available_versions), :major)
+    while version === nothing #&& !eof(state.ins)
+        msg = "Please enter the preferred $(compiler) version ($(minver)...$(maxver))"
+        version = tryparse(Int, nonempty_line_prompt("$(compiler) version", msg;
```
staticfloat: I don't think we should `tryparse(Int, nonempty_line_prompt(...))`; I think we should `VersionNumber(nonempty_line_prompt(...))`; that way, `"9"`, `"9.1"` and `"v9.1"` will all work.  Not that we allow point releases now, but it's better to by general, IMO.

src/wizard/obtain_source.jl
```diff
@@ -357,6 +357,49 @@ function get_name_and_version(state::WizardState)
...
+                                                     ins=state.ins, outs=state.outs))
+        if !(isnothing(version) || minver <= version <= maxver)
```
staticfloat: I'm not sure this is necessary; we'll always choose the closest version, so if someone asks for GCC `v"1000"`, we'll just give them `v"9"`.  While that would raise some eyebrows, I'm not sure it's necessary to disallow it in the wizard; better to be gracious, IMO.

giordano: This is a temporary implementation: I'd like to use `RadioMenu`, but in order to do that I need to make it work in tests (that's why I asked about simulating arrow keys in tests), but I'll go back to this in the next days.

src/Rootfs.jl
giordano: Now I see why Travis wasn't starting :sweat_smile:

staticfloat:
@giordano you've pushed this a few times since I approved; but go ahead and merge it when you're ready.

giordano:
Yeah, I basically did [this minor change](https://github.com/JuliaPackaging/BinaryBuilder.jl/compare/bbb42a072980736f040ab7c735b2ae9ae3654ae4..9756880bb34d9b15f1baeaa6855ae0054966c78f) and then lots of rebase (including a messy one the other day :wink:).  I'll merge if after CI is green.

PR #557, opened 2019-12-17
Set up JLL dependencies when running in the wizard.

fredrikekre:


src/wizard/interactive_build.jl
```diff
@@ -289,17 +290,7 @@ function step34(state::WizardState)
...
-        ctx = Pkg.Types.Context()
-        update_registry(ctx)
-        pkgspecify(name::String) = Pkg.Types.PackageSpec(;name=name)
-        pkgspecify(ps::Pkg.Types.PackageSpec) = ps
-        dependencies = pkgspecify.(deepcopy(state.dependencies))
-        Pkg.Types.registry_resolve!(ctx.env, dependencies)
-
-        setup_dependencies(prefix, dependencies, platform)
-    end
+    setup_dependencies(prefix, state.dependencies, platform)
```
staticfloat: Don't we still need to specify the dependencies from strings to `PackageSpec`s here?

fredrikekre: I changed it so that everything is PackageSpecified after step2

PR #562, opened 2019-12-18
Dev BinaryBuilder in Dockerfile, instead of adding master

giordano:
When adding master `BinaryBuilderversioninfo()` doesn’t provide useful insights
about the exact version used.

### DIFF
--- a/docker_images/v1.3/Dockerfile
+++ b/docker_images/v1.3/Dockerfile
@@ -21,7 +21,7 @@ RUN mkdir -p /storage/dev_packages
 
 # Install BinaryBuilder
 ADD https://api.github.com/repos/JuliaPackaging/BinaryBuilder.jl/git/refs/heads/master /usr/share/binarybuilder_version.json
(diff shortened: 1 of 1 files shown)


docker_images/v1.3/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN mkdir -p /storage/dev_packages
 
 # Install BinaryBuilder
 ADD https://api.github.com/repos/JuliaPackaging/BinaryBuilder.jl/git/refs/heads/master /usr/share/binarybuilder_version.json
-RUN julia -e 'using Pkg; Pkg.add(PackageSpec(name="BinaryBuilder", rev="master"))'
+RUN julia -e 'using Pkg; Pkg.dev(PackageSpec(name="BinaryBuilder"))'
```
giordano: Wait, the command should be Pkg.develop, right? I think I fixed it locally, but didn't commit or push

PR #564, opened 2019-12-18
Unsymlink deps from the prefixes in the wizard.

fredrikekre:


staticfloat:
I don't actually think we need to clean all these up; only the ones at the end of step 2, but I like the tidiness! :)

staticfloat:
Looks like there's a call to `step3_interactive()` in `step4()` that you need to fix: `MethodError: no method matching step3_interactive(::BinaryBuilder.WizardState, ::Prefix, ::Linux, ::BinaryBuilder.UserNSRunner, ::String)`

fredrikekre:
> I don't actually think we need to clean all these up; only the ones at the end of step 2, but I like the tidiness! :)

Isn't there some cases where prefixes might be reused?

Issue #566, opened 2019-12-19
Github auth issue when submitting the second pr using wizard


innerlee:
The first pr was successfully submitted to Yggdrasil, but the second one fails:

![微信图片_20191219222810](https://user-images.githubusercontent.com/9464825/71181316-dc115a00-22ae-11ea-9148-034b05daefa8.jpg)

PR #570, opened 2019-12-20
Use different note when obtain GitHub token

innerlee:
Fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/566


staticfloat:
Nice, good work @innerlee!

Issue #567, opened 2019-12-19
MacOS shard is missing libc++ and libc++abi headers.


vchuravy:
Looking into the reason why https://github.com/JuliaPackaging/Yggdrasil/pull/296 reliably fails on Clang based shards.

I extracted the reproducer that Clang creates and tried to look at it in isolation. https://gist.github.com/vchuravy/4147c7f28685461b352f48549ad8412f

I see includes like
```
# 508 "/opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/include/math.h" 3 4
```

but I also see many references to the host compiler

```
# 305 "/opt/x86_64-linux-musl/bin/../include/c++/v1/cmath" 3
# 1 "/opt/x86_64-linux-musl/bin/../include/c++/v1/math.h" 1 3
```
It compiles fine using my local clang-9 provided courtesy of Archlinux 
and trying to compile it with a debug clang (v"10") (the file ought to be fairly standalone).

```
vchuravy@antarctic:~$ sh SemaTemplate-b62941.sh
In file included from /workspace/srcdir/llvm-project/clang/lib/Sema/SemaTemplate.cpp:12:
In file included from /workspace/srcdir/llvm-project/clang/lib/Sema/TreeTransform.h:17:
In file included from /workspace/srcdir/llvm-project/clang/lib/Sema/CoroutineStmtBuilder.h:17:
In file included from /workspace/srcdir/llvm-project/clang/include/clang/AST/Decl.h:17:
In file included from /workspace/srcdir/llvm-project/clang/include/clang/AST/APValue.h:17:
In file included from /workspace/srcdir/llvm-project/clang/include/clang/Basic/LLVM.h:22:
In file included from /workspace/srcdir/llvm-project/llvm/include/llvm/Support/Casting.h:18:
In file included from /workspace/srcdir/llvm-project/llvm/include/llvm/Support/Compiler.h:20:
In file included from /opt/x86_64-linux-musl/bin/../include/c++/v1/new:85:
In file included from /opt/x86_64-linux-musl/bin/../include/c++/v1/exception:82:
In file included from /opt/x86_64-linux-musl/bin/../include/c++/v1/cstdlib:86:
In file included from /opt/x86_64-linux-musl/bin/../include/c++/v1/stdlib.h:94:
In file included from /opt/x86_64-apple-darwin14/x86_64-apple-darwin14/sys-root/usr/include/stdlib.h:63:
[...]

PR #571, opened 2019-12-20
Fix macOS libc++ include problems

staticfloat:
* Rebuilt `PlatformSupport` shards to include manual installation of `libc++` headers
* Disabled default search path in `clang`, manually overrode to ignore host headers
* Fixed some compiler warnings when using `-fsyntax-only`


vchuravy:
Should fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/567

PR #578, opened 2019-12-26
Print `supported_platforms()` when wizard recipe builds for all platforms

### LINKED ISSUE #464: When wizard chooses "all platforms", emit `supported_platforms()` in build_tarballs.jl

giordano:
Fix #464.


giordano:
#550 didn't work :disappointed:  Properly setting codecov looks like rocket science

staticfloat:
I think maybe it's because it has a `.` in the front of the name.  I've removed that, let's see if it helps.

PR #585, opened 2019-12-28
Print umount command to stdout/stderr when verbose is requested

giordano:


giordano:
Renaming codecov file worked! :tada:

PR #587, opened 2019-12-29
Write fc wrapper

giordano:
Needs https://github.com/JuliaPackaging/Yggdrasil/pull/345


staticfloat:
I don't think `fc` is a typical name for a fortran compiler (despite the fact that the environment variable is often called `FC`); most old makefiles instead directly call `f77` as the default compiler.  What are you trying to fix by naming it `fc` instead?

giordano:
Fix nothing, I just thought it was a standard name 😛

staticfloat:
Yeah, I don't think anyone actually uses `fc` for fortran.  :)

PR #592, opened 2019-12-31
Remove extra new line at the end of script generated by wizard

giordano:


innerlee:
line ending at the end of file is a good thing, isn't it?

giordano:
One yes, two... not so much. The string that is printed already has a trailing new line, println was adding another one.

innerlee:
ohh, okay

Issue #593, opened 2019-12-31
macos fails hash check


jstrube:
When trying to build a binary on MacOS, BB fails to install the environment with the following error:
```
[ Info: Downloading https://github.com/JuliaPackaging/Yggdrasil/releases/download/PlatformSupport-v2019.12.20+0/PlatformSupport-x86_64-apple-darwin14.v2019.12.20.x86_64-linux-musl.unpacked.tar.gz to /tmp/jl_v1E5Bj-download.gz...
######################################################################## 100.0%
[ Info: No hash cache found
[ Info: Calculated hash a40a3d3f198d5345d789e371a929714f77c062eb202ba677d5eb46091ec4f092 for file /tmp/jl_v1E5Bj-download.gz
[ Info: Unpacking /tmp/jl_v1E5Bj-download.gz into /home/jstrube/.julia/artifacts/jl_nEtsIi...
┌ Error: Tree Hash Mismatch!
│   Expected git-tree-sha1:   3616cf914f5072a044bf163aa950727001b443ba
│   Calculated git-tree-sha1: ea0705297868652a337daf0ccf6d6278796b5ed6
└ @ Pkg.Artifacts /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Pkg/src/Artifacts.jl:779
Unable to automatically install 'PlatformSupport-x86_64-apple-darwin14.v2019.12.20.x86_64-linux-musl.unpacked' from '/home/jstrube/.julia/packages/BinaryBuilder/oKsbb/Artifacts.toml'                                                                                                                                                          
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] #ensure_artifact_installed#42(::Platform, ::Bool, ::typeof(Pkg.Artifacts.ensure_artifact_installed), ::String, ::Dict{String,Any}, ::String) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Pkg/src/Artifacts.jl:873
 [3] (::Pkg.Artifacts.var"#kw##ensure_artifact_installed")(::NamedTuple{(:platform, :verbose),Tuple{Linux,Bool}}, ::typeof(Pkg.Artifacts.ensure_artifact_installed), ::String, ::Dict{String,Any}, ::String) at ./none:0
[...]

PR #597, opened 2020-01-06
Run tests with docker on Julia 1.3.0

giordano:
Tests with `BINARYBUILDER_RUNNER=docker` are failing because of #593.  This pull request uses Julia 1.3.0 in this case, which at least works around the issue.  When #593 is fixed, this change can be reverted.


staticfloat:
You're six hours ahead of me.  :)

PR #602, opened 2020-01-08
Wizard: check if there are other executable files

innerlee:
When no ordinary executable/library found.
For example, with this pr, we can use wizard to generate https://github.com/JuliaPackaging/Yggdrasil/pull/367/files, whose product is a single shell script.


staticfloat:
This won't work; an `ExecutableProduct` only matches actual binary files; it won't match e.g. `bash` scripts that have the executable bit set.  This is on purpose, because an `ExecutableProduct` is meant to be something that is platform-specific; if it's just a normal file, you should use a `FileProduct`.

innerlee:
they are set as FileProduct, see the neofetch yes example.

PR #604, opened 2020-01-11
Setup dependencies that match the GCC ABI of the compiler

giordano:
This should solve the issue we had in https://github.com/JuliaPackaging/Yggdrasil/pull/369 where the build environment had libgfortran3, but a dependency (namely OpenBLAS) was downloaded in its libgfortran5 flavour.

I tested this PR locally with https://github.com/JuliaPackaging/Yggdrasil/pull/378.

### DIFF
--- a/src/DockerRunner.jl
+++ b/src/DockerRunner.jl
@@ -72,6 +72,7 @@ function DockerRunner(workspace_root::String;
                       verbose::Bool = false,
                       compiler_wrapper_path::String = mktempdir(),
                       src_name::AbstractString = "",
+                      shards = nothing,
                       kwargs...)
     global use_ccache
 
@@ -97,8 +98,10 @@ function DockerRunner(workspace_root::String;
         push!(workspaces, "binarybuilder_ccache" => "/root/.ccache")
     end
 
-    # Choose the shards we're going to mount
-    shards = choose_shards(platform; extract_kwargs(kwargs, (:preferred_gcc_version,:preferred_llvm_version,:bootstrap_list,:compilers))...)
+    if isnothing(shards)
+        # Choose the shards we're going to mount
+        shards = choose_shards(platform; extract_kwargs(kwargs, (:preferred_gcc_version,:preferred_llvm_version,:bootstrap_list,:compilers))...)
+    end
 
     # Import docker image
     import_docker_image(shards[1], workspace_root; verbose=verbose)
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -349,6 +349,7 @@ const available_llvm_builds = [LLVMBuild(v"6.0.1"),
 
 """
     gcc_version(cabi::CompilerABI, GCC_builds::Vector{GCCBuild})
+
 Returns the closest matching GCC version number for the given CompilerABI
 representing a particular platofrm, from the given set of options.  If no match
 is found, returns an empty list.  This method assumes that `cabi` represents a
@@ -418,8 +419,11 @@ function choose_shards(p::Platform;
             Go_build::VersionNumber=v"1.13",
             archive_type::Symbol = (use_squashfs ? :squashfs : :unpacked),
             bootstrap_list::Vector{Symbol} = bootstrap_list,
-            # We prefer the oldest GCC version by default
+            # Because GCC has lots of compatibility issues, we always default to
+            # the earliest version possible.
             preferred_gcc_version::VersionNumber = getversion(GCC_builds[1]),
+            # Because LLVM doesn't have compatibilty issues, we always default
+            # to the newest version possible.
             preferred_llvm_version::VersionNumber = getversion(LLVM_builds[end]),
         )
 
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -532,6 +532,59 @@ function expand_cxxstring_abis(ps::Vector{<:Platform})
 end
 
 """
+    preferred_libgfortran_version(platform::Platform, shard::CompilerShard)
+
+Return the libgfortran version preferred by the given platform or GCCBootstrap shard.
+"""
+function preferred_libgfortran_version(platform::Platform, shard::CompilerShard)
+    # Some input validation
+    if shard.name != "GCCBootstrap"
+        error("Shard must be `GCCBootstrap`")
+    end
+    if shard.target.arch != platform.arch || shard.target.libc != platform.libc
+        error("Incompatible platform and shard target")
+    end
+
+    if compiler_abi(platform).libgfortran_version != nothing
+        # Here we can't use `shard.target` because the shard always has the
+        # target as ABI-agnostic, thus we have also to ask for the platform.
+        return compiler_abi(platform).libgfortran_version
+    elseif shard.version < v"7"
+        return v"3"
+    elseif v"7" <= shard.version < v"8"
+        return v"4"
+    else
+        return v"5"
+    end
+end
+
+"""
+    preferred_cxxstring_abi(platform::Platform, shard::CompilerShard)
+
+Return the C++ string ABI preferred by the given platform or GCCBootstrap shard.
+"""
+function preferred_cxxstring_abi(platform::Platform, shard::CompilerShard)
+    # Some input validation
+    if shard.name != "GCCBootstrap"
+        error("Shard must be `GCCBootstrap`")
+    end
+    if shard.target.arch != platform.arch || shard.target.libc != platform.libc
+        error("Incompatible platform and shard target")
+    end
+
+    if compiler_abi(platform).cxxstring_abi != nothing
+        # Here we can't use `shard.target` because the shard always has the
+        # target as ABI-agnostic, thus we have also to ask for the platform.
+        return compiler_abi(platform).cxxstring_abi
+    elseif shard.version < v"5"
+        return :cxx03
+    else
+        return :cxx11
+    end
+end
+
+
+"""
     download_all_shards(; verbose::Bool=false)
 
 Helper function to download all shards/helper binaries so that no matter what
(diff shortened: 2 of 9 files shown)


src/Rootfs.jl
```diff
@@ -528,6 +528,59 @@ function expand_cxxstring_abis(ps::Vector{<:Platform})
...
+    end
+    if shard.target.arch != platform.arch || shard.target.libc != platform.libc
+        error("Incompatible platform and shard target")
+    end
+
+    if compiler_abi(platform).libgfortran_version != nothing
+        # Here we can't use `shard.target` because the shard always has the
+        # target as ABI-agnostic, thus we have also to ask for the platform.
+        return compiler_abi(platform).libgfortran_version
+    elseif shard.version < v"7"
```
giordano: For these versions and the string ABIs below I followed [this table](https://github.com/JuliaPackaging/Yggdrasil/blob/df9066910be48af05fab2dfe480d0d4ec9547227/RootFS.md#compiler-shards), but note that this not up-to-date as we now have GCC 9, too.

staticfloat: Hmmm, I am a little conflicted, because I feel that we are duplicating a lot of the logic inside of [`Pkg.BinaryPlatforms.gcc_version()`](https://github.com/JuliaLang/Pkg.jl/blob/6bdf5fcffb89b2b401d42cad2ab62611f63e54c9/src/BinaryPlatforms.jl#L99); just backwards.  I wonder if we can consolidate the two somehow.

In any case, this change is clearly needed.

src/DockerRunner.jl
```diff
@@ -97,8 +98,10 @@ function DockerRunner(workspace_root::String;
         push!(workspaces, "binarybuilder_ccache" => "/root/.ccache")
     end
 
-    # Choose the shards we're going to mount
-    shards = choose_shards(platform; extract_kwargs(kwargs, (:preferred_gcc_version,:preferred_llvm_version,:bootstrap_list,:compilers))...)
+    if isnothing(shards)
+        # Choose the shards we're going to mount
+        shards = choose_shards(platform; extract_kwargs(kwargs, (:preferred_gcc_version,:preferred_llvm_version,:bootstrap_list,:compilers))...)
```
staticfloat: Can we move this into the `kwarg` initialization above?  I'd rather avoid the "default value of `nothing` means initialize it halfway through the method" pattern if we can, and I don't think there's a hidden reason why `choose_shards()` must be run after the stuff above it.

src/UserNSRunner.jl
```diff
@@ -55,8 +56,10 @@ function UserNSRunner(workspace_root::String;
         push!(workspaces, ccache_dir() => "/root/.ccache")
     end
 
-    # Choose the shards we're going to mount
-    shards = choose_shards(platform; extract_kwargs(kwargs, (:preferred_gcc_version,:preferred_llvm_version,:bootstrap_list,:compilers))...)
+    if isnothing(shards)
+        # Choose the shards we're going to mount
+        shards = choose_shards(platform; extract_kwargs(kwargs, (:preferred_gcc_version,:preferred_llvm_version,:bootstrap_list,:compilers))...)
+    end
```
staticfloat: Same here; let's try moving this initialization into the kwargs specification above.

src/Rootfs.jl
```diff
@@ -319,15 +319,77 @@ function select_closest_version(preferred::VersionNumber, versions::Vector{Versi
...
+                              GCCBuild(v"8.1.0", CompilerABI(libgfortran_version = v"5", libstdcxx_version = v"3.4.25", cxxstring_abi = :cxx11)),
+                              GCCBuild(v"9.1.0", CompilerABI(libgfortran_version = v"5", libstdcxx_version = v"3.4.26", cxxstring_abi = :cxx11))]
+const available_llvm_builds = [LLVMBuild(v"6.0.1"),
+                               LLVMBuild(v"7.1.0"),
+                               LLVMBuild(v"8.0.1"),
+                               LLVMBuild(v"9.0.1")]
+
+"""
+    gcc_version(cabi::CompilerABI, GCC_builds::Vector{GCCBuild})
+Returns the closest matching GCC version number for the given CompilerABI
```
staticfloat: There should be a newline between the function name and the description

src/Rootfs.jl
```diff
@@ -319,15 +319,77 @@ function select_closest_version(preferred::VersionNumber, versions::Vector{Versi
...
+                              GCCBuild(v"5.2.0", CompilerABI(libgfortran_version = v"3", libstdcxx_version = v"3.4.21", cxxstring_abi = :cxx11)),
+                              GCCBuild(v"6.1.0", CompilerABI(libgfortran_version = v"3", libstdcxx_version = v"3.4.22", cxxstring_abi = :cxx11)),
+                              GCCBuild(v"7.1.0", CompilerABI(libgfortran_version = v"4", libstdcxx_version = v"3.4.23", cxxstring_abi = :cxx11)),
+                              GCCBuild(v"8.1.0", CompilerABI(libgfortran_version = v"5", libstdcxx_version = v"3.4.25", cxxstring_abi = :cxx11)),
+                              GCCBuild(v"9.1.0", CompilerABI(libgfortran_version = v"5", libstdcxx_version = v"3.4.26", cxxstring_abi = :cxx11))]
+const available_llvm_builds = [LLVMBuild(v"6.0.1"),
+                               LLVMBuild(v"7.1.0"),
+                               LLVMBuild(v"8.0.1"),
+                               LLVMBuild(v"9.0.1")]
+
```
staticfloat: I can see us eventually adding `Rust`, `Go`, etc... things here.  Nice.

src/Rootfs.jl
```diff
@@ -350,19 +412,19 @@ function choose_shards(p::Platform;
...
+            LLVM_builds::Vector{LLVMBuild}=available_llvm_builds,
             Rust_build::VersionNumber=v"1.18.3",
             Go_build::VersionNumber=v"1.13",
             archive_type::Symbol = (use_squashfs ? :squashfs : :unpacked),
             bootstrap_list::Vector{Symbol} = bootstrap_list,
             # We prefer the oldest GCC version by default
-            preferred_gcc_version::VersionNumber = GCC_builds[1],
-            preferred_llvm_version::VersionNumber = LLVM_builds[end],
+            preferred_gcc_version::VersionNumber = getversion(GCC_builds[1]),
+            preferred_llvm_version::VersionNumber = getversion(LLVM_builds[end]),
```
staticfloat: Let's add a comment here saying "because GCC has lots of compatibility issues, we always default to the earliest version possible.  Because LLVM doesn't have compatibilty issues, we always default to the newest version possible".

test/basic.jl
```diff
@@ -310,3 +309,53 @@ end
     @test truefalse
     @test all(x->x.uuid !== nothing, resolved_deps)
 end
+
+@testset "Rootfs" begin
```
staticfloat: Let's call this `"Compiler Shards"` instead of `"Rootfs"`

test/wizard.jl
```diff
@@ -116,8 +118,8 @@ end
     @test state.source_urls == ["http://127.0.0.1:14444/a/source.tar.gz"]
     @test state.source_hashes == [libfoo_tarball_hash]
     @test Set(state.compilers) == Set([:c, :rust, :go])
-    @test state.preferred_gcc_version == BinaryBuilder.available_gcc_builds[1]
-    @test state.preferred_llvm_version == BinaryBuilder.available_llvm_builds[4]
+    @test state.preferred_gcc_version == getversion(available_gcc_builds[1])
+    @test state.preferred_llvm_version == getversion(BinaryBuilder.available_llvm_builds[4])
```
staticfloat: Why `[4]`?  Is that just `end`?  If so, we should use `[end]` instead.

src/Rootfs.jl
giordano: Awesome, I wondered why the we had the different choice.

test/wizard.jl
giordano: Yes, it should be `[end]`.  Initially in #553 I used `[4]` everywhere, then I realised it was just the last element and fixed, but forgot this one.  Thanks!

src/DockerRunner.jl
giordano: We don't know `kwargs` when we initialise `shards`, do we?

giordano:
One thing come out in https://github.com/JuliaPackaging/Yggdrasil/pull/293 that is slightly relevant for the issue this PR is meant to solve.  In https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/f4ba2911faf2d96853b2e6eebebee75d333e2042/src/Runner.jl#L311-L312 we generate the compiler wrappers for (first) the target and (second) the host.  The `unique` makes sure that we don't write things twice, but this actually happens: `host_platform` is usually `Linux(:x86_64, libc=:musl)`, but if the target has a non-`nothing` compiler ABI specification then the wrappers are written twice, with the ones for the host (always ABI-agonistic) overwriting the one for the target.  In the particular case the non-`nothing` ABI is the C++ string one, this means that the wrappers with the setting of `_GLIBCXX_USE_CXX11_ABI` are lost.

Possible solutions I came up with:
* just invert the order of the loop: write first the wrappers for the host and then those for the target
* force the host to have the same C++ string ABI as the target, when the host has `nothing` and the target has non-`nothing`.  When they're both `nothing` that's fine anyway, when they're both non-`nothing` and different.... that's a problem that we really can't solve.  The ABI coercion can happen in `generate_compiler_wrappers!`.  I was initially thinking about doing this in `autobuild`, but propagating `host_platform` requires a little bit more work to fully propagate the platform value.

What do you think?

giordano:
I went ahead and implemented a mixture of my proposed solutions: let the target wrappers override the host wrappers, but also error out in case we're asking for incompatible ABI.  The error situations should never happen in practice as the `host_platform` keyword argument to `generate_compiler_wrappers!` is never set from the callers.

staticfloat:
Yes, I agree that this is the best solution.

test/wizard.jl
giordano: Well, looking better at the code, it was better for this to be 4: it's matching the fact that the wizard emulator pressed arrow down three times.

PR #606, opened 2020-01-12
Disable ccache for gfortran, as it isn't well supported anyway

giordano:
In https://github.com/JuliaPackaging/Yggdrasil/pull/380 I spent a lot of time before figuring out that ccache was doing a mess and not generating the Fortran module files.

Basic support for Fortran was added in ccache v3.3, but then removed in v3.7 because it wasn't working very well (from [release notes](https://github.com/ccache/ccache/blob/87a4c8e65be9e70aa395246ada688a51bfd0749f/doc/NEWS.adoc#changes-1)), as I can confirm!  We're using ccache 3.5, which is well in between.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -215,7 +215,7 @@ function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractStrin
     # C/C++/Fortran
     gcc(io::IO, p::Platform)      = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-gcc $(gcc_flags(p))"; hash_args=true, link_only_flags=gcc_link_flags(p))
     gxx(io::IO, p::Platform)      = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-g++ $(gcc_flags(p))"; hash_args=true, link_only_flags=gcc_link_flags(p))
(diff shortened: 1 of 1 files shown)


src/Runner.jl
```diff
@@ -215,7 +215,7 @@ function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractStrin
     # C/C++/Fortran
     gcc(io::IO, p::Platform)      = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-gcc $(gcc_flags(p))"; hash_args=true, link_only_flags=gcc_link_flags(p))
     gxx(io::IO, p::Platform)      = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-g++ $(gcc_flags(p))"; hash_args=true, link_only_flags=gcc_link_flags(p))
-    gfortran(io::IO, p::Platform) = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-gfortran $(fortran_flags(p))")
+    gfortran(io::IO, p::Platform) = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-gfortran $(fortran_flags(p))", allow_ccache = false)
```
ararslan:
```suggestion
    gfortran(io::IO, p::Platform) = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-gfortran $(fortran_flags(p))"; allow_ccache=false)
```
For style consistency with the surrounding code

staticfloat:
Good catch, I did not know there were still problems!

PR #321, opened 2018-08-10
bb8

staticfloat:

PR #607, opened 2020-01-13
Make audit actually show error message when it can't open a library

giordano:
This has always struck me, as you don't have a clue why a library can't be dlopened.  This is particularly important when a build is failing remotely, but not locally, like in https://github.com/JuliaPackaging/Yggdrasil/pull/324 and https://github.com/JuliaPackaging/Yggdrasil/pull/336.

### DIFF
--- a/src/Auditor.jl
+++ b/src/Auditor.jl
@@ -124,8 +124,8 @@ function audit(prefix::Prefix, src_name::AbstractString = "";
                 end
             """
             try
-                p = open(`$(Base.julia_cmd()) -e $dlopen_cmd`)
-                wait(p)
+                p = run(`$(Base.julia_cmd()) -e $dlopen_cmd`,
+                        verbose ? (devnull, stdout, stderr) : (devnull, devnull, devnull))
                 if p.exitcode != 0
                     throw("Invalid exit code!")
                 end


src/Auditor.jl
```diff
@@ -124,8 +124,8 @@ function audit(prefix::Prefix, src_name::AbstractString = "";
                 end
             """
             try
-                p = open(`$(Base.julia_cmd()) -e $dlopen_cmd`)
-                wait(p)
+                p = run(`$(Base.julia_cmd()) -e $dlopen_cmd`,
```
giordano: I've just seen that this at some point this was already a `run`, but it was changed to `open` in https://github.com/JuliaPackaging/BinaryBuilder.jl/commit/a906d86d57a97ca6a84c1ed5336fc11fd89290de (part of #321).  The commit message is "Properly isolate `dlopen()` checks again", even though it's not clear to me why `open` is more isolated than `read`.  The problem with `open` is that I couldn't find any way to make it actually reprint the error message.

giordano: `run` was introduced in https://github.com/JuliaPackaging/BinaryBuilder.jl/commit/86f81b376169afd21567dacc82ded399f8f23aef, apparently with the goal to make dlopen errors more verbose, goal defeated by the use of `open`.

giordano:
Ok, I think I'm just blind, the message is already written, it was just hiding from my sight :confused:

PR #610, opened 2020-01-13
Stop compilers from using known unsafe options

giordano:
For context, it was reported in
https://github.com/JuliaCI/BaseBenchmarks.jl/issues/253 that dlopening a library
compiled with `-Ofast` caused the denormal mode to be changed.

This PR disallows using compiler flags that are known to cause troubles.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -46,7 +46,9 @@ and even simple programs will not compile without them.
 function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractString,
                                      host_platform::Platform = Linux(:x86_64; libc=:musl),
                                      rust_platform::Platform = Linux(:x86_64; libc=:glibc),
-                                     compilers::Vector{Symbol} = [:c])
+                                     compilers::Vector{Symbol} = [:c],
+                                     allow_unsafe_flags::Bool = false,
+                                     )
     global use_ccache
 
     # Wipe that directory out, in case it already had compiler wrappers
@@ -66,7 +68,8 @@ function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractStrin
                      hash_args::Bool = false,
                      extra_cmds::String = "",
                      link_only_flags::Vector = String[],
-                     env::Dict{String,String} = Dict{String,String}())
+                     env::Dict{String,String} = Dict{String,String}(),
+                     unsafe_flags = String[])
         write(io, """
         #!/bin/bash
         # This compiler wrapper script brought into existence by `generate_compiler_wrappers()`
@@ -107,6 +110,16 @@ function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractStrin
 
         for (name, val) in env
             write(io, "export $(name)=\"$(val)\"\n")
+        end
+
+        if length(unsafe_flags) >= 1
+            write(io, """
+            if [[ \"\$@\" =~ \"$(join(unsafe_flags, "\"|\""))\" ]]; then
+                echo -e \"You used one or more of the unsafe flags: $(join(unsafe_flags, ", "))\\nPlease repent.\" >&2
+                exit 1
+            fi
+            """)
+            println(io)
         end
 
         if allow_ccache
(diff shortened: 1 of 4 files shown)


src/Runner.jl
```diff
@@ -46,7 +46,9 @@ and even simple programs will not compile without them.
 function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractString,
                                      host_platform::Platform = Linux(:x86_64; libc=:musl),
                                      rust_platform::Platform = Linux(:x86_64; libc=:glibc),
-                                     compilers::Vector{Symbol} = [:c])
+                                     compilers::Vector{Symbol} = [:c],
+                                     allow_unsafe_flags::Bool = false,
```
giordano: This allows for optionally allowing them, if really necessary, but we should propagate this all the way to `build_tarballs`.  Does it make sense?

src/Runner.jl
```diff
@@ -109,6 +112,16 @@ function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractStrin
             write(io, "export $(name)=\"$(val)\"\n")
         end
 
+        if length(unsafe_flags) >= 1
+            write(io, """
+            if [[ \"\$@\" =~ \"$(join(unsafe_flags, '|'))\" ]]; then
+                echo \"You used one or more of the unsafe flags: $(join(unsafe_flags, ", "))\"
```
ararslan:
```suggestion
                echo \"You used one or more of the unsafe flags: $(join(unsafe_flags, ", "))\\nRepent.\"
```

src/Runner.jl
```diff
@@ -213,9 +226,9 @@ function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractStrin
     end
 
     # C/C++/Fortran
-    gcc(io::IO, p::Platform)      = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-gcc $(gcc_flags(p))"; hash_args=true, link_only_flags=gcc_link_flags(p))
-    gxx(io::IO, p::Platform)      = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-g++ $(gcc_flags(p))"; hash_args=true, link_only_flags=gcc_link_flags(p))
-    gfortran(io::IO, p::Platform) = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-gfortran $(fortran_flags(p))"; allow_ccache=false)
+    gcc(io::IO, p::Platform)      = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-gcc $(gcc_flags(p))"; hash_args=true, link_only_flags=gcc_link_flags(p), unsafe_flags = allow_unsafe_flags ? String[] : ["-Ofast"])
```
simonbyrne:
```suggestion
    gcc(io::IO, p::Platform)      = wrapper(io, "/opt/$(triplet(p))/bin/$(triplet(p))-gcc $(gcc_flags(p))"; hash_args=true, link_only_flags=gcc_link_flags(p), unsafe_flags = allow_unsafe_flags ? String[] : ["-Ofast", "-ffast-math", "funsafe-math-optimizations"])
```
and below

src/Runner.jl
staticfloat: Also, can you `>&2` it so that it's red?

ararslan:
```suggestion
                echo \"You used one or more of the unsafe flags: $(join(unsafe_flags, ", "))\\nPlease repent.\" >&2
```

ararslan: Politer version of https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/610#discussion_r366014789 with https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/610#discussion_r366031215 included.

mreineck:
There are legitimate use cases for `-funsafe-math-optimizations` which cannot be achieved by means of any other flag (at least not at the moment). And if the flag is used at compile time, it does not cause any of the nasty effects mentioned at the top. Only its use at link time causes trouble.

Is there a way of stripping the flags in a more selective fashion?

giordano:
> Is there a way of stripping the flags in a more selective fashion?

I don't think we can't really distinguish between compile and linking stage.  However, it is possible to allow unsafe flags (although this feature is not documented), if this is necessary and there is the possibility to not break someone else's code.

PR #611, opened 2020-01-13
Add audit check about troublesome libraries

### LINKED ISSUE #591: Audit should check if a library depends on troublesome system libraries
In https://github.com/JuliaGtk/GtkSourceWidget.jl/pull/9 we found that `Gettext_jl` had its `libgettextlib.dylib` for macOS depending on `/usr/lib/libxml2.2.dylib`, instead of `@rpath/libxml2.2.dylib`, because `XML2_jll` was not used as dependency and the `libxml2.2.dylib` in the BB sysroot was used instead.  This was fixed by https://github.com/JuliaPackaging/Yggdrasil/pull/350.

Similarly, `XML2_jll` had `libxml2.2.dylib` depending on `/usr/lib/libiconv.2.dylib` instead of `@rpath/libiconv.2.dylib`.  This was fixed by https://github.com/JuliaPackaging/Yggdrasil/pull/349.

I suggest that the audit should check if a {Library,Executable}Product depends on these known system libraries, as this will likely be a problem at runtime.

giordano:
Fix #591.

### DIFF
--- a/src/Auditor.jl
+++ b/src/Auditor.jl
@@ -288,6 +288,9 @@ function check_dynamic_linkage(oh, prefix, bin_files;
                     else
                         if !silent
                             @warn("Linked library $(libname) could not be resolved and could not be auto-mapped")
+                            if is_troublesome_library_link(libname, platform)
+                                @warn("Depending on $(libname) is known to cause problems at runtime, make sure to link against the JLL library instead")
+                            end
                         end
                         all_ok = false
                     end
--- a/src/auditor/dynamic_linkage.jl
+++ b/src/auditor/dynamic_linkage.jl
@@ -400,3 +400,20 @@ function update_linkage(prefix::Prefix, platform::Platform, path::AbstractString
 
     return new_libpath
 end
+
+
+"""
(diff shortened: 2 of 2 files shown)


src/Auditor.jl
```diff
@@ -288,6 +288,9 @@ function check_dynamic_linkage(oh, prefix, bin_files;
                     else
                         if !silent
                             @warn("Linked library $(libname) could not be resolved and could not be auto-mapped")
+                            if is_troublesome(libname, platform)
+                                @warn("Depending on $(libname) is known to cause problems at runtime, this needs to be fixed")
```
giordano: Something scarier like `@error`?  This is very likely going to be an issue.

staticfloat: We very clearly need more fine-grained error handling, but `@warn` is fine for now.

Let's call this `is_troublesome_library_link()` to make it a little more explicit what is troublesome.

giordano: Ok, I also made the warning message more clear about how to fix the issue.

PR #612, opened 2020-01-14
Add audit pass to check OS/ABI for FreeBSD libraries

giordano:
This is not a real issue for the dynamic loader, which seems to be happy to open a library with the OS/ABI bit not set to FreeBSD, but we found already a few times that a FreeBSD library with this bit set to 0x00 (System V) upsets our linker, which thinks that this is a foreign library.

I created a new file `auditor/extra_checks.jl` because I think there might be other platform-specific or not very well classified checks in the future.

### DIFF
--- a/src/auditor/extra_checks.jl
+++ b/src/auditor/extra_checks.jl
@@ -1,0 +1,16 @@
+# This file contains some extra checks that don't fall into the other
+# categories, for example because they're very platform-specific.
+
+check_os_abi(::ObjectHandle, ::Platform, rest...; verbose::Bool = false, kwargs...) = true
+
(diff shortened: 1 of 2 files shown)


src/auditor/extra_checks.jl
```diff
@@ -0,0 +1,16 @@
+# This file contains some extra checks that don't fall into the other
+# categories, for example because they're very platform-specific.
+
+check_extra(::ObjectHandle, ::Platform, rest...; verbose::Bool = false, kwargs...) = true
+
+function check_extra(oh::ELFHandle, ::FreeBSD, rest...; verbose::Bool = false, kwargs...)
```
staticfloat: Instead of calling this `check_extra`, I think we should call this `check_os_abi()`.  We can still put it in `extra_checks.jl`, but the function name should be different, IMO.

Issue #614, opened 2020-01-15
Something wrong with cmake/binutils/ar leading to plugin error


quinnj:
Reported originally on slack:
```
[ 65%] Linking C static library libmariadbclient.a
cd /workspace/srcdir/mariadb-connector-c-3.1.6-src/build/libmariadb && /usr/bin/cmake -P CMakeFiles/mariadbclient.dir/cmake_clean_target.cmake
cd /workspace/srcdir/mariadb-connector-c-3.1.6-src/build/libmariadb && /usr/bin/cmake -E cmake_link_script CMakeFiles/mariadbclient.dir/link.txt --verbose=1
[...]

PR #615, opened 2020-01-15
Bump `PlatformSupport` to fix `cmake` trying to use `gcc-ar` instead of `ar`

staticfloat:
This should fix Jacob's cmake woes


staticfloat:
Fixes https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/614

PR #616, opened 2020-01-15
Fix dead `cxxstring_abi` code paths

staticfloat:
We were accidentally stripping out ABI information from our platform
objects while generating compiler wrappers, resulting in `cxxstring_abi`
settings being ignored.  In order to ensure that this does not happen
again, we've added some more tests ensuring that when we ask to build a
`cxx11` library, we actually get what we asked for.

Also, the `cppfilt()` function wasn't working at all because I forgot to `seek(input, 0)` before sending it into the `run_interactive()`.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -289,7 +289,7 @@ function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractStrin
     # Default these tools to the "target tool" versions, will override later
     for tool in (:ar, :as, :cpp, :ld, :nm, :libtool, :objcopy, :objdump, :otool,
                  :ranlib, :readelf, :strip, :install_name_tool, :dlltool, :windres, :winmc, :lipo)
-        @eval $(tool)(io::IO, p::Platform) = $(wrapper)(io, string("/opt/", triplet(abi_agnostic($p)), "/bin/", triplet(abi_agnostic($p)), "-", $(string(tool))); allow_ccache=false)
+        @eval $(tool)(io::IO, p::Platform) = $(wrapper)(io, string("/opt/", triplet(abi_agnostic(p)), "/bin/", triplet(abi_agnostic(p)), "-", $(string(tool))); allow_ccache=false)
     end
  
     # c++filt is hard to write in symbols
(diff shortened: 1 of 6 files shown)


src/auditor/compiler_abi.jl
```diff
@@ -115,14 +115,15 @@ function cppfilt(symbol_names::Vector, platform::Platform)
     for name in symbol_names
         println(input, name)
     end
+    seek(input, 0)
```
giordano: Minor nitpicking:
```suggestion
    seekstart(input)
```

test/auditing.jl
```diff
@@ -51,6 +51,58 @@
...
+                @test isfile(tarball_path)
+
+                # Unpack it somewhere else
+                @test verify(tarball_path, tarball_hash)
+                testdir = joinpath(build_path, "testdir")
+                mkdir(testdir)
+                unpack(tarball_path, testdir)
+                prefix = Prefix(testdir)
+
+                # Ensure that the library detects as the correct cxxstring_abi:
```
giordano: Maybe add also a unit test on `cppfilt` that checks it returns what is supposed to?

src/Runner.jl
```diff
@@ -288,16 +289,16 @@ function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractStrin
     # Default these tools to the "target tool" versions, will override later
     for tool in (:ar, :as, :cpp, :ld, :nm, :libtool, :objcopy, :objdump, :otool,
                  :ranlib, :readelf, :strip, :install_name_tool, :dlltool, :windres, :winmc, :lipo)
-        @eval $(tool)(io::IO, p::Platform) = $(wrapper)(io, string("/opt/", triplet(p), "/bin/", triplet(p), "-", $(string(tool))); allow_ccache=false)
+        @eval $(tool)(io::IO, p::Platform) = $(wrapper)(io, string("/opt/", triplet(abi_agnostic($p)), "/bin/", triplet(abi_agnostic($p)), "-", $(string(tool))); allow_ccache=false)
```
giordano:
```suggestion
        @eval $(tool)(io::IO, p::Platform) = $(wrapper)(io, string("/opt/", triplet(abi_agnostic(p)), "/bin/", triplet(abi_agnostic(p)), "-", $(string(tool))); allow_ccache=false)
```

PR #617, opened 2020-01-16
Upgrade dependencies, add compatibility for Julia 1.4+

staticfloat:
We may finally be able to use BB on 1.4+ again


giordano:
Doesn't this require a bump of the compat of RegistryTools?

staticfloat:
Hmmm, yes, I think it does.

PR #626, opened 2020-01-21
Write to the generated README the actual list of products

giordano:
Suggested by @aminya on Slack.  See an example of README generated with this PR: https://github.com/giordano/Fontconfig/blob/9c04ac01ef658dbb778ff74281427fd2fd3964cb/README.md


src/AutoBuild.jl
```diff
@@ -1218,44 +1219,63 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
+
+              ```julia
+              products = [
+                  FileProduct("src/data.txt", :data_txt),
+                  LibraryProduct("libdataproc", :libdataproc),
+                  ExecutableProduct("mungify", :mungify_exe)
+              ]
+              ```
+
+              With such products defined, `Example_jll` would contain `data_txt`, `libdataproc` and `mungify_exe` symbols exported. For `FileProduct` variables, the exported value is a string pointing to the location of the file on-disk.  For `LibraryProduct` variables, it is a string corresponding to the `SONAME` of the desired library (it will have already been `dlopen()`'ed, so typical `ccall()` usage applies), and for `ExecutableProduct` variables, the exported value is a function that can be called to set appropriate environment variables.  Example:
```
simonbyrne: Could you make the example use the library in question? i.e. base the `ccall` on the first `LibraryProduct` (or drop it if it doesn't exist), etc.

giordano: I thought so initially, but the call would be completely made up, potentially confusing a user reading the example and trying it.

simonbyrne: Could you allow it to be set in the `build_tarballs.jl`? I think the benefit of having runnable examples would be worth the additional complexity.

aminya:
I added a PR to make it possible to give a custom example.
https://github.com/giordano/BinaryBuilder.jl/pull/2

src/AutoBuild.jl
staticfloat: Some recipes build dozens of libraries; I don't think it's going to be easy to get people to give examples for all of them.  On the other hand, it would be neat to have people be able to specify them.  We could have a "example `ccall()`" parser where in the wizard we can ask you "just paste in an exemplary `ccall()` if one exists", and you can provide that ccall in the `build_tarballs.jl` as well.

Hmmm, I'm not sold, but it's an interesting idea.

simonbyrne: I think a single `ccall` would be fine.

fredrikekre: Could also just link to proper docs about how to use executable products, library products and file products.

staticfloat:
@aminya we're going to go ahead without your PR for now because we want to break the API for providing customized usage examples down a bit more.  Feel free to open future PRs directly against BinaryBuilder instead of Mosè's fork.

I'm opening an issue first to discuss how best to implement this: https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/647

aminya:
> 
> 
> @aminya we're going to go ahead without your PR for now because we want to break the API for providing customized usage examples down a bit more. Feel free to open future PRs directly against BinaryBuilder instead of Mosè's fork.
> 
> I'm opening an issue first to discuss how best to implement this: #647

Oh I see. I created a PR here for now:
https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/645

Let me know when you changed the API. You may apply my change to the new API too.

Issue #513, opened 2019-11-23
Integrate AZP logging rules


staticfloat:
It would be neato burrito to have things like our audit passes be cognizant of Azure Pipelines' [logging output infrastructure](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash).  This would allow us to have the audit warnings all be visible from GitHub's summary page, which would be especially useful for Yggdrasil.

oxinabox:
To talk on how I would do this with LoggingExtras:
Becausee you need to control exactly how the message is formatted a `TransformLogger` can't help you.
As the composable loggers are all about the content.
Decising on its presentation is entirely the problem of the Sink.
You can maybe do it by customizing the metaformatter of the ConsoleLogger.
but maybe easier just to declare a custom sink

Here is a untested sketch of the code for the sink
```
import Logging: AbstractLogger, shouldlog, catch_exceptions,min_enabled_level,handle_message

struct AzureSinkLogger{F} <: AbstractLogger
    action_chooser::F
    io::IO
end

AzureSinkLogger(f=_->"task.logissue")=AzureSinkLogger(f, stderr)


function handle_message(logger::AzureSinkLogger, args...; kwargs...)
    #Make it a named tuple for easier working
    log = LoggingExtras.handle_message_args(args...)
    task::String = logger.action_chooser(log)
    
    properties=String[]
    for key in (:level, :_module, :group, :id, :file, :line)
        push!(properties, "$key=$(log[key])")
    end
    for key in keys(args.kwargs)
        push!(properties, "$key=$(log.kwargs[key])")
    end
    props=join(properties, ";")
    # format: ##vso[area.action property1=value;property2=value;...]message
    print(logger.io, "##[$task $props]$(log.message)")
end
shouldlog(::AzureSinkLogger, arg...) = true
min_enabled_level(::AzureSinkLogger) = BelowMinLevel
catch_exceptions(::AzureSinkLogger) = true
```

Then say you only want to do this for warnings and errors
then you use a TeeLogger to send the logging stream to two childs
and on one you put a MinLevelFilter going to a AzureSinkLogger.
And you leave the other with the old `current_logger()`

giordano:
In my fork I issued the following commands in an AZP script:
```sh
      echo "##vso[task.logissue type=error]Something went very wrong."
      echo "##vso[task.logissue type=warning]Something went somewhat wrong."
```
[...]

PR #627, opened 2020-01-22
Layman integration with AZP logging

giordano:
This is a less than ideal implementation of #513.  I believe a long-term solution would need to properly use Julia's logging features (for this reason #513 will remain open), but before we get there we can try with this one to see how it goes, if we like it, and what we need to improve for a more stable implementation.

This gives you an idea of how it appears:
image
The warning is a clickable link to the message printed here:
image

### DIFF
--- a/src/Auditor.jl
+++ b/src/Auditor.jl
@@ -1,5 +1,19 @@
 export audit, collect_files, collapse_symlinks
 
+# Logging functions.  TODO: improve them using Julia logging functionalities, or
+# third-party packages like `LoggingExtras`.
+function audit_warn(short::AbstractString, long::AbstractString, filename::AbstractString, line_number::Int; show_long::Bool = true)
+    if haskey(ENV, "BUILD_BUILDNUMBER")
+        # This is printed only if we're on Azure Pipelines
+        println("##vso[task.logissue type=warning;sourcepath=$(filename);linenumber=$(line_number);]$(short)")
+    end
+    if show_long
+        @warn long
+    end
+end
+audit_warn(long::AbstractString, filename::AbstractString, line_number::Int; kwargs...) =
+    audit_warn(long, long, filename, line_number; kwargs...)
+
 include("auditor/instruction_set.jl")
 include("auditor/dynamic_linkage.jl")
 include("auditor/symlink_translator.jl")
@@ -68,7 +82,8 @@ function audit(prefix::Prefix, src_name::AbstractString = "";
             readmeta(f) do oh
                 if !is_for_platform(oh, platform)
                     if verbose
-                        @warn("Skipping binary analysis of $(relpath(f, prefix.path)) (incorrect platform)")
+                        audit_warn("Skipping binary analysis of $(relpath(f, prefix.path)) (incorrect platform)",
+                                   @__FILE__, @__LINE__)
                     end
                 else
                     # Check that the ISA isn't too high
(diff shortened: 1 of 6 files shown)


src/Auditor.jl
```diff
@@ -120,6 +135,7 @@ function audit(prefix::Prefix, src_name::AbstractString = "";
                     return 0
                 catch e
                     if $(repr(verbose))
+                        println("##vso[task.logissue type=warning;;]Could not dlopen $(repr(f))")
```
staticfloat: Is this supposed to be here?

src/Auditor.jl
```diff
@@ -1,5 +1,19 @@
 export audit, collect_files, collapse_symlinks
 
+# Logging functions.  TODO: improve them using Julia logging functionalities, or
+# third-party packages like `LoggingExtras`.
+function audit_warn(short::AbstractString, long::AbstractString, filename::AbstractString, line_number::Int; show_long::Bool = true)
```
staticfloat: I kind of think we should have an `@audit_warn` that auto-collects `@__LINE__` and `@__FILE__` and can accept `show_long` as a kwarg.  So we would invoke as:

```
@audit_warn "this is my message" show_long=true
```

src/Auditor.jl
giordano: It was intended by me to be there, yes, but it's open for discussion.  I can never find this warning :grimacing:

staticfloat: The reason I ask is because this will print a weird message on people's computers who aren't doing anything with Azure, right?

giordano: Yes, that's true :disappointed:

src/Auditor.jl
giordano: I quickly hacked (to be improved)
```julia
macro audit_warn(long, short = long)
    return quote
        println("##vso[task.logissue type=warning;sourcepath=", $(String(__source__.file::Symbol)),
                ";linenumber=", $(__source__.line), ";]", $(esc(short)))
        @warn($(esc(long)))
    end
end
```
which does the job for the `println`, but reports the wrong file:line for the `@warn` call :disappointed:

staticfloat: Instead of `@warn` try directly calling `Base.Logging.logmsg_code((__module__, String(__source__.file), __source__.line), :Warn, $(esc(long)))`

I got that from figuring out what `@warn` actually lowers to: https://github.com/JuliaLang/julia/blob/2d5741174ce3e6a394010d2e470e4269ca54607f/base/logging.jl#L214

giordano: Yes, right before going to have dinner I tried something along these lines:
```julia
macro audit_warn(long, short = long)
    return quote
        println("##vso[task.logissue type=warning;sourcepath=", $(String(__source__.file::Symbol)),
                ";linenumber=", $(__source__.line), ";]", $(esc(short)))
        Base.CoreLogging.logmsg_code((__module__, String(__source__.file), __source__.line), :Warn, $(esc(long)))
    end
end
```
But I get
```
ERROR: LoadError: UndefVarError: __source__ not defined
Stacktrace:
 [1] top-level scope at /tmp/foo.jl:5
 [2] include at ./boot.jl:328 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1105
 [4] include(::Module, ::String) at ./Base.jl:31
 [5] include(::String) at ./client.jl:424
 [6] top-level scope at REPL[39]:1
```
Need to figure out hygiene here.

staticfloat: Try this:
```
macro audit_warn(long, short = long)
    return quote
        println("##vso[task.logissue type=warning;sourcepath=", $(String(__source__.file::Symbol)), ";linenumber=", $(__source__.line), ";]", $(esc(short)))
        Base.CoreLogging.logmsg_code($(__module__), $(String(__source__.file)), $(__source__.line), :Warn, $(esc(long)))
    end
end
```

Issue #381, opened 2018-10-19
Linking on macOS fails


dpo:
Trying to build a simple Makefile for macOS(:x86_64) results in the error messages:
```
ld: file not found: /usr/lib/system/libcache.dylib for architecture x86_64
```
There is indeed no folder `/usr/lib/system`. This happens with both docker and qemu. I'm using `master` with Julia 0.7.

Thanks.

staticfloat:
Are you using `clang` as your compiler?  Do you have the `build_tarballs.jl` file handy?

helgee:
We are also having linker problems on macOS with a `build_tarballs.jl` that used to work fine.

The respective PR is here: https://github.com/JuliaAstro/SPICEBuilder/pull/1
Travis log: https://api.travis-ci.org/v3/job/460539694/log.txt

I reproduced the problem in a fresh local VM

CC: @giordano

staticfloat:
@helgee your issue is different than this one, I will discuss on your PR thread.

mtsch:
I have a similar problem with Julia 1.0.3 and `BinaryBuilder#master`. I get the following output:
```
sandbox:$WORKSPACE/srcdir# uname
Darwin
sandbox:$WORKSPACE/srcdir# EXT=dylib
sandbox:$WORKSPACE/srcdir# g++ -std=c++11 -Ofast -fPIC -shared -L. -D NDEBUG -D USE_COEFFICIENTS -D ASSEMBLE_REDUCTION_MATRIX -D LIBRIPSER ripser.py/ripser/ripser.cpp -o libripser.$EXT
/tmp/cc62QItK.s:508:11: warning: section "__textcoal_nt" is deprecated
        .section __TEXT,__textcoal_nt,coalesced,pure_instructions
                 ^      ~~~~~~~~~~~~~
/tmp/cc62QItK.s:508:11: note: change section name to "__text"
        .section __TEXT,__textcoal_nt,coalesced,pure_instructions
                 ^      ~~~~~~~~~~~~~
/tmp/cc62QItK.s:1182:11: warning: section "__textcoal_nt" is deprecated
        .section __TEXT,__textcoal_nt,coalesced,pure_instructions
                 ^      ~~~~~~~~~~~~~
/tmp/cc62QItK.s:1182:11: note: change section name to "__text"
        .section __TEXT,__textcoal_nt,coalesced,pure_instructions
                 ^      ~~~~~~~~~~~~~
[...]

PR #630, opened 2020-01-26
`syslibroot` for macOS needs to be manually fixed also for GCC v5

giordano:
Fix error reported in https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/381#issuecomment-578458020, observed in https://github.com/JuliaPackaging/Yggdrasil/pull/421.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -198,7 +198,7 @@ function generate_compiler_wrappers!(platform::Platform; bin_path::AbstractStrin
 
         # On macOS, if we're on an old GCC, the default -syslibroot that gets
         # passed to the linker isn't calculated correctly, so we have to manually set it.
-        if select_gcc_version(p).major == 4
+        if select_gcc_version(p).major in (4, 5)
             FLAGS *= " -Wl,-syslibroot,/opt/$(aatriplet(p))/$(aatriplet(p))/sys-root"
         end
         return FLAGS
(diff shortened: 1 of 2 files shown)


staticfloat:
Can we get a MWE for something that requires the syslibroot set like this?  I want to run this against all the platforms and see if it's actually needed for more than just GCC 4/5.  And I'd like to set up a test so it doesn't bitrot.  :)

giordano:
`HelloWorldFortran` reproduces the error when building for `x86_64-apple-darwin14-libgfortran3` with GCC v5

staticfloat:
Okay so it looks like we dodge this in the default `testsuite` but not in `HelloWorldFortran` because in the default testsuite, we actually run `$(FC) -c -o blah.o blah.f` and then run `$(CC) -o blah blah.o`, so the thing doing the linking is `$(CC)` which is `clang`.  Whereas in `HelloWorldFortran`, we directly run `$(FC) -o blah blah.f`, which means that it's actually `gfortran` doing the linking.

I will update the test suite to compile `HelloWorldFortran` twice so we can exercise both codepaths.

staticfloat:
I've got a slightly different testing approach bundled in with https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/640, so once that goes green we can just merge that.  Thanks for the find!

Issue #636, opened 2020-01-29
Disallow build numbers in `src_version`


staticfloat:
`OpenSS` [had a version number](https://github.com/JuliaPackaging/Yggdrasil/blob/cd7b3f897593030ab4580a61c38b2f5033c96938/O/OpenSSL/build_tarballs.jl#L5) of `1.1.1+c` which caused issues [like this](https://github.com/JuliaRegistries/General/pull/8589/files) because auto-finding the next wrapper version bailed out.  We should just not allow build numbers in the source version of a dependency.

PR #637, opened 2020-01-29
Disallow `src_version` with build number already specified

staticfloat:
Fixes https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/636

### DIFF
--- a/test/building.jl
+++ b/test/building.jl
@@ -197,8 +197,9 @@ shards_to_test = expand_cxxstring_abis(expand_gfortran_versions(shards_to_test))
     end
 end
 
-@testset "Dependency Specification" begin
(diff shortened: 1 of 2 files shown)


giordano:
Why are tests run only without any environment variable? :thinking:

test/building.jl
```diff
@@ -220,3 +220,22 @@ end
         end
     end
 end
+
+@testset "Illegal arguments" begin
+    mktempdir() do build_path
+        @test_logs (:error, r"build number already specified") match_mode=:any begin
```
giordano: You maybe need to swap `@test_logs` and `@test_throws`?

PR #643, opened 2020-01-30
Add custom types for sources and dependencies

### LINKED ISSUE #634: Overhaul of sources and dependencies
Currently, sources can be specified in this form:
```julia
sources = [
    "https://example.org/source.tar.gz" => "0123456789abcdf0123456789abcdf0123456789abcdf0123456789abcdf",
    "http://example.org/repo.git" => "0123456789abcdf0123456789abcdf0123456789",
    "./local-dir",
]
```
I propose to define an abstract type `AbstractSource`, with the following subtypes:
```julia
struct FileSource <: AbstractSource
    url::String
    hash::String
end

struct GitSource <: AbstractSource
    url::String
    hash::String
end

struct DirectorySource <: AbstractSource
    path::String
end
```
The benefit would be more Julian code, [`download_sources`](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/2eec6a4437f666760a263a7dc3ecc5d4110f5e5d/src/AutoBuild.jl#L349-L425) would be much cleaner, with less trial-and-error to guess the type of the source, and better error	messages (#588 was mostly due to the fact that we don't exactly know what the source should be, so the message is very general and vague).

Currently dependencies are of the type
```julia
dependencies = [
    "HelloWorld_jll",
    PackageSpec(name="Ncurses_jll", uuid="68e3532b-a499-55ff-9963-d1c0c0748b3a")
]
```
However, there is a strong desire for having compile-time-only dependencies (e.g., see https://github.com/JuliaPackaging/Yggdrasil/issues/321).  I propose, similarly to the sources above, to have an abstract type `AbstractDependency`, with subtypes:
```julia
struct Dependency{T<:Union{PkgSpec,String}} <: AbstractDependency
    pkg::T
end

struct CompileDependency{T<:Union{PkgSpec,String}} <: AbstractDependency
    pkg::T
end
```
They are basically equivalent to current dependencies, with the exception that a `CompileDependency` would not be added to the list of dependencies of the generated JLL package.  For example, there are many Xorg JLL packages that actually contain only header files and are needed only at compile time, not to actually load the libraries.  Given the fact that the `artifact"..."` macro is not super efficient, being able to remove these packages from the runtime dependencies of Julia packages would also slightly alleviate the problem of loading times.

giordano:
Fix #634.


giordano:
Julia nightly is already in the future:
```
  Cannot locate artifact 'PlatformSupport-armv7l-linux-gnueabihf.v2020.1.15.x86_64-linux-musl.squashfs' in '/home/travis/build/JuliaPackaging/BinaryBuilder.jl/Artifacts.toml'
```

src/Sources.jl
```diff
@@ -0,0 +1,124 @@
...
+    if isfile(source.url)
+        # Immediately abspath() a src_url so we don't lose track of
+        # sources given to us with a relative path
+        src_path = abspath(source.url)
+
+        # And if this is a locally-sourced tarball, just verify
+        verify(src_path, source.hash; verbose=verbose)
+    else
+        # Otherwise, download and verify
+        src_path = storage_dir("downloads", string(source.hash, "-", basename(source.url)))
```
giordano: In this line I had `src_hash` instead of `source.hash`.  This wasn't caught by the tests.... does this mean that tests aren't downloading any file as source? :thinking:

staticfloat: That's entirely possible.  They might all be local sources.  :/

giordano:
Tests are passing and I should have implemented the main features outlined in #634, including comments by Elliot.  So, let's review it!

giordano:
Ok, this should be kind of ready now.

I created the following `build_tarballs.jl` with the wizard (none of the new warnings introduced was issued during the process):
```julia
# Note that this script can accept some limited command-line arguments, run
# `julia build_tarballs.jl --help` to see a usage message.
using BinaryBuilder, Pkg

name = "bmon"
version = v"4.0.0"

# Collection of sources required to complete build
sources = [
    FileSource("https://github.com/tgraf/bmon/releases/download/v4.0/bmon-4.0.tar.gz", "02fdc312b8ceeb5786b28bf905f54328f414040ff42f45c83007f24b76cc9f7a")
]

# Bash recipe for building across all platforms
script = raw"""
cd $WORKSPACE/srcdir
cd bmon-4.0/
export CFLAGS=-I$prefix/include
make -j${nproc}
make install
install_license LICENSE.*
"""

# These are the platforms we will build for by default, unless further
# platforms are passed in on the command line
platforms = [
    MacOS(:x86_64)
]


# The products that we will ensure are always built
products = [
    ExecutableProduct("bmon", :bmon)
]

dependencies = [
    Dependency(PackageSpec(name="Ncurses_jll", uuid="68e3532b-a499-55ff-9963-d1c0c0748b3a"))
    Dependency(PackageSpec(name="libnl_jll", uuid="7c700954-19d3-5208-81e2-8fa5fe7c0bd8"))
    Dependency(PackageSpec(name="libconfuse_jll", uuid="9f03c2a6-2865-5578-ae11-af8a29163b66"))
]

# Build the tarballs, and possibly a `build.jl` as well.
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
```
Then, I made the last dependency build-time, i.e. I replaced
```julia
    Dependency(PackageSpec(name="libconfuse_jll", uuid="9f03c2a6-2865-5578-ae11-af8a29163b66"))
```
with
```julia
    BuildDependency(PackageSpec(name="libconfuse_jll", uuid="9f03c2a6-2865-5578-ae11-af8a29163b66"))
```
and finally I built and deployed the JLL package to my repo: https://github.com/giordano/bmon_jll.jl.  You can see that `libconfuse_jll` is not the [`Project.toml`](https://github.com/giordano/bmon_jll.jl/blob/014ed8bfe2a2995b16c58d3d1b4b19bff7387080/Project.toml) file.

src/AutoBuild.jl
```diff
@@ -1329,12 +1248,12 @@ function push_jll_package(name, build_version;
 end
 
 jll_uuid(name) = Pkg.Types.uuid5(Pkg.Types.uuid_package, "$(name)_jll")
-function build_project_dict(name, version, dependencies)
+function build_project_dict(name, version, dependencies::Array{PkgSpec})
     project = Dict(
         "name" => "$(name)_jll",
         "uuid" => string(jll_uuid("$(name)_jll")),
```
giordano: Shouldn't this be
```julia
"uuid" => string(jll_uuid(name)), 
```
? :thinking:  `jll_uuid` already appends `_jll`.

giordano:
Uhm, it looks like there is still something wrong in the generated JLL: https://github.com/giordano/bmon_jll.jl/blob/014ed8bfe2a2995b16c58d3d1b4b19bff7387080/src/wrappers/x86_64-apple-darwin14.jl#L4-L13 :disappointed:

giordano:
Ok, I should have fixed the above issue in the new repo (I had to delete the previous one):

* [`Project.toml`](https://github.com/giordano/bmon_jll.jl/blob/295474e5e7e8d3c387e05e70d7ad00c33141c745/Project.toml)
* [`src/wrappers/x86_64-apple-darwin14.jl`](https://github.com/giordano/bmon_jll.jl/blob/295474e5e7e8d3c387e05e70d7ad00c33141c745/src/wrappers/x86_64-apple-darwin14.jl)

giordano:
Ok, I'm losing control of this PR: while adding documentation for the new features I also fixed/improved unrelated docstrings :grimacing:

src/Prefix.jl
```diff
@@ -271,62 +271,65 @@ the environment variables that will be defined within the sandbox environment.
...
                     end
-                    run(`unzip -q $(src_path)`)
+                    run(`unzip -q $(source.path)`)
                else
                    if verbose
-                       @info "Copying in $(basename(src_path))..."
+                       @info "Copying in $(basename(source.path))..."
                    end
-                   cp(src_path, basename(src_path))
+                   cp(source.path, joinpath(pwd(), source.unpack_target, basename(source.path)))
```
staticfloat: So if we have an un-extractable file (like a `.dmg` as a source, but we provide an `unpack_target`, it gets put into `${SRCDIR}/${unpack_target}/${basename}`?

src/wizard/deploy.jl
```diff
@@ -1,7 +1,12 @@
...
-    sources_string = strip(join(map(urlhashes) do x
-        string(repr(x[1])," =>\n    ", repr(x[2]), ",\n")
+    urlfiles = zip(state.source_urls, state.source_files)
+    sources_string = strip(join(map(urlfiles) do x
+        url_string = repr(x[1])
+        if endswith(url_string, ".git")
+            "GitSource($(url_string), $(repr(x[2].hash)))"
+        else
+            "FileSource($(url_string), $(repr(x[2].hash)))"
+        end
```
staticfloat: Yeah, this is smarter and cleaner to do this kind of splitting in the wizard, rather than at `build_tarballs()` time.

src/AutoBuild.jl
```diff
@@ -569,19 +494,21 @@ function autobuild(dir::AbstractString,
                    lazy_artifacts::Bool = false,
                    meta_json_stream = nothing,
                    kwargs...)
+    # XXX: These are needed as long as we support old-style sources and
+    # dependencies.
+    sources = coerce_source.(sources)
+    dependencies = coerce_dependency.(dependencies)
```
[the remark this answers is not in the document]
giordano: Yes, it's just that `autobuild` is called directly in a bunch of tests :stuck_out_tongue:   I've removed this and updated the tests accordingly (I hope)

src/Prefix.jl
giordano: Well, I may have swapped the two arguments :sweat_smile:

giordano: Actually, there was no need of the whole `joinpath` thing here, I only had to change `src_path` to `source.path`.  There is a `cd()` above that moves the current directory to the desired place directory.

giordano:
Should be good to go?

src/AutoBuild.jl
giordano: Julia 1.4:
```julia
julia> Pkg.Types.uuid5(Pkg.Types.uuid_package, "LibCURL_jll")
ERROR: UndefVarError: uuid_package not defined
Stacktrace:
 [1] getproperty(::Module, ::Symbol) at ./Base.jl:26
 [2] top-level scope at REPL[3]:1
```
:disappointed:

PR #646, opened 2020-01-30
Add `.julia/artifacts` to Travis cache

giordano:
`deps` is probably not caching much anymore, most of the files are now saved to `.julia/artifacts`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
 cache:
   directories:
     - deps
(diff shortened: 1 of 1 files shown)


staticfloat:
Welp, I don't think this is going to work.  Especially as we continue to find new shards that we need to download during the testing process. :(

Perhaps the long-term strategy is to switch to `xz` or `zstd` compression on our shards to minimize download/extract time.  :P

PR #649, opened 2020-02-06
Add compatibility hack for `arm-linux` -> `armv7l-linux` renaming.

staticfloat:
This is necessary for Julia 1.4+

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ os:
   #- osx
 julia:
   - 1.3
+  - 1.4
   - nightly
 notifications:
   email: false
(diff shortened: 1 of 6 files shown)


giordano:
We should maybe run Travis also on Julia 1.4?

staticfloat:
I think 1.3 + nightly is enough for now; don't want to spend more time waiting for CI.  :P

giordano:
Tests on nightly aren't really running though, Julia v1.4 should be "more working" than nightly and it's the version that is going to be used in Yggdrasil

PR #651, opened 2020-02-07
Add SnoopCompile precompilation harness

staticfloat:
'sup dog, I heard you like compile times

### DIFF
--- a/generate_precompile.jl
+++ b/generate_precompile.jl
@@ -1,0 +1,63 @@
+using SnoopCompile
+
+@snoopiBot "BinaryBuilder" begin
+    using BinaryBuilder
+    # Do an actual build
+    products = Product[
+        ExecutableProduct("hello_world_c", :hello_world_c),
+        ExecutableProduct("hello_world_cxx", :hello_world_cxx),
+        ExecutableProduct("hello_world_fortran", :hello_world_fortran),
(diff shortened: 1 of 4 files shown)


generate_precompile.jl
```diff
@@ -0,0 +1,57 @@
...
+        ExecutableProduct("hello_world_rust", :hello_world_rust),
+    ]
+
+    # First, do the build, but only output the meta json, since we definitely want that to be fast
+    build_tarballs(
+        ["--meta-json=/dev/null"],
+        "testsuite",
+        v"1.0.0",
+        # No sources
+        DirectorySource[],
```
giordano: I don't know much about SnoopCompile, but does it make sense to use a different type of source here since DirectorySource is already used above?

staticfloat: I don't think the compilation time of `DirectorySource()` really matters here.... but I could be wrong!

giordano: I was more thinking about the signature of the method of `build_tarballs` being called, I thought that maybe varying the signature would precompile more methods

staticfloat: Perhaps we should instead `@nospecialize` things, since this really isn't something we want to fight with that much.

staticfloat:
On my machine, this lowers the total time for a `--meta-json` run from 17s to 7s.  That's pretty good!

PR #652, opened 2020-02-08
Add functions for JSON serialise and deserialise dependencies

giordano:


src/Dependencies.jl
```diff
@@ -89,6 +89,36 @@ function resolve_jlls(dependencies::Vector; ctx = Pkg.Types.Context(), outs=stdo
     return all_resolved, dependencies
 end
 
+# Add JSON serialization of dependencies
+JSON.lower(d::Dependency) = Dict("type" => "dependency", "name" => d.pkg.name, "uuid" => d.pkg.uuid,
+                                 "version-major" => d.pkg.version.ranges[1].lower.t[1],
+                                 "version-minor" => d.pkg.version.ranges[1].lower.t[2],
+                                 "version-patch" => d.pkg.version.ranges[1].lower.t[3])
+JSON.lower(d::BuildDependency) = Dict("type" => "builddependency", "name" => d.pkg.name, "uuid" => d.pkg.uuid,
```
giordano: I don't expect this to be called in practice: `BuildDependency` should not make it to the JLL package.  I'm defining this method just not to break things in case this is not the case, but we may want to remove this method in the future.

staticfloat: If we serialize a json representation of a build, this will happen; so I don’t think we can ever drop this. :)

giordano: Oh, right!

giordano: Ok, right, we need to write to the `meta.json` file all dependencies, but at some point we need to filter out the build-time-only dependencies, before publishing the package.  I'm doing this in [`.ci/register_package.jl`](https://github.com/JuliaPackaging/Yggdrasil/commit/46599920af88e74818971b3539fc66b601266171) on Yggdrasil.

PR #656, opened 2020-02-08
Fix building of project dict for registration

giordano:
The call to `jll_uuid` to get the UUID of the dependencies was erroneously
removed in #643, this commit reverts that change and adds regression tests to
avoid similar errors in the future.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1256,13 +1256,15 @@ function bb_specific_uuid5(namespace::UUID, key::String)
     return UUID(u)
 end
 const uuid_package = UUID("cfb74b52-ec16-5bb7-a574-95d9e393895e")
-jll_uuid(name) = bb_specific_uuid5(uuid_package, "$(name)_jll_jll")
-function build_project_dict(name, version, dependencies::Array)
+# For even more interesting historical reasons, we append an extra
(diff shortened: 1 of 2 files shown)


src/AutoBuild.jl
```diff
@@ -1256,13 +1256,15 @@ function bb_specific_uuid5(namespace::UUID, key::String)
     return UUID(u)
 end
 const uuid_package = UUID("cfb74b52-ec16-5bb7-a574-95d9e393895e")
-jll_uuid(name) = bb_specific_uuid5(uuid_package, "$(name)_jll_jll")
-function build_project_dict(name, version, dependencies::Array)
+# For even more interesting historical reasons, we append an extra
+# "_jll" to the name of the new package before computing its UUID.
+jll_uuid(name) = bb_specific_uuid5(uuid_package, "$(name)_jll")
+function build_project_dict(name, version, dependencies::Array{Dependency})
```
giordano: I wasn't 100% this was a good idea, but I'm glad I restricted the signature to `::Array{Dependency}`: this blocked publishing JLL packages with wrong dependencies.

staticfloat:
Great.  I'm glad you fixed my mess. :)

giordano:
Strictly speaking, this error in particular is my fault :joy:

PR #659, opened 2020-02-09
Fix ccall arguments in example usage

laborg:
The wrong ordering might confuse users.


staticfloat:
I always do this backwards.  Thanks.

Issue #635, opened 2020-01-29
Automatic MIT licensing


pabloferz:
It seems that currently [BB automatically assigns a MIT licenses](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/2eec6a4437f666760a263a7dc3ecc5d4110f5e5d/src/AutoBuild.jl#L1261-L1271) to jll packages.

It may be wise to avoid this or to give the user the option to at least opt-in if it is known that the jll package will be MIT compatible.

staticfloat:
So the MIT license bundled with the JLL packages is supposed to apply to the actual JLL code itself, since it comes from an MIT-licensed program (BinaryBuilder.jl).

After discussing this with our resident licensing expert, I think we're going to take the following steps:

* Relicense the auto-generated wrapper code under CC0 (essentially putting it out there as public domain).

* Alter the autogenerated `LICENSE.md` to make it explicit that the CC0 applies only to the Julia source code within `src/`, and that the JLL package will download artifacts that may contain their own licenses.  Each JLL should be shipping with a license within it located in `share/licenses/${src_name}`.

* Extract metadata about the build and store it in plain format within the JLL package itself.  For the purposes of this issue, I'm thinking just the licenses and source locations are important, but I'm opening another issue to track other things we might want to store: https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/639

pabloferz:
OK. I like the idea. That way it would be much clearer what is going. And sourcing licenses and URLs seems enough to avoid issues with GPL licenses.

stevengj:
> Relicense the auto-generated wrapper code under CC0 (essentially putting it out there as public domain).

Have you read CC0?  It's way more complicated than the MIT license: https://creativecommons.org/publicdomain/zero/1.0/legalcode

Much easier to just use the MIT license, which has essentially the same effect as being in the public domain and is compatible with all other licenses.

pabloferz:
That is a good point. I still think the other two points proposed by Elliot are worth implementing.

StefanKarpinski:
May I suggest the [ISC license](https://en.wikipedia.org/wiki/ISC_license)? It’s similar to MIT but without and advertising or attribution clauses, which is very minimal.

stevengj:
[...]

PR #660, opened 2020-02-09
Remove duplicate license file

giordano:
The license file now is created by `init_jll_package`, no need to create another
one in `build_jll_package`.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1202,19 +1202,6 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
               """)
     end
 
-    # Generate LICENSE.md
-    open(joinpath(code_dir, "LICENSE.md"), "w") do io
-        print(io, """
-        Copyright $(year(now()))
-
-        Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
-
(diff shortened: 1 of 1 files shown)


staticfloat:
Mmmmm, I think we actually want to keep this, if we're going to solve https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/635#issuecomment-580018826

giordano:
Why two license files though?  Wouldn't the one generated by `init_jll_package` sufficient?

staticfloat:
This overwrites the one made by GitHub through `init_jll_package`; that one is also not flexible, so if we're going to write out own license file, we're going to need to write it out like this one.

The reason the license file being made in `init_jll_package` is important though, is that it forces GH to initialize the JLL wrapper code repository with at least one commit, which means that the repository is then cloneable; otherwise we can get into weird edge cases where the repository is already created, but it's not cloneable.

giordano:
> This overwrites the one made by GitHub through init_jll_package

They have different names, so we get `LICENSE` and `LICENSE.md`: https://github.com/JuliaBinaryWrappers/CUTEst_jll.jl/tree/1b9564896cf746876cff057022f9eebd12186249

staticfloat:
Oh.  The packages I checked I guess were old enough that they weren't showing the `LICENSE` file.

We'll now get rid of old stuff, once https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/663 is merged.  But we could also write our own license file out to `LICENSE` instead of `LICENSE.md`?

giordano:
> But we could also write our own license file out to LICENSE instead of LICENSE.md?

Yeah, I think also PkgTemplates nowadays uses `LICENSE` instead of `LICENSE.md`.

PR #662, opened 2020-02-11
Check the string of the URL, not its representation

giordano:
Should fix the error in https://github.com/JuliaPackaging/Yggdrasil/pull/491#discussion_r377583378, but I'd like to double check before merging this.


giordano:
Yes, I confirm it works.  We should really have some tests for this, though...

PR #666, opened 2020-02-12
Don't allow invalid identifiers in the wizard

staticfloat:


giordano:
Probably also `build_tarballs` and/or `autobuild` should have this check.  It's easy to do the mistake also outside the wizard :slightly_smiling_face:

staticfloat:
Hmmm, since it's only for registration that it's a problem, I'll add it to `build_jll` and `register_jll` methods.

Issue #668, opened 2020-02-12
Rework downloading procedure


staticfloat:
We should download sources as artifacts, instead of downloading things to the "storage directory".  (We continue the march towards completely eliminating that storage directory.  Fully immutable package directories are the future!).  I'm imagining the download routine looking something like this:

```julia
source_treehash = create_artifact() do art_dir
	src_path = joinpath(art_dir, basename(src.url))
	download_verify(source.url, source.hash, src_path; verbose=verbose)
end
```

This has the following advantages:

* We won't need to [prefix downloaded files with hashes](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/8c89c90bbcc641227c88be6c941f6d1b3d700f29/src/Sources.jl#L86); the (content tree, not tarball) hash will be implicit in the directory it's being downloaded to.  If we store a mapping of `(url, tarball_hash) => tree_hash` mappings somewhere, we can even hit that cache, check if the artifact still exists, and if it does, skip downloading the file.  (we need to store a database because we don't know the tree hash beforehand; but if we did, we could even skip this step and just directly check to see if the artifact is installed!)

* We can handle filename conflicts more intelligently.  Instead of prefixing things with hashes, we can rely on `unpack_target` and `filename` on `FileSource`s and `ArchiveSource`s (once they exist).  We can detect when we would overwrite a file that already exists and error out, forcing the user to disambiguate things.

* We may one day go super hard on the full reproducibility angle, where having all sources canonicalized as artifacts (and uploaded to a Pkg server somewhere) is desirable; having things using artifacts anyway will make that easier for us in the future, so that we can always rebuild an older version of a recipe.

PR #671, opened 2020-02-13
Add ArchiveSource type

giordano:
Partially address #668 by adding the type `ArchiveSource`

### DIFF
--- a/test/building.jl
+++ b/test/building.jl
@@ -273,3 +273,32 @@ end
     end
 end
 
+@testset "Building from remote file" begin
+    build_output_meta = nothing
+    mktempdir() do build_path
+        build_output_meta = autobuild(
+            build_path,
+            "libconfuse",
+            v"3.2.2",
+            # libconfuse source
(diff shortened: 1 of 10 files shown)


giordano:
Interestingly enough, the test that is currently failing passes locally for me

giordano:
Ok, should be ready for review now

test/building.jl
```diff
@@ -273,3 +273,39 @@ end
     end
 end
 
+@testset "Building from remote file" begin
```
giordano: I've added a very simple test for building a remote source, as this code path is completely untested.  If you have better ideas about what and how to test here, please share

staticfloat: This seems fine to me; it does mean that our tests may get sad if the network goes down; we could run a local webserver to fix that, but let's cross that bridge once we get to it.

staticfloat:
Wow, this looks really good to me.  Great improvement!  No comments!

staticfloat:
I'm merging, but remember that we'll need to bump the `Manifest` in Yggdrasil before we will see these changes.

PR #676, opened 2020-02-17
Do not replace existing artifacts

### LINKED ISSUE #554: Check if the tag is already existing before attempting new registration and upload of artifacts
Some history:
* https://github.com/JuliaPackaging/Yggdrasil/pull/289 was merged before https://github.com/JuliaRegistries/General/pull/6701 was merged
* `get_next_wrapper_version` decided that the next wrapper version would have been `v"5.4.0+2"`, as `v"5.4.0+1"` is the [last registered version](https://github.com/JuliaRegistries/General/blob/62741f41b7b8c23609a807338abdd4a815bec2d2/S/SuiteSparse_jll/Versions.toml)
* the new artifacts have been uploaded to release [`SuiteSparse-v5.4.0+2`](https://github.com/JuliaBinaryWrappers/SuiteSparse_jll.jl/releases/tag/SuiteSparse-v5.4.0%2B2), **overwriting** the existing ones
* this wreaked havoc into Julia, causing https://github.com/JuliaLang/julia/pull/34082#issuecomment-565837625

We need to check if the tag is already existing before attempting the upload and registration.

giordano:
This feature is useful in theory, but causes a lot of harm in practice, for example if `get_next_wrapper_version` fails to get the version number right for unknown reasons.

Fix #554.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -305,7 +305,7 @@ function upload_to_github_releases(repo, tag, path; attempts::Int = 3, verbose::
     for attempt in 1:attempts
         try
             ghr() do ghr_path
-                run(`$ghr_path -replace -u $(dirname(repo)) -r $(basename(repo)) $(tag) $(path)`)
+                run(`$ghr_path -u $(dirname(repo)) -r $(basename(repo)) $(tag) $(path)`)
             end
             return
         catch


staticfloat:
I'll bump Yggdrasil and we can push a `HelloWorldX` to make sure this is working.

Issue #125, opened 2018-01-30
Automatically generate diff of any source files


Keno:
Sometimes you end up having to modify files to make the build go through.
We should identify this situation, generate a diff of the file at the end of the build
and give the user the option to apply the diff at the beginning of the build.

staticfloat:
I wonder if we can auto-generate this from the `overlayfs` work directories?  That would be pretty sweet.

staticfloat:
Ah, no, the workspace is the _one_ place that isn't covered by an overlayfs.  We'd just have to keep track of the original state of the files, then diff everything.

PR #678, opened 2020-02-17
Diff source tree after interactive build

Keno:
Implements https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/125.

Example PR: https://github.com/JuliaPackaging/Yggdrasil/pull/519

### DIFF
--- a/src/Prefix.jl
+++ b/src/Prefix.jl
@@ -314,7 +314,6 @@ function setup(source::SetupSource{DirectorySource}, targetdir, verbose)
 end
 
 function setup(source::PatchSource, targetdir, verbose)
-    # Need to strip the trailing separator also here
     if verbose
         @info "Adding patch $(source.name)..."
     end
--- a/src/Prefix.jl
+++ b/src/Prefix.jl
@@ -313,6 +313,16 @@ function setup(source::SetupSource{DirectorySource}, targetdir, verbose)
     end
 end
 
+function setup(source::PatchSource, targetdir, verbose)
+    # Need to strip the trailing separator also here
+    if verbose
+        @info "Adding patch $(source.name)..."
+    end
+    patches_dir = joinpath(targetdir, "patches")
+    mkdir(patches_dir)
+    open(f->write(f, source.patch), joinpath(patches_dir, source.name), "w")
+end
+
 """
     setup_workspace(build_path::String, sources::Vector{SetupSource};
                     verbose::Bool = false)
@@ -324,8 +334,8 @@ the environment variables that will be defined within the sandbox environment.
 This method returns the `Prefix` to install things into, and the runner
 that can be used to launch commands within this workspace.
 """
-function setup_workspace(build_path::AbstractString, sources::Vector{<:SetupSource};
-                         verbose::Bool = false)
+function setup_workspace(build_path::AbstractString, sources::Vector;
+                         verbose::Bool = false, srcdir_overlay = false)
     # Use a random nonce to make detection of paths in embedded binary easier
     nonce = randstring()
     workspace = joinpath(build_path, nonce)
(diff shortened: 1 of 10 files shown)


src/wizard/interactive_build.jl
```diff
@@ -136,44 +136,101 @@ end
...
+        cmd = `/bin/bash -c $cmd`
+    end
+
+    run_interactive(ur, cmd, stdin=state.ins, stdout=state.outs, stderr=state.outs)
+
+    had_patches = false
+    if srcdir_overlay
+        # Check if there were any modifications made to the srcdir
+        if length(readdir(joinpath(prefix, "srcdir"))) > 0
+            # Note that we're mounting to /workspace/srcdir/upper so comparing
```
staticfloat: Is this comment right?  It looks like you're mounting to `/workspace/srcdir`, not `/workspace/srcdir/upper`.

src/wizard/deploy.jl
```diff
@@ -107,7 +112,7 @@ end
 Write out a WizardState to a `build_tarballs.jl` in an `Yggdrasil` clone, then
 open a pull request against `Yggdrasil`.
 """
-function yggdrasil_deploy(state::WizardState)
+function yggdrasil_deploy(state::WizardState, open_it::Bool = true)
```
staticfloat: Instead of `open_it` let's call it `open_pr`.

src/wizard/deploy.jl
```diff
@@ -199,15 +222,22 @@ function step7(state::WizardState)
...
             "Write to a local file",
             "Print to stdout",
         ])
     )
     println(state.outs)
 
     if deploy_select == 1
-        yggdrasil_deploy(state)
+        yggdrasil_select = request(terminal,
+            "Would you like to actuall open the pull request or just prepare it?",
```
staticfloat:
```suggestion
            "Would you like to actually open the pull request or just prepare it?",
```

src/Prefix.jl
```diff
@@ -338,13 +348,23 @@ function setup_workspace(build_path::AbstractString, sources::Vector{<:SetupSour
     wrapperdir = joinpath(workspace, "compiler_wrappers")
     mkdir.((srcdir, destdir, metadir))
 
+    if srcdir_overlay
+        mkdir(joinpath(build_path, nonce, "srcdir_work"))
+        srcdir = joinpath(build_path, nonce, "srcdir_lower")
+        mkdir(srcdir)
+    end
```
staticfloat: I would prefer that this is always `srcdir`, not `srcdir_lower`; reduces complexity and makes it so I don't have to think about what's going on.  We can bindmount it to `${WORKSPACE}/srcdir_lower` just before we mount the overlay to get this to work.

src/Sources.jl
```diff
@@ -107,6 +107,10 @@ function SetupSource(url::String, path::String, hash::String, target::String)
     end
 end
 
+struct PatchSource
+    name::String
+    patch::String
+end
```
staticfloat: I'm a little worried someone will use this in their `build_tarballs.jl`; we probably want to give this a different name to make sure it's an internal-only thing.

giordano: Note that it isn't an `AbstractSource` (nor it is exported), so someone using it as source in a builder won't go too far

src/Prefix.jl
```diff
@@ -313,6 +313,16 @@ function setup(source::SetupSource{DirectorySource}, targetdir, verbose)
     end
 end
 
+function setup(source::PatchSource, targetdir, verbose)
+    # Need to strip the trailing separator also here
```
giordano: Copied by mistake?
```suggestion
```

src/wizard/deploy.jl
```diff
@@ -1,15 +1,20 @@
...
         elseif any(endswith(x[1], ext) for ext in archive_extensions)
             "ArchiveSource($(url_string), $(repr(x[2].hash)))"
         else
-            "FileSource($(url_string), $(repr(x[2].hash))),"
+            "FileSource($(url_string), $(repr(x[2].hash)))"
         end
-    end,"\n    "))
+    end
+    if !isempty(state.patches)
+        push!(sources_strings, "\"./bundled\"")
```
giordano:
```suggestion
        push!(sources_strings, "DirectorySource(\"./bundled\")")
```

Issue #677, opened 2020-02-17
Entering product names is annoying


Keno:
Particularly if there are many products. We should try to heuristically suggest a good name (e.g. the executable name for the executable and the library name without `lib` for libraries).

giordano:
The easy part is guessing a default name: we can use `parse_dl_name_version` from `Pkg.BinaryPlatforms` for the libraries:
```julia
replace(parse_dl_name_version(libname, platform)[1], "-" => "_")
```
and `splitext` for the executables:
```
replace(splitext(exename)[1], "-" => "_")
```

The hardest part is to change `nonempty_line_prompt` to accept a default answer in the command line that a user can edit if necessary

Keno:
I have a patch for this. Just need to get through my patch stack.

PR #681, opened 2020-02-19
Wizard: Don't ignore error paths

Keno:
Also a minimal fix to #677. This is on top of #678 because it touches some of the files and I'm hoping we can merge that soon-ish. If not I can do the rebase dance to switch the order of these though.

### DIFF
--- a/src/DockerRunner.jl
+++ b/src/DockerRunner.jl
@@ -236,8 +236,9 @@ function run_interactive(dr::DockerRunner, cmd::Cmd; stdin = nothing, stdout = n
                 end
             end
             wait(process)
+            return success(process)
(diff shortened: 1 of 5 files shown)


Keno:
Rebased and updated base branch.

Keno:
1.4 test failure is due to rc2 not liking something we do. Also happens on master.

giordano:
Is it only when using the wizard?  We use Julia 1.4 on Yggdrasil

Keno:
Yes, it's in the wizard, but it literally just started failing because 1.4rc2 contains a Pkg bump. Wouldn't be surprised if something else broke as well.

PR #685, opened 2020-02-22
Add two build tips about building with dependencies

sloede:
I find it highly confusing that there are two sources for build tips for BinaryBuilder.jl: one is the Yggdrasil GitHub wiki and one can be found in the official docs. Since I am not familiar with the original rationale behind this choice, I would at least propose to include these two build tips in the official docs, as they seem to be very helpful for novice users and at least I was only able to find them by asking for help.

If it were possible to get consensus from everyone involved, I'd actually propose to close down the wiki at Yggdrasil and move all documentation to here. IMHO, that would be the best solution for new users.

### DIFF
--- a/docs/src/build_tips.md
+++ b/docs/src/build_tips.md
@@ -182,3 +182,39 @@ In addition to the standard Unix tools, in the build environment there are some 
   update_configure_scripts
   ```
   to get a newer version
+
+
+## Header files of the dependencies can't be found
+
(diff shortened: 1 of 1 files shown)


sloede:
I don't know what to do to make the codecov check pass, but I am willing to try to improve it if someone can give me hint to what I should be looking at.

staticfloat:
You raise good points @sloede!  What we're going to do is to link from the BB docs to the Yggdrasil build wiki, as it's easier to maintain an up-to-date list of these kinds of tricks there, especially as they are rarely BB-version specific, so no use in versioning them along with the source code.

PR #688, opened 2020-02-23
Factor out some `build_tarball` inner functions

pabloferz:
I saw one of these being redefined in `CUDABuilder`, and it seemed a good idea to factor these out for cases where you have a builder that provides different versions of a library within the same release.


pabloferz:
I'm not sure about leaving the `build_tarballs.jl` help text out of the the function definition, and I can drop that commit, but other than that is this acceptable?

giordano:
Uh, sorry, I forgot to get back to this PR, thanks for reminding me.  Generally speaking I personally like short functions, so that I can have an easier look at the entire body.  I'm going to merge this, thanks!

PR #690, opened 2020-02-24
Update travis configuration to match comments

Keno:
On travis, we default to squashfs shards, so in order to test unpacked shards,
we need to set the env var to false. Also it was misspelled for the one instance
where it was set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,17 +28,19 @@ jobs:
     - julia: 1.3
       env:
         - BINARYBUILDER_RUNNER=privileged
-        - BINARYBUILDER_USER_SQUASHFS=true
+        - BINARYBUILDER_USE_SQUASHFS=true
 
     # Add a job that uses the unprivileged builder with unpacked shards
     - julia: 1.3
       env:
         - BINARYBUILDER_RUNNER=unprivileged
+        - BINARYBUILDER_USE_SQUASHFS=false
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -28,17 +28,19 @@ jobs:
     - julia: 1.3
       env:
         - BINARYBUILDER_RUNNER=privileged
-        - BINARYBUILDER_USER_SQUASHFS=true
+        - BINARYBUILDER_USE_SQUASHFS=true
 
     # Add a job that uses the unprivileged builder with unpacked shards
     - julia: 1.3
```
giordano: If we want to respect the comment and have `BINARYBUILDER_USE_SQUASHFS=false`, this needs to be
```suggestion
    - julia: 1.3.1
```
because of #593

.travis.yml
```diff
@@ -28,17 +28,19 @@ jobs:
...
       env:
         - BINARYBUILDER_RUNNER=privileged
-        - BINARYBUILDER_USER_SQUASHFS=true
+        - BINARYBUILDER_USE_SQUASHFS=true
 
     # Add a job that uses the unprivileged builder with unpacked shards
-    - julia: 1.3
+    - julia: 1.3.1
```
giordano: I meant
```suggestion
    - julia: 1.3.0
```
:man_facepalming:

Issue #393, opened 2018-11-23
Qt build issues


barche:
When building Qt using the following configure command from inside a `BinaryBuilder.runshell(Linux(:x86_64, compiler_abi=CompilerABI(:gcc7,:cxx11)))`

```
./configure -v -prefix /workspace/install -opensource -confirm-license -skip qtactiveqt -skip qtandroidextras -skip qtcanvas3d -skip qtconnectivity -skip qtdatavis3d -skip qtdoc -skip qtgamepad -skip qtnetworkauth -skip qtpurchasing -skip qtremoteobjects -skip qtscript -skip qtscxml -skip qtsensors -skip qtserialbus -skip qtserialport -skip qtspeech -skip qtvirtualkeyboard -skip qtwayland -skip qtwebchannel -skip qtwebengine -skip qtwebglplugin -skip qtwebsockets -skip qtwebview -release
```

This results in the compile command and error:

```
[...]

PR #698, opened 2020-02-27
Add basic support for macOS frameworks

barche:
As discussed in https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/393#issuecomment-581182957, this adds basic support for macOS frameworks, limited to the following:
* Find a framework that matches the provided name with the `.framework` suffix
* Resolve the library that is inside the framework with its version

On non-macOS platforms, the `FrameworkProduct` behaves just like a `LibraryProduct`, so this is mainly intended for libraries like Qt that offer a Framework on macOS and normal libraries on other platforms.

### DIFF
--- a/src/Products.jl
+++ b/src/Products.jl
@@ -1,12 +1,12 @@
-export Product, LibraryProduct, FileProduct, ExecutableProduct, satisfied,
+export Product, LibraryProduct, FileProduct, ExecutableProduct, FrameworkProduct, satisfied,
        locate, write_deps_file, variable_name
 import Base: repr
 
 """
 A `Product` is an expected result after building or installation of a package.
 
-Examples of `Product`s include [`LibraryProduct`](@ref),
-[`ExecutableProduct`](@ref) and [`FileProduct`](@ref).  All `Product` types must
+Examples of `Product`s include [`LibraryProduct`](@ref), [`FrameworkProduct`](@ref),
+[`ExecutableProduct`](@ref) and [`FileProduct`](@ref). All `Product` types must
 define the following minimum set of functionality:
 
 * [`locate(::Product)`](@ref): given a `Product`, locate it within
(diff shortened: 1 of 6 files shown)


src/Products.jl
```diff
@@ -207,6 +207,86 @@ function locate(lp::LibraryProduct, prefix::Prefix; platform::Platform = platfor
     return nothing
 end
 
+"""
+A `FrameworkProduct` is a  [`Product`](@ref) that encapsulates a macOS Framework.
+It behaves mostly as a [`LibraryProduct`](@ref) and on non-macOS platforms falls back
+to the behavior of `LibraryProduct`, so there is no need to define the products
+differently on non-macOS platforms.
```
giordano: To be honest, I'm not overly happy about this in principle.  However, I guess we aren't going to use this product very often, and on the other hand this does make things much easier, so in the end I'm happy enough

src/Products.jl
```diff
@@ -207,6 +207,86 @@ function locate(lp::LibraryProduct, prefix::Prefix; platform::Platform = platfor
     return nothing
 end
 
+"""
+A `FrameworkProduct` is a  [`Product`](@ref) that encapsulates a macOS Framework.
```
giordano: Maybe you want to mention this product also in the docstring of `Product`?

src/Products.jl
```diff
@@ -207,6 +207,86 @@ function locate(lp::LibraryProduct, prefix::Prefix; platform::Platform = platfor
...
+                            dir_paths::Vector{<:AbstractString}=String[];
+                            kwargs...)
+        # If some other kind of AbstractString is passed in, convert it
+        return new(LibraryProduct(fwnames, varname, dir_paths; kwargs...))
+    end
+
+    FrameworkProduct(meta_obj::Dict) = new(LibraryProduct(meta_obj))
+end
+
+Base.:(==)(a::FrameworkProduct, b::FrameworkProduct) = (a.libraryproduct == b.libraryproduct)
```
giordano: Is this necessary?  I though that since `FrameworkProduct` is an immutable struct `==` would basically compare its fields.

src/Products.jl
barche: Done!

src/Products.jl
barche: I think so, because `FrameworkProduct` is not stored as a bits type. In the pushed update this is tested, and it doesn't work without the additional `==` method.

barche:
Wow, look at all the green checkmarks now ;)

staticfloat:
The only thing I'm hesitant about here is the fact that this "falls back" to LibraryProduct behavior when it can't find a true Framework.  I think I would rather force users to create `build_tarballs.jl` files that perform one invocation of `build_tarballs()` with FrameworkProducts, and one invocation with LibraryProducts, and then ensure that the user-facing API from the JLL side is the same; that way, `LibFoo_jll` can expose a `FrameworkProduct` on macOS where it makes sense, and a `LibraryProduct` elsewhere, but if we decide to add functionality to the `FrameworkProduct` we can.  I'm imagining someday it may be useful to be able to call something like `content_dir(libfoo)` to get the `Contents `directory, or somesuch.  And that won't make any sense if the `FrameworkProduct` is actually just a library.  However, if all you're going to do is use the product like a library, we can ensure that the user doesn't care what kind of library or framework it is, by just exporting the same API.

Other than that design issue, the actual implementation here looks good.  So just tweak that fallthrough behavior, and I think this is good to go.

barche:
OK, fair enough. The reason I made it fall through to a normal `LibraryProduct` is to make writing the build_tarballs script a bit simpler and also I was under the impression that on Yggdrasil the build_tarballs is parallel over the platforms, so a single call would be more efficient. But I agree that from a design point of view separating them is better.

barche:
I removed the fallthrough, so assuming the tests still pass this should be good to go. Qt built successfully with this, test repo here: https://github.com/barche/qt5_jll.jl

staticfloat:
> I was under the impression that on Yggdrasil the build_tarballs is parallel over the platforms, so a single call would be more efficient

We jump through [crazy hoops](https://github.com/JuliaPackaging/Yggdrasil/blob/master/azure-pipelines.yml#L93-L97) to get the best of both worlds.  ;)

PR #699, opened 2020-02-27
Add an audit warning if a package doesn't depend on CSL when it should

### LINKED ISSUE #448: Track dependency on libgomp
In order to properly solve #442, we should auto-determine linkage against "non-shipped" CSLs and explicitly require `CompilerSupportLibraries_jll` for those packages.

giordano:
Fix #448


giordano:
If a library depends on both libgomp and libgfortran, the warning about CSL would be printed twice.  But I think it's good to increase the probability for it to be read :slightly_smiling_face:

PR #701, opened 2020-02-29
Dlopen flags 2

barche:
This is an update of PR #453 with the following changes:

* Retain the `dont_dlopen` field and parameter name
* Fix a bug where instead of the library, it tried to dlopen a file with the variable name as name
* Some basic testing


staticfloat:
@giordano if this looks good to you, let's merge it.

staticfloat:
(Although this needs to be rebased to deal with the recent merge of the frameworks PR. :) )

barche:
OK, rebased and fingers crossed that the tests still pass.

PR #708, opened 2020-03-10
Add warning messages about deadlinks

staticfloat:
This should fix issues such as the recently-diagnosed `HELICS_jll`
issues installing things on windows, where `7z.exe` decided it didn't
want to extract tarballs with deadlinks.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -708,6 +708,9 @@ function autobuild(dir::AbstractString,
         # Unsymlink all the deps from the dest_prefix
         cleanup_dependencies(prefix, artifact_paths)
 
+        # Search for dead links; raise warnings about them.
+        warn_deadlinks(prefix.path)
+
         # Cull empty directories, for neatness' sake, unless auditing is disabled
         if !skip_audit
             for (root, dirs, files) = walkdir(dest_prefix.path; topdown=false)
--- a/src/auditor/symlink_translator.jl
+++ b/src/auditor/symlink_translator.jl
@@ -18,3 +18,20 @@ function translate_symlinks(root::AbstractString; verbose::Bool=false)
         end
     end
 end
+
+"""
+    warn_deadlinks(root::AbstractString)
+
+Walks through the given `root` directory, finding broken symlinks and warning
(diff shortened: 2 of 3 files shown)


src/auditor/symlink_translator.jl
```diff
@@ -28,7 +28,7 @@ copying a symlink that points to a dependency; by doing so, it implicitly
 breaks relocatability.
 """
 function warn_deadlinks(root::AbstractString)
-    for f in collect_files(root, islink)
+    for f in collect_files(root, islink; exclude_externalities=false)
```
giordano: :smirk:

src/AutoBuild.jl
```diff
@@ -707,6 +707,9 @@ function autobuild(dir::AbstractString,
         # Unsymlink all the deps from the dest_prefix
         cleanup_dependencies(prefix, artifact_paths)
 
+        # Search for dead links; raise warnings about them.
+        warn_deadlinks(prefix)
```
giordano: Should this be
```suggestion
        warn_deadlinks(prefix.path)
```
or something like that?  `prefix::Prefix`, but `warn_deadlinks` takes an `AbstractString` as input

giordano:
Should we specifically check if a link is pointing outside of the prefix, that is outside of the tarball?  That sounds like something strange.  This is to avoid links outside of the prefix that by accident point to an existing file

staticfloat:
> that by accident point to an existing file

I think you have it backwards, it checks for `!isfile()`.  I am doing this after the dependencies have been unlinked from the prefix, so hopefully this will find things that pointed to dependencies that are now gone.

If someone makes a symlink to `/bin/bash` that exists, I think that's...... okay.  I know that MacOS does this all the time, creating symlinks to `/System/Frameworks/foo`.

giordano:
Uhm, ok ok, makes sense, especially because of the macOS frameworks

giordano:
This doesn't really work as expected: for Unices we usually have symlinks of the form `libfoo.so` -> `libfoo.so.1.2.3`, which makes `isfile` return false, thus issuing the warning.

PR #710, opened 2020-03-12
Fix `dlopen_flags_str`

giordano:
The product can be also a `FrameworkProduct`, which doesn't have a
`dlopen_flags` field.


barche:
Looks good to me, approved! 🙃

PR #711, opened 2020-03-12
Fix warning about dead symlinks

giordano:
The current implementation shows a warning also if the link target is a relative
path to an existing file, because the check `isfile(link_target)` isn't
performed in the same directory as where the symlink is.

Addresses https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/708#issuecomment-597956747.


staticfloat:
I love how I get halfway toward writing a review comment saying "why not use `isabspath()`?" and then see that you're already one step ahead of me. :)

PR #713, opened 2020-03-14
Comment out extraneous `resolve_versions!()`

staticfloat:
Fredrik says this is unnecessary, and he's probably right, but I
distinctly remember adding this for a reason.  Let's comment it out and
see if things still work.  :P

### DIFF
--- a/src/Auditor.jl
+++ b/src/Auditor.jl
@@ -331,6 +331,25 @@ function check_dynamic_linkage(oh, prefix, bin_files;
 end
 
 
+if VERSION < v"1.5-DEV"
+function walkdir_nosymlinks(path::String)
+    function adjuster(out_c::Channel)
+        for (root, dirs, files) in walkdir(path)
+            for d in dirs
+                if islink(joinpath(root, d))
+                    push!(files, d)
+                end
+            end
+            filter!(d -> !islink(joinpath(root, d)), dirs)
+            put!(out_c, (root, dirs, files))
+        end
+    end
+    return Channel(adjuster)
+end
+else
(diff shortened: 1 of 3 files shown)


giordano:
For the record it was introduced in [this commit](https://github.com/JuliaPackaging/BinaryBuilder.jl/commit/9e5af673a05237569c7d2eeed3dfd74c307ede66), but it's not clear why it was necessary

staticfloat:
Yes yes, the smoking gun that I'm pointing to with my right hand is often in my left....

staticfloat:
Okay I figured out why this is here.  :P

When adding packages, we need to get their dependencies!  `Pkg.add()` does this internally, but it doesn't share with us.

@fredrikekre @KristofferC is there a way I can do something like `Pkg.add()` then get a list of all `PackageSpec`'s that were added?  I need to walk all `_jll` packages added, find their package source path and parse all the `Artifacts.toml` files so that I can symlink them all together into the build environment.

What I have before this PR works, but it complains when I do things like provide `PackageSpec`'s with a specific gitsha (to _force_ usage of a particular JLL), whereas `Pkg.add()` can deal with it just fine.  So I'd rather use `Pkg.add()` if I can, but I just need to get that internal list of PackageSpecs that is the full dependency tree post-installation.

KristofferC:
> When adding packages, we need to get their dependencies! Pkg.add() does this internally, but it doesn't share with us.

Isn't it enough to look in the manifest after `add`? It will have a list of all dependencies for all packages.

staticfloat:
Good call Kristoffer; I'll do that.  :)

PR #726, opened 2020-04-02
Fix a few typos

fingolfin:


src/wizard/github.jl
```diff
@@ -41,7 +41,7 @@ function obtain_token(; ins=stdin, outs=stdout, github_api=GitHub.DEFAULT_API)
         ],
         "note" => "BinaryBuilder.jl generated token at $(now())",
         "note_url" => "https://github.com/JuliaPackaging/BinaryBuilder.jl",
-        "fingerpint" => randstring(40)
+        "fingerprint" => randstring(40)
```
fingolfin, line 44: Note that this is a change in the code! But I found nothing that references either `fingerprint` nor `fingerpint`, so perhaps it could even be removed? But perhaps the reference is in code outside this repo; I didn't try to understand it... anyway, I thought I should point it out in case this typo exists elsewhere and deliberately matched here (or perhaps this isn't a typo but a reference to some special kind of "microbrew" ??? :-) )

PR #730, opened 2020-04-06
Experimental Azure logging

staticfloat:
Use `LoggingExtras` to override all `@info`/`@warn`/`@error` outputs to
also spit out Azure-compatible records.

Closes https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/513
Closes #627


Manifest.toml
```diff
@@ -232,9 +238,9 @@ version = "1.1.0"
 
 [[RegistryTools]]
 deps = ["AutoHashEquals", "LibGit2", "Pkg", "UUIDs"]
-git-tree-sha1 = "8c48f34b5884366b2b9e76311b38b2a6235f9095"
+git-tree-sha1 = "3dfa671318ac8af835cc621b08fcdcc28a4aee67"
 uuid = "d1eb7eb1-105f-429d-abf5-b0f65cb9e2c4"
-version = "1.3.1"
+version = "1.3.2"
```
giordano, line 243: While we are here, can we actually require v1.3.4 in Project.toml and update the Manifest.toml accordingly?  See https://github.com/JuliaPackaging/Yggdrasil/pull/777

src/Logging.jl
```diff
@@ -14,23 +14,25 @@ function Logging.handle_message(logger::AzureSinkLogger, args...; kwargs...)
     task::String = logger.action_chooser(log)
     
     properties=String[]
-    for key in (:level, :_module, :group, :id, :file, :line)
+    # AZP calls it `type` not level`, and `warning` not `warn`:
+    levelmap = Dict(Logging.Error => "error", Logging.Warn => "warn")
```
giordano, line 18:
```suggestion
    levelmap = Dict(Logging.Error => "error", Logging.Warn => "warning")
```

staticfloat: ......hilarious.  I can get it right in the commit message but not in the code?

staticfloat: My Japanese TA always chided me for making 'careless mistakes' in my tests.  I guess the legacy continues.

staticfloat:
@giordano if you're happy with this implementation, feel free to merge this.  Note the testing in https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=2697&view=results

giordano:
I'm (super!) happy, but is the license intended to stay as an `@error` or that was only a test?

PR #737, opened 2020-04-11
Add environment variables to {,not} link whole archive

giordano:
Since we often need to convert a static archive into a shared library, I'd like to have as environment variables the flags for the linker needed to do the conversion.  I know we don't export other flags in the same way, so this may not be desirable.

Actually, my ideal solution would be to have a bash command for doing this, but I think that would be a bit too complicated: you have to specify the compiler, the list of archives you want to put together -- ok, usually only one --, extra linker options and extra libraries.  This isn't super easy to do in bash without multiple dispatch :pensive:   This is why I think at least having the whole archive flags available as environment variables would help

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -32,6 +32,11 @@ exeext(p::Windows) = ".exe"
 exeext(p::Union{Linux,FreeBSD,MacOS}) = ""
 exeext(p::Platform) = error("Unknown exeext for platform $(p)")
 
+whole_archive(p::Platform) = "--whole-archive"
+whole_archive(p::MacOS) = "-all_load"
+nowhole_archive(p::Platform) = "--no-whole-archive"
+nowhole_archive(p::MacOS) = "-noall_load"
+
 # Convert platform to a triplet, but strip out the ABI parts.
 # Also translate `armv7l` -> `arm` for now, since that's what most
 # compiler toolchains call it.  :(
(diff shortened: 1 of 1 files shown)


staticfloat:
I can see lots of future flags we might want to include.  Things like rpath linking flags, or whatnot.  Why don't we start a new tool, something like `pkg-config` or `llvm-config` that we can ask for flags?

As an example, on Linux/FreeBSD, linking to something relative via `rpath` would be done via:

```
$ flagon --relative-rpath-link=/../lib
-Wl,-rpath=$ORIGIN/../lib -Wl,-rpath-link=$ORIGIN/../lib -Wl,-z,origin
```

Whereas, on MacOS it would be:
```
$ flagon --relative-rpath-link=/../lib
-Wl,-rpath=@loader_path/../lib -Wl,-rpath-link=@loader_path/../lib
```

And on Windows, it would be:
```
$ flagon --relative-rpath-link=/../lib

```

`flagon` (or `flag-config` or `flaggy` or `cflagg` or whatever we want to call it) would make its best guess using the appropriate default compiler (e.g. use `$CC` to inspect the current compiler, figure out if it's `gcc` or `clang`, use `${target}` to figure out what OS it's targeting, etc....) and you could override its choice by setting `CC=gcc flagon` on MacOS, for instance, to get `gcc`-compatible flags on MacOS.

To continue your example, we could then have a `flagon --no-whole-archive` and a `flagon --whole-archive` that would do the right thing across all compilers.

giordano:
Superseded by https://github.com/JuliaPackaging/Yggdrasil/pull/937

PR #739, opened 2020-04-12
Add "Wizard service" to modify Wizard state during interactive build

Keno:
Say I build a library and then discover that it actually depends
on Zlib. At the moment, I have to go back to the start of the
wizard and declare my additional dependency. It would be nice
to just request that the wizard add it from inside the environment.
This is what this PR implements by exposing an AF_UNIX server that
responds to requests from the sandbox.
With a suitable `bb` command, the following works from inside the sandbox:
```
sandbox:${WORKSPACE}/srcdir # bb add Zlib_jll
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Resolving package versions...
   Updating `/tmp/jl_q0IePp/mRAcrdQ3/.project/Project.toml`
  [83775a58] + Zlib_jll v1.2.11+9
   Updating `/tmp/jl_q0IePp/mRAcrdQ3/.project/Manifest.toml`
  [83775a58] + Zlib_jll v1.2.11+9
  [2a0f44e3] + Base64
  [ade2ca70] + Dates
  [b77e0a4c] + InteractiveUtils
  [76f85450] + LibGit2
  [8f399da3] + Libdl
  [56ddb016] + Logging
  [d6f4376e] + Markdown
  [44cfe95a] + Pkg
  [de0858da] + Printf
  [3fa0cd96] + REPL
  [9a3f8284] + Random
  [ea8e919c] + SHA
  [9e88b42a] + Serialization
  [6462fe0b] + Sockets
  [cf7118a7] + UUIDs
  [4ec0a83e] + Unicode
sandbox:${WORKSPACE}/srcdir #
```

Ideally, that `bb` command, could just be a simple wrapper over
netcat. However, that runs into https://github.com/JuliaLang/julia/issues/35442.
I will add a suitable pure-C version of that BB command to the RootFS
instead. This is just the BinaryBuilder version.

### DIFF
--- a/src/wizard/interactive_build.jl
+++ b/src/wizard/interactive_build.jl
@@ -199,6 +199,87 @@ function diff_srcdir(state::WizardState, prefix::Prefix, ur::Runner)
     return false
 end
 
+function bb_add(client, state::WizardState, prefix::Prefix, platform::Platform, jll::AbstractString)
+    if any(dep->getpkg(dep).name == jll, state.dependencies)
+        println(client, "ERROR: Package was already added")
+        return
+    end
+    new_dep = Dependency(jll)
+    try
+        # This will redo some work, but that may be ok
+        concrete_platform = get_concrete_platform(platform;
+                                                  preferred_gcc_version = state.preferred_gcc_version,
+                                                  preferred_llvm_version = state.preferred_llvm_version,
+                                                  compilers = state.compilers)
+        setup_dependencies(prefix, getpkg.([state.dependencies; new_dep]), concrete_platform)
+        push!(state.dependencies, new_dep)
+    catch e
+        showerror(client, e)
+    end
+end
+
+using ArgParse
+
+function bb_parser()
+    s = ArgParseSettings()
+
+    @add_arg_table! s begin
+        "add"
+            help = "Add a jll package as if specified at the start of the wizard"
+            action = :command
+    end
+
+    @add_arg_table! s["add"] begin
+        "jll"
+        help = "The jll to add"
+        required = true
+    end
+
+    s
+end
+
+function setup_bb_service(state::WizardState, prefix, platform)
+    fpath = joinpath(prefix, "metadir", "bb_service")
+    server = listen(fpath)
+    @async begin
+        while isopen(server)
+            client = accept(server)
+            function client_error(err)
+                println(client, "ERROR: $err")
+                close(client)
+            end
+            @async while isopen(client)
+                try
+                    cmd = readline(client)
+                    ARGS = split(cmd, " ")
+                    s = bb_parser()
+                    try
+                        popfirst!(ARGS)
+                        parsed = parse_args(ARGS, s)
+                        if parsed == nothing
+                        elseif parsed["%COMMAND%"] == "add"
+                            bb_add(client, state, prefix, platform, parsed["add"]["jll"])
+                        end
+                        close(client)
+                    catch e
+                        if isa(e, ArgParseError)
+                            println(client, "ERROR: ", e.text, "\n")
+                            ArgParse.show_help(client, s)
+                            close(client)
+                        else
+                            rethrow(e)
+                        end
+                    end
+                catch e
+                    showerror(stderr, e)
+                    close(client)
+                end
+            end
+        end
+    end
+    (fpath, server)
+end
+
 """
     interactive_build(state::WizardState, prefix::Prefix,
                       ur::Runner, build_path::AbstractString)
@@ -207,7 +288,8 @@ end
     reproducible steps for building this source. Shared between steps 3 and 5
 """
 function interactive_build(state::WizardState, prefix::Prefix,
-                           ur::Runner, build_path::AbstractString;
+                           ur::Runner, build_path::AbstractString,
+                           platform::Platform;
                            hist_modify = string, srcdir_overlay = true)
     histfile = joinpath(prefix, "metadir", ".bash_history")
     cmd = `/bin/bash -l`
@@ -226,6 +308,8 @@ function interactive_build(state::WizardState, prefix::Prefix,
             cmd = `/bin/bash -c $cmd`
         end
 
+        (fpath, server) = setup_bb_service(state, prefix, platform)
+
         script_successful = run_interactive(ur, ignorestatus(cmd), stdin=state.ins, stdout=state.outs, stderr=state.outs)
 
         had_patches = srcdir_overlay && diff_srcdir(state, prefix, ur)
@@ -236,6 +320,10 @@ function interactive_build(state::WizardState, prefix::Prefix,
         workdir = joinpath(prefix, "metadir", "work", "work")
         if isdir(workdir)
             rm(workdir)
+        end
+        if @isdefined server
+            close(server)
+            rm(fpath)
         end
     end
 
@@ -301,7 +389,7 @@ function step3_interactive(state::WizardState, prefix::Prefix,
                            platform::Platform,
                            ur::Runner, build_path::AbstractString, artifact_paths::Vector{String})
 
-    if interactive_build(state, prefix, ur, build_path)
+    if interactive_build(state, prefix, ur, build_path, platform)
         # Unsymlink all the deps from the dest_prefix before moving to the next step
         cleanup_dependencies(prefix, artifact_paths)
         state.step = :step3_retry
@@ -520,7 +608,7 @@ function step5_internal(state::WizardState, platform::Platform)
                     )
 
                     if choice == 1
-                        if interactive_build(state, prefix, ur, build_path;
+                        if interactive_build(state, prefix, ur, build_path, platform;
                                           hist_modify = function(olds, s)
                             """
                             $olds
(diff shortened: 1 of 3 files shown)


giordano:
Is the command recorded in the history, and hence in the build script?

Keno:
> Is the command recorded in the history, and hence in the build script?

Yes, you'll have to edit it out. Bonus points for a follow up commit that suppresses it from history, or a bash macro that deletes it from the history right away ;).

giordano:
Yeah, that's not hard

src/wizard/interactive_build.jl
```diff
@@ -184,6 +184,60 @@ function diff_srcdir(state::WizardState, prefix::Prefix, ur::Runner)
...
+        setup_dependencies(prefix, getpkg.([state.dependencies; new_dep]), platform)
+        push!(state.dependencies, new_dep)
+    catch e
+        showerror(client, e)
+    end
+end
+
+function setup_bb_service(state::WizardState, prefix, platform)
+    fpath = joinpath(prefix, "metadir", "bb_service")
+    server = listen(fpath)
```
staticfloat, line 204: What OSes do we expect this to work on? I know Linux is the first class citizen, but in my testing it doesn't look like this works on MacOS or Windows.  Perhaps we should use a TCP port instead.

src/wizard/interactive_build.jl
```diff
@@ -184,6 +184,60 @@ function diff_srcdir(state::WizardState, prefix::Prefix, ur::Runner)
...
+                            if length(ARGS) == 2
+                                println(client, "Usage: bb add <jll>")
+                                close(client)
+                            end
+                            bb_add(client, state, prefix, platform, ARGS[3])
+                            close(client)
+                        else
+                            client_error("Unrecognized subcommand $(ARGS[2])")
+                        end
+                    else
```
staticfloat, line 228: We should add a `bb`/`bb help`/`bb --help` that lists all options and examples.

src/wizard/interactive_build.jl
```diff
@@ -184,6 +184,60 @@ function diff_srcdir(state::WizardState, prefix::Prefix, ur::Runner)
...
+        println(client, "ERROR: Package was already added")
+        return
+    end
+    new_dep = Dependency(jll)
+    try
+        # This will redo some work, but that may be ok
+        setup_dependencies(prefix, getpkg.([state.dependencies; new_dep]), platform)
+        push!(state.dependencies, new_dep)
+    catch e
+        showerror(client, e)
```
staticfloat, line 198: Is this a sufficiently helpful error message when `jll` doesn't exist in the registry?

src/wizard/interactive_build.jl
Keno: It works on all OSes on the julia side. There's a question of how to forward it to the runner. Not sure docker can forward named pipes (it's not hard, but it is a bit of code). I would prefer this approach where possible, since TCP sockets kinda need to be authenticated to not open up a security issue. I was hoping to punt on that for now.

src/wizard/interactive_build.jl
Keno: Yes, you get the same error message you would as if you tried to `Pkg.add` it.

src/wizard/interactive_build.jl
Keno: I guess at that point, we might as well just use ArgParse.jl directly?

staticfloat: Sure

src/wizard/interactive_build.jl
staticfloat: > It works on all OSes on the julia side.

What am I doing wrong?

```
$ cat listen.jl 
using Sockets
listen("./foo")

$ julia listen.jl
ERROR: LoadError: ArgumentError: could not listen on path ./foo
Stacktrace:
 [1] listen(::String) at /Users/sabae/tmp/julia-build/julia-release-1.4/usr/share/julia/stdlib/v1.4/Sockets/src/PipeServer.jl:76
 [2] top-level scope at /Users/sabae/listen.jl:2
 [3] include(::Module, ::String) at ./Base.jl:377
 [4] exec_options(::Base.JLOptions) at ./client.jl:288
 [5] _start() at ./client.jl:484
in expression starting at /Users/sabae/listen.jl:2
```

staticfloat: It works on Linux, BTW.

Keno: Do you already have a "foo" in your current directory?

staticfloat: Ah, you're right, that was the problem on MacOS.  But on Windows, I still get the same error:

```
julia> listen("foo")
ERROR: ArgumentError: could not listen on path foo
Stacktrace:
 [1] listen(::String) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.3\Sockets\src\PipeServer.jl:76
[2] top-level scope at REPL[8]:1

julia>

Administrator@win64bot3 ~
$ ls -la foo
ls: cannot access 'foo': No such file or directory
```

staticfloat: > TCP sockets kinda need to be authenticated to not open up a security issue

Binding to `localhost` on a randomized port (provided to the client through an environment variable) would address most of the security concerns, I think.

Keno: Windows is complicated. You need to use the pipe namespace. I guess that means we should make this unix only for now.

Keno: > Binding to localhost on a randomized port (provided to the client through an environment variable) would address most of the security concerns, I think.

Not on a shared machine. I can scan all of localhost in about 50ms. We had a CVE on julia because we used to do that.

Keno:
Now with more ArgParse:
```
sandbox:${WORKSPACE}/srcdir # ./bb add Zlib_jll
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Resolving package versions...
Updating `/tmp/jl_OYcgTh/qwT8cFf6/.project/Project.toml`
  [83775a58] + Zlib_jll v1.2.11+9
Updating `/tmp/jl_OYcgTh/qwT8cFf6/.project/Manifest.toml`
  [83775a58] + Zlib_jll v1.2.11+9
  [2a0f44e3] + Base64
  [ade2ca70] + Dates
  [b77e0a4c] + InteractiveUtils
  [76f85450] + LibGit2
  [8f399da3] + Libdl
  [56ddb016] + Logging
  [d6f4376e] + Markdown
  [44cfe95a] + Pkg
  [de0858da] + Printf
  [3fa0cd96] + REPL
  [9a3f8284] + Random
  [ea8e919c] + SHA
  [9e88b42a] + Serialization
  [6462fe0b] + Sockets
  [cf7118a7] + UUIDs
  [4ec0a83e] + Unicode
sandbox:${WORKSPACE}/srcdir # ./bb
ERROR: no command given

usage: <PROGRAM> {add}

commands:
  add  Add a jll package as if specified at the start of the wizard

sandbox:${WORKSPACE}/srcdir # ./bb add
ERROR: required argument jll was not provided

usage: <PROGRAM> {add}

commands:
  add  Add a jll package as if specified at the start of the wizard

sandbox:${WORKSPACE}/srcdir # ./bb add Zlib_jll
ERROR: Package was already added
sandbox:${WORKSPACE}/srcdir # ./bb --help
usage: <PROGRAM> [-h] {add}

commands:
  add         Add a jll package as if specified at the start of the
              wizard

optional arguments:
  -h, --help  show this help message and exit
```

src/wizard/interactive_build.jl
```diff
@@ -184,6 +184,83 @@ function diff_srcdir(state::WizardState, prefix::Prefix, ur::Runner)
...
 
+function bb_add(client, state::WizardState, prefix::Prefix, platform::Platform, jll::AbstractString)
+    if any(dep->getpkg(dep).name == jll, state.dependencies)
+        println(client, "ERROR: Package was already added")
+        return
+    end
+    new_dep = Dependency(jll)
+    try
+        # This will redo some work, but that may be ok
+        setup_dependencies(prefix, getpkg.([state.dependencies; new_dep]), platform)
```
giordano, line 195: This should now be
```suggestion
        concrete_platform = get_concrete_platform(platform;
                                                  preferred_gcc_version = state.preferred_gcc_version,
                                                  preferred_llvm_version = state.preferred_llvm_version,
                                                  compilers = state.compilers)
        setup_dependencies(prefix, getpkg.([state.dependencies; new_dep]), concrete_platform)
```
see #743.

giordano:
Can this work also in any interactive shell not just the wizard? Sometimes when I'm in a debug shell I realise I need extra dependencies

staticfloat:
> Can this work also in any interactive shell not just the wizard?

As it stands, no, because it's tangled up with the WizardState to record the dependency into the list of needed packages.  I would be in favor of disentangling that and having the `bb` command available in every invocation of `run_interactive()`, but we'll have to put that on the back burner for now.

PR #745, opened 2020-04-13
Correctly print the search path of a LibraryProduct in the wizard

giordano:
We were doing this already for the `ExecutableProduct`, but not for a `LibraryProduct`

### DIFF
--- a/src/wizard/deploy.jl
+++ b/src/wizard/deploy.jl
@@ -39,8 +39,12 @@ function print_build_tarballs(io::IO, state::WizardState)
                 return "ExecutableProduct($(repr(basename(file))), $(repr(varname)), $(repr(dir_path)))"
             end
         elseif kind == :library
-            # Normalize library names, since they are almost always in `lib`/`bin`
-            return "LibraryProduct($(repr(normalize_name(file))), $(repr(varname)))"
+            dir_path = dirname(file)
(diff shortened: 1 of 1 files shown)


giordano:
With this PR, I get this for the builder of fakechroot (https://github.com/JuliaPackaging/Yggdrasil/pull/820):
```julia
julia> BinaryBuilder.print_build_tarballs(stdout, state)
# Note that this script can accept some limited command-line arguments, run
# `julia build_tarballs.jl --help` to see a usage message.
using BinaryBuilder, Pkg

name = "fakechroot"
version = v"2.20.2"

# Collection of sources required to complete build
sources = [
    GitSource("https://github.com/dex4er/fakechroot.git", "b42d1fb9538f680af2f31e864c555414ccba842a")
]

# Bash recipe for building across all platforms
script = raw"""
cd $WORKSPACE/srcdir
cd fakechroot/
./autogen.sh 
make -j${nproc}
make install
"""

# These are the platforms we will build for by default, unless further
# platforms are passed in on the command line
platforms = supported_platforms()

# The products that we will ensure are always built
products = [
    LibraryProduct("libfakechroot", :libfakechroot, "lib/fakechroot")
]

dependencies = Dependency[
]

# Build the tarballs, and possibly a `build.jl` as well.
build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
```

PR #747, opened 2020-04-13
Fix looking up paths in Yggdrasil

### LINKED ISSUE #631: Warn the user if the build already exists in Yggdrasil
In https://github.com/JuliaPackaging/Yggdrasil/pull/441 Valentin opened a PR for a library which was actually already in Yggdrasil.

When the user is prompted for the name of the package in the wizard it might be good to check if this is already available in Yggdrasil and ask whether they want to continue anyway: one for example can build a newer version of the package.

Name checking is a bit tricky since we can choose different names, for example with or without "lib", or with different capitalisation (this latter problem can of course be worked around by checking the lowercase names).

### LINKED ISSUE #684: `get_yggdrasil` never updates the repo
This https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/a994f7e8a57ae8ffcf2c62f13d0b3937ff05e339/src/wizard/yggdrasil.jl#L3-L19 only runs `git fetch`, but not `git merge`, so the repo is never actually updated.  Ref: https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/631#issuecomment-589906002

Keno:
Fixes #684, fixes #631

### DIFF
--- a/src/wizard/obtain_source.jl
+++ b/src/wizard/obtain_source.jl
@@ -322,7 +322,7 @@ function obtain_source(state::WizardState)
 end
 
 function get_name_and_version(state::WizardState)
-    yggdrasil_path = get_yggdrasil()
+    ygg = get_yggdrasil()
 
     while state.name === nothing
         msg = "Enter a name for this project.  This will be used for filenames:"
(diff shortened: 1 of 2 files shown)


giordano:
I assume this should fix #631, not #684?

Keno:
Well, both #684 isn't really an issue if we don't care about the local refspecs.

Keno:
Unless there's some other place where we want to use them also, in which case we should fix that.

giordano:
The fact is that my local yggdrasil bare clone is stuck at a very old revision and is never updated, so this check won't tell me if I'm trying to (re)build a recently added recipe.

giordano:
Nevermind, it looks like it works also with recently recipes!

giordano:
I see, you're looking at the tree in `origin/master`, so what's in the local checkout is not important.

Keno:
> I see, you're looking at the tree in `origin/master`, so what's in the local checkout is not important.

Yes, we should just be completely ignoring any local refspecs in operations on the repo.

PR #755, opened 2020-04-19
Compress log files in the tarball

giordano:
I'd like to be able to keep shipping the log files in the tarball, as they are very valuable for later debugging of the build.  Compressing them should address some complaints about their large size in the artifacts directory.

### DIFF
--- a/src/Prefix.jl
+++ b/src/Prefix.jl
@@ -2,7 +2,7 @@
 #  on disk.  Things like the name of where downloads are stored, and what
 #  environment variables must be updated to, etc...
 import Base: convert, joinpath, show, withenv
-using SHA
+using SHA, CodecZlib
 
 export Prefix, bindir, libdirs, includedir, logdir, activate, deactivate,
        isinstalled,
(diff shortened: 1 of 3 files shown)


ViralBShah:
It won’t make people happy as it probably won’t change download size (which is always a constant source of complaints), but it will at least take lesser space on disk.

giordano:
The logs are plain text files, they can be compressed very effectively compared to the binary blobs in the tarballs.  Their main "issue" is their uncompressed size, which however in most cases scales with the size of the binary products and it's usually dwarfed by it.  This PR further reduces this problem.

For example, for OpenBLAS (which in https://github.com/JuliaPackaging/Yggdrasil/issues/864 is reported to have a large log file), for my platform the uncompressed size of the log files is:
```
% ls -lh logs 
total 5.7M
-r--r--r-- 1 mose mose 5.7M Apr 19 14:17 OpenBLAS.log
-r--r--r-- 1 mose mose  199 Apr 19 14:17 relink_to_rpath_libopenblas64_.0.3.9.so.log
```
With this compression they would become:
```
% ls -lh logs                       
total 348K
-rw-r--r-- 1 mose mose 343K Apr 19 14:19 OpenBLAS.log.gz
-rw-r--r-- 1 mose mose  165 Apr 19 14:19 relink_to_rpath_libopenblas64_.0.3.9.so.log.gz
```
For comparison, the size of the library is
```
% ls -lh lib/libopenblas64_.0.3.9.so 
-r-xr-xr-x 1 mose mose 29M Apr 19 14:17 lib/libopenblas64_.0.3.9.so
```

ViralBShah:
FWIW, I don't mind the extra download size, and keeping them compressed on disk means they won't contribute much to the size of people's `.julia`.

src/Prefix.jl
```diff
@@ -475,3 +475,33 @@ function cleanup_dependencies(prefix::Prefix, artifact_paths)
...
+                      verbose::Bool = false)
+    if isdir(dir)
+        if verbose
+            @info "Compressing files in $(dir)"
+        end
+        for f in readdir(dir)
+            filename = joinpath(dir, f)
+            if isfile(filename)
+                text = read(filename, String)
+                open(compression_stream, filename * extension, "w") do stream
```
staticfloat, line 500: I haven't figured out how to do this with the `open() do` form yet, but we should probably compress with level 9.

giordano: Done!  For the OpenBLAS example [above](https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/755#issuecomment-616135648) there isn't a large improvement though:
```
shell> ls -lh
total 336K
-rw-r--r-- 1 mose mose 331K Apr 20 22:34 OpenBLAS.log.gz
-rw-r--r-- 1 mose mose  165 Apr 20 22:34 relink_to_rpath_libopenblas64_.0.3.9.so.log.gz
```

staticfloat: Every little bit helps.  :)

PR #760, opened 2020-04-20
Fix warning about existing project in the wizard

giordano:
As is it now, pressing "n" at the question whether you want to choose a
different name breaks the `while` loop, leaving `state.name` undefined.  We
should instead continue the loop if we say we want to choose a different name.

I even managed to write a test for this!


staticfloat:
Good catch!

PR #761, opened 2020-04-20
Allow deploying a partial wizard state

### LINKED ISSUE #589: Allow exporting build_tarballs.jl even from an unfinished wizard state
Related to https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/222, if the user kills a build partway through (due to, e.g., a failure), it would be nice to allow them to still generate a build_tarballs.jl file with just the state as recorded so far.

This is especially useful if you've already gone through a couple of the platforms and things worked as expected, and you just want to rush through to the finish line; and/or try something else but not lose all your work so far; and/or maybe the build failed, but you don't know why, and want to upload the result anyway to share it with others and get some help.

This could perhaps be an option when you resume `run_wizard()` after exiting, in addition to "resume," there could be an "export build_tarballs.jl from current state" option or something?

giordano:
Fix #589

### DIFF
--- a/src/wizard/deploy.jl
+++ b/src/wizard/deploy.jl
@@ -25,30 +25,34 @@ function print_build_tarballs(io::IO, state::WizardState)
         """
     end
 
-    stuff = collect(zip(state.files, state.file_kinds, state.file_varnames))
-    products_string = join(map(stuff) do x
-        file, kind, varname = x
+    if state.files === nothing
+        products_string = "Product[\n]"
+    else
+        stuff = collect(zip(state.files, state.file_kinds, state.file_varnames))
+        products_string = "[\n    " * join(map(stuff) do x
+            file, kind, varname = x
 
-        if kind == :executable
-            # It's very common that executable products are in `bin`,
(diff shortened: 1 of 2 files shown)


src/wizard/deploy.jl
```diff
@@ -260,3 +257,26 @@ function step7(state::WizardState)
...
+        println(state.outs, "This state has been already deployed")
+        return
+    end
+    if state.step in (:step1, :step2)
+        println(state.outs, "This state cannot be deployed")
+        return
+    end
+    q = "This is an incomplete state, are you sure you want to deploy it?"
+    if yn_prompt(state, q, :n) == :y
+        _deploy(state)
```
giordano, line 280: I'm undecided whether to reset the saved state at this point (for my own record: this can be done with `BinaryBuilder.save_last_wizard_state(BinaryBuilder.WizardState())`) or not.  For the time being I went with not resetting, in case the user wanted only to print the partial script and try to resume it later.

staticfloat: I think that's fine.

PR #764, opened 2020-04-25
Throw an error if a variable name is already defined in Base

giordano:
This should alleviates issues like https://github.com/JuliaPackaging/Yggdrasil/pull/930#issuecomment-618997706, at least for symbols in Base -- at least for the version of Julia used to build the package.  I'm happy also to make this an error instead of a simple warning.

### DIFF
--- a/src/wizard/interactive_build.jl
+++ b/src/wizard/interactive_build.jl
@@ -101,23 +101,34 @@ function step4(state::WizardState, ur::Runner, platform::Platform,
     println(state.outs, "Please provide a unique variable name for each build artifact:")
     for f in state.files
         default = basename(f)
-        if Base.isidentifier(default)
-            varname = line_prompt(
-                "variable name",
-                string(f, " (default '$(default)'):");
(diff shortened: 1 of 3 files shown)


staticfloat:
Yes, I think this should be an error.

staticfloat:
Is this going to play nicely when we are giving product names in the wizard?

src/wizard/interactive_build.jl
```diff
@@ -101,23 +101,34 @@ function step4(state::WizardState, ur::Runner, platform::Platform,
...
+            else
+                varname = nonempty_line_prompt(
+                    "variable name",
+                    string(f, ":");
+                    force_identifier=true,
+                    ins=state.ins,
+                    outs=state.outs,
+                )
+            end
+            if isdefined(Base, Symbol(varname))
```
giordano, line 125: I've added a check in the wizard to prevent the user from setting a name that is already defined in `Base`.

PR #766, opened 2020-04-27
Add AnyPlatform that matches all platforms

### LINKED ISSUE #455: Define an Any platform that matches all platforms
In [`Yggdrasil`](https://github.com/JuliaPackaging/Yggdrasil/) we have now a few platform-independent packages, like all `Xorg_*proto` (they're just a bunch of header files).  To save a little bit of compilation time and space on disk we could have an `Any` platform that matches all platforms, so that we can build a single tarball and is it everywhere.  This kind of match-them-all target is used also in GNU/Linux distros, like Debian and Arch Linux (I'm familiar with these, but probably other distros adopt the same strategy)

giordano:
This allows us building a single tarball for platform-independent packages, like
those containing only header files.

Fix #455.


giordano:
I _think_ this is working.  Unfortunately it's a bit hackish because the JLL package can't really deal with platforms unknown to `Pkg.jl`, the easiest thing I could do was to generate a wrapper for all supported platforms and adding an entry in `Artifacts.toml` for all of them, but all actually using the same tarball.  It's not the best solution, but at least we save building time and and storage.

src/AutoBuild.jl
```diff
@@ -988,6 +997,18 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
+        # very hard to write a single "any" wrapper that can be easily resolved
+        # by current version of `Pkg.jl` without hacking the generated JLL
+        # package.  The easiest thing to do is to generate the wrappers for all
+        # the supported platforms, and populate `Artifacts.toml` with an entry
+        # for all platforms, all refererencing the same single tarball.  Until
+        # the far away day we'll have proper support for `AnyPlatform` also on
+        # the JLL side.
+        platforms = supported_platforms()
+        build_output_meta = Dict(p => build_output_meta[AnyPlatform()] for p in platforms)
+    end
```
staticfloat, line 1011: Can't we instead create an Artifact that is not platform-specific?  I don't think we need to teach Pkg about AnyPlatform at all.

giordano: > Can't we instead create an Artifact that is not platform-specific?

This is what I tried initially, and it failed in several places, because [`select_platform`](https://github.com/JuliaLang/Pkg.jl/blob/581e70c64ba1e409c292b190cabbf99530b2142a/src/BinaryPlatforms.jl#L769-L796) and [`do_artifact""`](https://github.com/JuliaLang/Pkg.jl/blob/581e70c64ba1e409c292b190cabbf99530b2142a/src/Artifacts.jl#L1005-L1024): when they have to do with an unknown platform they go haywire.

giordano:
Updated to make the artifact truly platform-independent, see the result at https://github.com/giordano/Xorg_xorgproto_jll.jl, in particular [`Artifacts.toml`](https://github.com/giordano/Xorg_xorgproto_jll.jl/blob/421b47be3913104d3e08eca870f48606740177dd/Artifacts.toml) and [`src/Xorg_xorgproto_jll.jl`](https://github.com/giordano/Xorg_xorgproto_jll.jl/blob/421b47be3913104d3e08eca870f48606740177dd/src/Xorg_xorgproto_jll.jl).

src/AutoBuild.jl
```diff
@@ -1207,13 +1221,18 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
-        # Filter platforms based on what wrappers we've generated on-disk
-        filter!(p -> isfile(joinpath(@__DIR__, "wrappers", replace(triplet(p), "arm-" => "armv7l-") * ".jl")), platforms)
+            # Filter platforms based on what wrappers we've generated on-disk
+            filter!(p -> isfile(joinpath(@__DIR__, "wrappers", replace(triplet(p), "arm-" => "armv7l-") * ".jl")), platforms)
 
-        # From the available options, choose the best platform
-        best_platform = select_platform(Dict(p => triplet(p) for p in platforms))
+            # From the available options, choose the best platform
+            best_platform = select_platform(Dict(p => triplet(p) for p in platforms))
+        end
```
giordano, line 1235: The ideal solution here would be to directly write
```julia
include(joinpath(@__DIR__, "wrappers", "any.jl"))
```
for `AnyPlatform` and what was here otherwise, but in this function we don't know the list of platforms and I'd like to avoid breaking registration again :upside_down_face:

staticfloat: It will be illegal to have a mixture of `AnyPlatform` and anything else; so I think we do know the list of platforms here.  If it's not `AnyPlatform`, do the normal thing.  If it is `AnyPlatform`, write out only the one file.

giordano: Of course I was wrong :upside_down_face:

giordano:
This is how the package looks like: https://github.com/giordano/Xorg_xorgproto_jll.jl/blob/e4969c4cf73f41b75d99a8d793d4a96123cff20a/src/Xorg_xorgproto_jll.jl

staticfloat:
yes yes good my young apprentice

PR #768, opened 2020-04-29
Tell the user how to accept Xcode EULA when non-interactive

fingolfin:
See https://github.com/JuliaPackaging/Yggdrasil/issues/948

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -189,6 +189,12 @@ function mount(cs::CompilerShard, build_prefix::AbstractString; verbose::Bool = 
             This is not an interactive Julia session, so we will not prompt you
             to download and install the macOS SDK.  Because you have not agreed
             to the Xcode license terms, we will not be able to build for MacOS.
+            Please read the Apple and Xcode SDK agreement for more information
+            on the restrictions and legal terms you agree to when using the SDK
+            to build software: https://images.apple.com/legal/sla/docs/xcode.pdf.
+            Should you then wish to agree to them, you can do so by setting the
+            environment variable BINARYBUILDER_AUTOMATIC_APPLE to `true` and then
+            restarting this computation.
             """)
             @error(msg)
             error("macOS SDK not installable")


src/Rootfs.jl
```diff
@@ -189,6 +189,8 @@ function mount(cs::CompilerShard, build_prefix::AbstractString; verbose::Bool =
             This is not an interactive Julia session, so we will not prompt you
             to download and install the macOS SDK.  Because you have not agreed
             to the Xcode license terms, we will not be able to build for MacOS.
+            You can set the BINARYBUILDER_AUTOMATIC_APPLE environment variable
+            to `true` and then restart this computation to accept the license.
```
giordano, line 193: I think the idea should be to let the user know about the EULA to make an informed decision about whether they want to accept it or not, rather then saying "just set this variable to `true`"

fingolfin: Of course. Although I think "accepting the license" kinda makes it clear that you need to think about whether you really want to do so. But, sure, I can make it a bigger warning. Some ideas:

1. Extend the verbiage, add something like this (to be improved, just meant as outlining the general gist)
   > Doing so comes with legal obligations which you should carefully consider by studying the EULA terms carefully. You can find them at <https://images.apple.com/legal/sla/docs/xcode.pdf>

2. I could point them to the FAQ entry then, I guess? But I am worried about how stable an URL <https://juliapackaging.github.io/BinaryBuilder.jl/dev/FAQ/#What-is-this-I-hear-about-the-macOS-SDK-license-agreement?-1> is.

3. I could just mention the env var name `BINARYBUILDER_AUTOMATIC_APPLE` and refer the user to the documentation. Alas, the documentation says nothing about "the terms" either, so that's really not better. So I guess the existing documentation also fails the standard you are setting for this PR ;-)

fingolfin: To be clear, I'll be happy to implement either of the 3 options, or some other, just let me know which you prefer

staticfloat: I think (1) is the best here; just point them directly to the Xcode SLA.

fingolfin: Done, by taking the message shown for interactive use.

PR #775, opened 2020-05-05
[JLL gen]: Set default value for `DYLD_FALLBACK_LIBRARY_PATH`

staticfloat:
Simon Byrne reported that on MacOS 10.13-, setting
`DYLD_FALLBACK_LIBRARY_PATH` failed to load `libc++`, which is referred
to by `@rpath/libc++.dylib` by `libLLVM`.  This is most likely because
these older MacOSs have a default `DYLD_FALLBACK_LIBRARY_PATH` that gets
overridden, and the new value searches for `libc++.dylib`
unsuccessfully.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1083,8 +1083,9 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
                         end
                     end
                     if adjust_LIBPATH
-                        if !isempty(get(ENV, LIBPATH_env, expanduser(LIBPATH_default)))
-                            env_mapping[LIBPATH_env] = string(LIBPATH, $(repr(pathsep)), get(ENV, LIBPATH_env, expanduser(LIBPATH_default)))
+                        LIBPATH_base = get(ENV, LIBPATH_env, expanduser(LIBPATH_default))
+                        if !isempty(LIBPATH_base)
+                            env_mapping[LIBPATH_env] = string(LIBPATH, $(repr(pathsep)), LIBPATH_base)
                         else
                             env_mapping[LIBPATH_env] = LIBPATH
                         end
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1034,12 +1034,15 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
             # The LIBPATH is called different things on different platforms
             if platform isa Windows
                 LIBPATH_env = "PATH"
+                LIBPATH_default = ""
                 pathsep = ';'
             elseif platform isa MacOS
                 LIBPATH_env = "DYLD_FALLBACK_LIBRARY_PATH"
+                LIBPATH_default = "~/lib:/usr/local/lib:/lib:/usr/lib"
                 pathsep = ':'
             else
                 LIBPATH_env = "LD_LIBRARY_PATH"
+                LIBPATH_default = ""
                 pathsep = ':'
             end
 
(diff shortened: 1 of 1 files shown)


staticfloat:
@simonbyrne can I get you to try modifying your local MPICH package to change this line:

https://github.com/JuliaBinaryWrappers/MPICH_jll.jl/blob/master/src/wrappers/x86_64-apple-darwin14-libgfortran4.jl#L70

to this:

```
if !isempty(get(ENV, LIBPATH_env, expanduser("~/lib:/usr/local/lib:/lib:/usr/lib"))
```

You'll need to do it for whatever libgfortran version you're actually loading (you can check via `Pkg.BinaryPlatforms.platform_key_abi()`).  This should populate the `DYLD_FALLBACK_LIBRARY_PATH` with the proper defaults, I think.

staticfloat:
Oh wait, darn, this only fails on CI for you.  Ugh.  Nevermind.

staticfloat:
Okay, I pushed [a custom branch of `MPICH_jll`](https://github.com/JuliaBinaryWrappers/MPICH_jll.jl/tree/sf/dyld_testing) could you try it out on CI?

giordano:
Doesn't the default value need to be appended to the rest on macOS?

staticfloat:
That's what this should be doing, unless the user has explicitly already set something.

simonbyrne:
From https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryUsageGuidelines.html
it states that the default value is `$HOME/lib;/usr/local/lib;/usr/lib`

simonbyrne:
Not sure why they're using semicolons: it clearly states they should be colons, and `man dyld` says it is `$(HOME)/lib:/usr/local/lib:/lib:/usr/lib`

simonbyrne:
```
ERROR: LoadError: KeyError: key "DYLD_FALLBACK_LIBRARY_PATH" not found
```

src/AutoBuild.jl
```diff
@@ -1079,7 +1083,7 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
                         end
                     end
                     if adjust_LIBPATH
-                        if !isempty(get(ENV, LIBPATH_env, ""))
+                        if !isempty(get(ENV, LIBPATH_env, expanduser(LIBPATH_default)))
                             env_mapping[LIBPATH_env] = string(LIBPATH, $(repr(pathsep)), ENV[LIBPATH_env])
```
simonbyrne, line 1087:
```suggestion
                            env_mapping[LIBPATH_env] = string(LIBPATH, $(repr(pathsep)), get(ENV, LIBPATH_env, expanduser(LIBPATH_default)))
```

simonbyrne: @staticfloat if you update your branch with this, I can try it again.

giordano: You can try now

simonbyrne:
Success!
https://travis-ci.com/github/simonbyrne/mpichtest/jobs/328408776#L631-L632

src/AutoBuild.jl
simonbyrne: Success!
https://travis-ci.com/github/simonbyrne/mpichtest/jobs/328408776#L631-L632

simonbyrne:
Can we retrigger a build of MPICH with this?

giordano:
There we go: https://github.com/JuliaPackaging/Yggdrasil/pull/1022

PR #776, opened 2020-05-05
Do not write platforms to JSON if building for AnyPlatform

giordano:
The platform will be brought back when cleaning up the merged objects.

Also, we need to special-case `platform_key_abi("any")` in `build_tarballs()`.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -161,7 +161,8 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
     # If the user passed in a platform (or a few, comma-separated) on the
     # command-line, use that instead of our default platforms
     if length(ARGS) > 0
-        platforms = platform_key_abi.(split(ARGS[1], ","))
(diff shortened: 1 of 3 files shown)


src/AutoBuild.jl
```diff
@@ -161,7 +161,8 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
     # If the user passed in a platform (or a few, comma-separated) on the
     # command-line, use that instead of our default platforms
     if length(ARGS) > 0
-        platforms = platform_key_abi.(split(ARGS[1], ","))
+        _platform_key_abi(p::AbstractString) = p == "any" ? AnyPlatform() : platform_key_abi(p)
+        platforms = _platform_key_abi.(split(ARGS[1], ","))
```
staticfloat, line 165: Ah, this is necessary because of how we build on Yggdrasil.  yeah, okay.

giordano: :heavy_check_mark:

PR #783, opened 2020-05-09
Move building of documentation to GitHub Actions

giordano:
Also decrease frequency of TagBot to once per day


giordano:
The advantage of separating documentation from the rest of CI is that for documentation-only PRs we can use `[skip travis]` (or any other of [these keywords](https://docs.travis-ci.com/user/customizing-the-build/#skipping-a-build)) to skip CI and build only the documentation, without having to wait for all other jobs to finish

PR #786, opened 2020-05-15
Make sure the artifacts of dependencies are downloaded, even if lazy

### LINKED ISSUE #785: Force installation of artifacts of dependencies, including lazy ones
The artifacts of the dependencies are installed with https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/8fa38e9fe146d72cf6b2e61530ef1c75b5215fc2/src/Prefix.jl#L417 which supposes that the artifacts will be downloaded at installation time.  This assumption however fails when having a lazy artifact, like `MKL_jll`.  We need to ensure that the artifact is downloaded, even if it's lazy.

giordano:
Fix #785

### DIFF
--- a/src/Prefix.jl
+++ b/src/Prefix.jl
@@ -451,6 +451,9 @@ function setup_dependencies(prefix::Prefix, dependencies::Vector{PkgSpec}, platf
                 continue
             end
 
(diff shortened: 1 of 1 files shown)


giordano:
I'm not entirely sure what happens when the artifact is not available for the given platform :thinking:

Issue #687, opened 2020-02-23
Help relocatability of libraries/executables in the JLL wrappers


giordano:
We have several cases where we need to set some environment variables at runtime (or using `withenv`) to relocate the libraries/executables:

* for `git` we need to set `GIT_EXEC_PATH` and `GIT_SSL_CAINFO`, see https://github.com/JuliaPackaging/Yggdrasil/issues/284 and https://github.com/JuliaPackaging/Yggdrasil/issues/452
* for `ncurses` we need to set `TERMINFO` or `TERMINFINO_DIRS`, see https://github.com/JuliaPackaging/Yggdrasil/issues/455
* for `Tk`/`Tcl` we need to set `TCL_LIBRARY` and `TK_LIBRARY`, see https://github.com/JuliaPackaging/Yggdrasil/issues/462
* for `OpenMPI` we need to set `OPAL_PREFIX`, see https://github.com/JuliaPackaging/Yggdrasil/issues/390
* for `Gtk` we need to set a bunch of environment variables, see the [`__init__` function]( https://github.com/JuliaGraphics/Gtk.jl/blob/08d802bc8e0ce0c1925f70570539d417f2e910ba/src/Gtk.jl#L72-L134) of `Gtk.jl`
* for `Fontconfig` we need to set `FONTCONFIG_FILE`, see the `__init__` functions of [`Fontconfig.jl`](https://github.com/JuliaGraphics/Fontconfig.jl/blob/130ca0cc855ec42c49cb456df53c2031e1285dec/src/Fontconfig.jl#L16-L26) and [`Cairo.jl`](https://github.com/JuliaGraphics/Cairo.jl/blob/c4029ce288b6c2c8b92cd1dacb14b1e631f9be06/src/Cairo.jl#L20-L24)

We usually delegate the Julia packages calling the libraries to set these variables, but this has a couple of drawbacks:

* all packages calling the JLL package need to do it, see the example of `Cairo.jl` and `Fontconfig.jl`, both using `Fontconfig_jll`
* in some cases, the Julia packages doesn't need to know what JLL package it's using.  This is for example the case of `MPI.jl` reported by @simonbyrne in https://github.com/JuliaPackaging/Yggdrasil/issues/390

Maybe we should allow setting these variables directly in the JLL wrappers, to alleviate the burden of the developers of the high-level interface of the package.

[...]

Issue #647, opened 2020-01-30
Improving JLL READMEs with example usage


staticfloat:
It would be good to provide easy ways for users to showcase how to use JLLs.  Right now we spit out a pretty bland example usage with fake products, but this can be improved upon if authors so desire.  I'm thinking there should be a per-`Product` optional field that takes in a usage example.  I'm thinking perhaps what we can do here is to give the opportunity to provide something like a little "doctest" for JLLs; not only can we use this as a way to provide documentation, but we can also run it to ensure that the JLL itself is working.

So now the question is, what is the API?  Perhaps something as simple as the following:

```julia
products = Product[
    FileProduct("src/data.txt", :data_txt; usage="""
		str = read(open(data_txt, "r"))
	"""),
    LibraryProduct("libdataproc", :libdataproc, usage="""
        data_lines = "this is a test\nline two of my test!"
		ccall((libdataproc, :count_characters), Cint, (Cstring, Cint), data_lines, length(data_lines))
	"""),
    ExecutableProduct("mungify", :mungify_exe, usage="""
		mungify_exe() do exe_path
			run(`$exe_path --input=$(data_txt) --find-all=jlbuildbot`)
		end
	"""),
]
```

These would get integrated into the README.md generation, and they would _also_ get run when building a platform that is compatible with the host platform, to ensure that the products work at least as well as the author is claiming they should.  We wouldn't check that output matches or anything like that like a normal doctest (although we would do a `using Test` beforehand, in case someone _wanted_ to do that) but the idea here is just to give some guidelines to what someone can do with this package.

laborg:
In the library ccall example the function and the library appear to mixed up: I think it should be `ccall((:count_characters, libdataproc)...`

PR #791, opened 2020-05-19
Add optional init_block kwarg to build_tarball

fingolfin:
This optional argument can be set to a string, which is then inserted at
the end of the generated __init__ methods. This allows inserting code
which initializes the JLL or in other ways makes it easier to use; e.g
for ncurses, it could set the TERMINFO env variable; for GMP, it could call
the `mp_set_memory_functions` function to ensure this copy of GMP uses
the same alloc/free functions as the GMP bundled with Julia.

This is motivated by issue #687 and would e.g. allow us to get rid of https://github.com/oscar-system/LoadFlint.jl (by changing https://github.com/JuliaBinaryWrappers/FLINT_jll.jl to contain the appropriate initialization calls).

Full disclaimer: the code in here is 100% untested! So it may even contain basic syntax errors. The reasons for this is that while I am willing to spend time making it work etc., I also don't want to waste time figuring everything out from scratch if I then might be told later on that the whole approach is meritless and won't ever have a chance of being merged :-)

Hence, I thought I'd first solicit some basic feedback. E.g.
- is the idea of inserting code into `__init__` in principle OK? Perhaps the code should be inserted elsewhere (many variations come to mind; e.g. it could be a single function defined in the main file, which then is invoked by `__init__`; etc.)
- is the same `init_block` OK? Other suggestions?
- it OK that I am simply inserting a string? The rationale is that this is dead easy to do. In issue #687, @simonbyrne suggested that one could also pass a `quote` block -- I can also look into this, but it seems to be more complex with no obvious (to me) gain. But I'll do whatever the consensus is.
- I wanted to add tests, but I found no existing tests for `build_jll_package` or `build_tarballs` so I don't know how I should go about that...

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -76,6 +76,17 @@ download, build and package the tarballs, generating a `build.jl` file when
 appropriate.  Note that `ARGS` should be the top-level Julia `ARGS` command-
 line arguments object.  This function does some rudimentary parsing of the
 `ARGS`, call it with `--help` in the `ARGS` to see what it can do.
+
+The `kwargs` are passed on to [`autobuild`](@ref), see there for a list of
+supported ones. In addition, the keyword argument `init_block` may be set to
+a string containing Julia code; if present, this code will be inserted into
+the initialization path of the generated JLL package. This can for example be
+used to invoke an initialization API of a shared library.
+
+!!! note
+
+    The `init_block` keyword argument is experimental and may be removed
+    in a future version of this package. Please use it sparingly.
 """
 function build_tarballs(ARGS, src_name, src_version, sources, script,
                         platforms, products, dependencies; kwargs...)
@@ -227,7 +238,7 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
         bin_path = "https://github.com/$(deploy_jll_repo)/releases/download/$(tag)"
         build_jll_package(src_name, build_version, sources, code_dir, build_output_meta,
                           dependencies, bin_path; verbose=verbose,
-                          extract_kwargs(kwargs, (:lazy_artifacts,))...,
+                          extract_kwargs(kwargs, (:lazy_artifacts, :init_block))...,
                           )
         push_jll_package(src_name, build_version; code_dir=code_dir, deploy_repo=deploy_repo)
         if register
@@ -567,6 +578,7 @@ function autobuild(dir::AbstractString,
                    code_dir::Union{String,Nothing} = nothing,
                    require_license::Bool = true,
                    lazy_artifacts::Bool = false,
+                   init_block::String = "",
                    meta_json_stream = nothing,
                    kwargs...)
     @nospecialize
@@ -580,6 +592,7 @@ function autobuild(dir::AbstractString,
             "products" => products,
             "dependencies" => dependencies,
             "lazy_artifacts" => lazy_artifacts,
+            "init_block" => init_block,
         )
         # Do not write the list of platforms when building only for `AnyPlatform`
         if platforms != [AnyPlatform()]
@@ -888,6 +901,8 @@ function init_jll_package(name, code_dir, deploy_repo;
     end
 end
 
+# rebuild_jll_package is not called from anywhere in BinaryBuilder,
+# but rather from JuliaPackaging/Yggdrasil/.ci/register_package.jl
 function rebuild_jll_package(obj::Dict;
                              download_dir = nothing,
                              upload_prefix = nothing,
@@ -920,6 +935,7 @@ function rebuild_jll_package(obj::Dict;
         upload_prefix;
         verbose=verbose,
         lazy_artifacts = lazy_artifacts,
+        init_block = get(obj, "init_block", ""),
         from_scratch = from_scratch,
     )
 end
@@ -929,7 +945,7 @@ function rebuild_jll_package(name::String, build_version::VersionNumber, sources
                              download_dir::String, upload_prefix::String;
                              code_dir::String = joinpath(Pkg.devdir(), "$(name)_jll"),
                              verbose::Bool = false, lazy_artifacts::Bool = false,
-                             from_scratch::Bool = true)
+                             init_block::String = "", from_scratch::Bool = true)
     # We're going to recreate "build_output_meta"
     build_output_meta = Dict()
 
@@ -982,7 +998,7 @@ function rebuild_jll_package(name::String, build_version::VersionNumber, sources
             )
         end
 
-        # If `from_scartch` is set (the default) we clear out any old crusty code
+        # If `from_scratch` is set (the default) we clear out any old crusty code
         # before generating our new, pristine, JLL package within it.  :)
         if from_scratch
             rm(joinpath(code_dir, "src"); recursive=true, force=true)
@@ -990,7 +1006,9 @@ function rebuild_jll_package(name::String, build_version::VersionNumber, sources
         end
 
         # Finally, generate the full JLL package
-        build_jll_package(name, build_version, sources, code_dir, build_output_meta, dependencies, upload_prefix; verbose=verbose, lazy_artifacts=lazy_artifacts)
+        build_jll_package(name, build_version, sources, code_dir, build_output_meta,
+                          dependencies, upload_prefix; verbose=verbose,
+                          lazy_artifacts=lazy_artifacts, init_block=init_block)
     end
 end
 
@@ -1002,7 +1020,8 @@ function build_jll_package(src_name::String,
                            dependencies::Vector,
                            bin_path::String;
                            verbose::Bool = false,
-                           lazy_artifacts::Bool = false)
+                           lazy_artifacts::Bool = false,
+                           init_block = "")
     if !Base.isidentifier(src_name)
         error("Package name \"$(src_name)\" is not a valid identifier")
     end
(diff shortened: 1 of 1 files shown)


fingolfin:
@giordano If you have any thoughts on this, I'd appreciate your input, if your time permits.

staticfloat:
I was hoping to avoid this. 😞

I've been hoping against hope that binary packages can be mostly loaded with a very cookie-cutter approach, so that we can identify how most binary packages can be loaded, then do things like eliminate JLL packages completely and have that binary loading logic stored within Julia base itself.  The less complexity and the more regularity we have in code loading, the faster and more reliable it will be.

Allowing arbitrary code to be packed into the `__init__()` method of your JLL is only going to increase the surface area of things that can go wrong and cause you problems.  However, you raise great points that not allowing this makes life very difficult in certain strange situations.  Therefore, I'm open to adding this as an escape hatch, but I think we should be very careful in using it, and we should do our best to add more structured methods that make dealing with classes of these problems easier.  Examples:

* For the environment variable setting issues in https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/687, we could have users provide products that should get mapped to environment variables that get set automatically upon loading a JLL.  Something that looks like:

```julia
build_tarballs(...; jll_environment_vars = Dict(
    # Mapping to a string sets a constant value across all platforms
    "CONSTANT_VAR" => "This environment variable is always set to the same thing",
    # Mapping to a product sets the result of `locate()`'ing the product
    "LIBFOO_PATH" => LibraryProduct("libfoo", :libfoo),
))
```

This is probably a good idea, and I think could be implemented now.

* For the need to call initialization functions.... I'm not sure what a more structured API would look like.  Perhaps this can be integrated into our ideas for a [simple JLL test suite](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/647) as well; we're already thinking about how to embed pieces of code into the `build_tarballs.jl`.  I'm not sure.

> I wanted to add tests, but I found no existing tests for `build_jll_package` or `build_tarballs` so I don't know how I should go about that...

.....yeah, that's a long-standing dev wishlist item.  ;)

fingolfin:
> this makes life very difficult in certain strange situations

I don't think those are strange situations; depending on what you do, they are common!

E.g. Gmp_jll does not init GMP to use Julia's allocation functions. Hence one must be very careful to either initialize it very, very early (but that may be impossible if multiple packages using it are involved, and not all are under your control), or else generally ensure that the shared library in there is not actually used (and that Julia's bundled GMP is used instead), otherwise one can run into nasty situations if one ever wants to use a GMP integer allocated in one copy of GMP inside another copy -- segfaults or worse are otherwise likely.

Anyway, I totally understand your wish, but I am concerned that trying to find the perfect solution here will delay implementing *any* solution indefinitely. I understand this feature can be easily abused, and use should be minimized, but *some* escape hatch should be there.


So, how can we move forward? Sorry, the env var setting (which indeed sounds useful) simply isn't enough for my usecases.


Note that I did think about alternatives. e.g. for GMP and Flint, one could also apply a patch to the source code which changes the default memory allocation functions to use those of Julia. The problem with that is that it requires linking to Julia; of course now we have Julia_JLL; but during runtime, we *really* want to link against the active Julia, not another copy. That may be possible (I did not try this), but my point here is that this is rather non-trivial.

For the env variables, in some cases it might also be possible (and even better) to patch source code to insert different default paths that work under relocation; I did such things for a different kind of "reloctable binary distribution" I created years ago (nothing fancy, see https://github.com/fingolfin/gap-osx-bundle). But that particular one was much easier than what you are doing here, in that it only targeted macOS, and everything was in an `.app` bundle, so I could leverage OS support to make this easier; getting this kind of patch to work across Linux, macOS, Windows, FreeBSD is highly non-trivial, and one must be wary of accidentally breaking the patched code; still, perhaps it could be used in some cases, esp. if BinaryBuilder provides some help. I only have some fuzzy ideas on what that "help" might look like; really, the best way is as you describe it: attempt to do it, then watch out for patterns that you can factor out.

I mean, you already do some of that by tweaking the shared libraries to be relocatable; but e.g. for shell scripts, this might also be useful.

BTW, some JLLs contain hardcoded paths derived from the build system. This is usually a source of non-relocatability. In https://github.com/fingolfin/gap-osx-bundle I thus added extra checks which grep through the produced binaries for the builddir (which had a very unique name to avoid accidental hits), and produced warnings/errors in that case. I then tried hard to get rid all of these (to be "warning free", and also to not leak info about the build system, which is an important step towards [reproducible builds](https://reproducible-builds.org), something I hope JLLs will slowly converge towards).

fingolfin:
@staticfloat it would be really good to make progress on this. Right now, though I don't know what you think needs to be done before this can be merged? I understand that you'd prefer a "better" solution to the initialization problem, but I don't know what that's supposed to look like, so I can't work on it either. 

I agree that a way to just set env vars would be desirable, and it also seems quite doable (see also https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/687#issuecomment-635867414). But that doesn't solve the issue of having to call some initialization APIs...

Perhaps @giordano has some insights?

staticfloat:
@fingolfin Alright, your persistence has paid off. :)

Rebase this, and I'll merge the `init_block` as a string being inserted into the end of `__init__()`.  I would like you to document it as an experimental feature that can disappear in the future.  I would like to work toward having better APIs that don't require arbitrary code execution (or if they do, to have it in a slightly more user-friendly manner, e.g. a bundled `.jl` file that gets included from within `__init__()` perhaps?) but that shouldn't block you from getting real work done.

fingolfin:
I've rebased this now. Marking this as experimental seems sensible; I'll be happy to adjust if / once "better" approaches are found. 

In the meantime, being able to watch how some packages use `init_block` might be helpful in figuring out alternative approaches.

fingolfin:
Ahh, before this is merged, I should point out that I still did not actually test this thoroughly, partially because I am not sure *how* to test it.

src/AutoBuild.jl
```diff
@@ -920,6 +930,7 @@ function rebuild_jll_package(obj::Dict;
         upload_prefix;
         verbose=verbose,
         lazy_artifacts = lazy_artifacts,
+        init_block = get(obj, "init_block", ""),
```
fingolfin, line 933: Is this the correct way to do it?

src/AutoBuild.jl
```diff
@@ -982,15 +993,15 @@ function rebuild_jll_package(name::String, build_version::VersionNumber, sources
...
+        # If `from_scratch` is set (the default) we clear out any old crusty code
         # before generating our new, pristine, JLL package within it.  :)
         if from_scratch
             rm(joinpath(code_dir, "src"); recursive=true, force=true)
             rm(joinpath(code_dir, "Artifacts.toml"); force=true)
         end
 
         # Finally, generate the full JLL package
-        build_jll_package(name, build_version, sources, code_dir, build_output_meta, dependencies, upload_prefix; verbose=verbose, lazy_artifacts=lazy_artifacts)
+        build_jll_package(name, build_version, sources, code_dir, build_output_meta, dependencies, upload_prefix; verbose=verbose, lazy_artifacts=lazy_artifacts, init_block=init_block)
```
fingolfin, line 1004: Perhaps this line should be wrapped? Are you using `JuliaFormatter` or something like that?

fingolfin:
@staticfloat  build passes, but I marked one spot where I am not 100% sure. Also, should I added comments indicating "this is experimental"?

src/AutoBuild.jl
staticfloat: No, but I usually try to wrap things at about 90 characters.  So yes please, feel free to wrap it.

src/AutoBuild.jl
staticfloat: We can assume that the only `obj` objects coming in are ones generated by this same BB version; so we will always have an `init_block` key in `obj`.

src/AutoBuild.jl
```diff
@@ -76,6 +76,12 @@ download, build and package the tarballs, generating a `build.jl` file when
 appropriate.  Note that `ARGS` should be the top-level Julia `ARGS` command-
 line arguments object.  This function does some rudimentary parsing of the
 `ARGS`, call it with `--help` in the `ARGS` to see what it can do.
+
+The `kwargs` are passed on to [`autobuild`](@ref), see there for a list of
+supported ones. In addition, the keyword argument `init_block` may be set to
+a string containing Julia code; if present, this code will be inserted into
+the initialization path of the generated JLL package. This can for example be
+used to invoke an initialization API of a shared library.
```
staticfloat, line 84: This is where you should mention that `init_block` is experimental and may be removed in a future release.

src/AutoBuild.jl
fingolfin: Done. (Lots of lines > 90 in the file, but ... *shrug* ;-)

src/AutoBuild.jl
fingolfin: Done via `!!! note`, let me know if you prefer something else.

staticfloat:
I manually tested this by making this change to the [`HelloWorldC` build script](https://github.com/JuliaPackaging/Yggdrasil/blob/master/H/HelloWorldC/build_tarballs.jl):

```diff
-build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies;
+               init_block="""
+               ENV["JULIA_HELLO_WORLD"]="C"
+               """,
+               )
```

Then, I built it with:
```
julia --color=yes build_tarballs.jl --verbose --deploy=staticfloat/HelloWorldC_jll.jl
```

Which resulted in wrappers that have the relevant piece of init code: https://github.com/staticfloat/HelloWorldC_jll.jl/blob/f7788fa180b68067edc305755df8d20617bb0a9f/src/wrappers/x86_64-apple-darwin14.jl#L58

Looks good to me, merging!

PR #793, opened 2020-05-22
Fix isolated dlopen

### LINKED ISSUE #194: build.jl files should include dependencies
Now that building dependent packages is working pretty well (see, for instance, [GlibBuilder](https://github.com/staticfloat/GlibBuilder)) we should make sure that the generated `build.jl` files chain up with the other `build.jl` files automatically, so that if the user downloads the auto-generated `build.jl` file, it will automatically fetch and install all necessary dependencies.

fingolfin:
In `locate`, when launching an external Julia to test whether dlopen work
on a library, pass `--startup-file=no` to the subprocess to avoid any issues
with commands in a user's startup.jl file.

While I don't know a specific bug this causes, in the corresponding code in
BinaryProvider it triggered a [very annoying and difficult to debug issue](https://github.com/JuliaPackaging/BinaryProvider.jl/pull/195), so
better be safe than sorry?

### DIFF
--- a/src/Products.jl
+++ b/src/Products.jl
@@ -195,7 +195,7 @@ function locate(lp::LibraryProduct, prefix::Prefix; platform::Platform = platfor
                     if (!lp.dont_dlopen && !skip_dlopen) && platforms_match(platform, platform_key_abi())
                         if isolate
                             # Isolated dlopen is a lot slower, but safer
-                            if success(`$(Base.julia_cmd()) -e "import Libdl; Libdl.dlopen(\"$dl_path\")"`)
+                            if success(`$(Base.julia_cmd()) --startup-file=no -e "import Libdl; Libdl.dlopen(\"$dl_path\")"`)
                                 return dl_path
                             end
                         else


giordano:
I don't know whether this fixes the issue you're experiencing since I can't reproduce it, but not using the startup file is definitely a good thing to do anyway!

giordano:
Hold on, I just realised this is BinaryBuilder!  I assume you wanted to fix this in BinaryProvider?

fingolfin:
LOL that's both hilarious and embarrassing: I've never ever before accidentally submitted a PR to the "wrong" repository. But it turns out both BinaryProvider and BinaryBuilder have files `src/Products.jl` with similar content which perhaps had a common root but has diverged? So both have a `locate`, and both have `dlopen` calls in there with isolation, and both have this bug... except that BinaryProvider actually fixed it *on master* but not on rollback.

So I still think this PR should be applied here, but of course with a different description. I'll also open a suitable PR on BinaryProvider.

Thank you!

staticfloat:
Yes, this is a good, subtle change!  Thanks @fingolfin!

PR #794, opened 2020-05-22
Reorganise README and LICENSE of the JLL package

### LINKED ISSUE #780: Improvements to jll readme
* It might be useful if the README for jll packages showed the sources from which they are built.  So maybe enumerate all the files in a `FileSource`, and display the URL and SHA for git and archive sources. 

* It will also be useful if the README can display the current latest version of the package. 

* Now that there is BB documentation on how to use jll packages, maybe the _Usage_ section (with Examples_jll) might be made into a link. That will reduce a lot of duplication (and IMO resulting confusion)

### LINKED ISSUE #792: Two license files per JLL?
I am not sure whether this is the right place for this issue, but...

It seems that JLLs are created with TWO license files; `LICENSE` and `LICENSE.md`; see e.g. <https://github.com/JuliaBinaryWrappers/lib4ti2_jll.jl>. This is somewhat confusing.

The `LICENSE` file is added as part of the "initial commit" which has only this file and nothing else. And `LICENSE.md` is added by BinaryBuilder.

giordano:
* remove "Usage example" section from the README and improve JLL packages
section in BinaryBuilder documentation (ref #780)
* add version of the package to the README (#780)
* add list of sources to the README (#780)
* add list of dependencies to the README
* add a link to Yggdrasil repository in the README when the builder comes from
there (#780)
* add the list of platforms in the README (ref #780)
* do not generate a duplicate license file (fixes #792)
* explicitly point out in the README that the license does not apply to the
wrapped binary package (ref #635)

You can see a preview of what the README and the license would look like at https://github.com/giordano/libsodium_jll.jl.  I believe this fully fixes #780.


src/AutoBuild.jl
```diff
@@ -1294,52 +1309,37 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
-
-        The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
-
-        THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+        The code of this Julia library is released under the terms of the MIT
+        "Expat" License.  This license does not apply to the binary package
+        wrapped by the present library and whose license is shipped in the
+        accompanying tarball.
```
giordano, line 1322: Terrible wording, but  I'm sure someone will suggest good improvements :slightly_smiling_face:

src/AutoBuild.jl
```diff
@@ -1268,6 +1278,10 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
         print(io, jll_jl)
     end
 
+    print_source(io, s::ArchiveSource) = println(io, "* compressed archive ", s.url, " (SHA256 checksum: `", s.hash,"`)")
+    print_source(io, s::GitSource) =     println(io, "* git repository ", s.url, " (revision: `", s.hash,"`)")
+    print_source(io, s::FileSource) =    println(io, "* file ", s.url, " (SHA256 checksum: `", s.hash,"`)")
+    print_source(io, ::DirectorySource) = nothing
```
giordano, line 1284: I was not sure whether we want to list also the list of local directories, as we don't have a direct reference from here.  I'm open to suggestions

src/AutoBuild.jl
```diff
@@ -1276,15 +1290,36 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
+            print_source.(Ref(io), sources)
+            println(io)
+        else
+            print(io, " ")
+        end
+        println(io, "for the following platforms:")
+        println(io)
+        for p in platforms
+            println(io, "* `", p, "` (`", triplet(p), "`)")
+        end
```
giordano, line 1318: I'm not sure I like the format of the README of https://github.com/giordano/libsodium_jll.jl: the sources and the platforms are verbosely listed in a bullet list, the products are in a new section, printed as Julia code.

It's probably worth to add the dependencies as well

src/AutoBuild.jl
```diff
@@ -1268,79 +1278,78 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
 
-        The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+                  ## License
 
-        THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-        """)
+                  The code of this Julia library is released under the terms of the MIT "Expat" License.  This license does not apply to the binary package wrapped by the present library and whose license is shipped in the accompanying tarball.
```
giordano, line 1351: The text with the terrible wording is now here.  Suggestions about how to improve it are welcome :slightly_smiling_face:

src/AutoBuild.jl
giordano: I've moved each piece of information to its own section, I think it's easier to parse in this way.  I've added the dependencies as well and the products are in a bullet list

src/AutoBuild.jl
aviks: It would be nice if we list the files inside local directories, but I don't know of that's possible.

giordano: Not really, not now at least: we don't have the list of files in the directory anywhere.

staticfloat: Let's list the name of the directory, and if we're building from Yggdrasil, we can auto-detect the gitsha of the current commit, then emit a link to the directory at that commit.

src/AutoBuild.jl
```diff
@@ -894,7 +894,8 @@ function rebuild_jll_package(obj::Dict;
                              gh_org::String = "JuliaBinaryWrappers",
                              verbose::Bool = false,
                              lazy_artifacts::Bool = false,
-                             from_scratch::Bool = true)
+                             from_scratch::Bool = true,
+                             from_yggdrasil::Bool = false)
```
staticfloat, line 898: Is it too magical to find if there's an enclosing `git` repository, check its remote URL, and auto-set `from_yggdrasil`?  ;). It's fine to say "yes", this works fine since we fully control how this method is called.

src/AutoBuild.jl
```diff
@@ -1268,79 +1278,78 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
-              ```
+              # `$(src_name)_jll.jl` (`v$(build_version)`)
 
-              With such products defined, `Example_jll` would contain `data_txt`, `libdataproc` and `mungify_exe` symbols exported. For `FileProduct` variables, the exported value is a string pointing to the location of the file on-disk.  For `LibraryProduct` variables, it is a string corresponding to the `SONAME` of the desired library (it will have already been `dlopen()`'ed, so typical `ccall()` usage applies), and for `ExecutableProduct` variables, the exported value is a function that can be called to set appropriate environment variables.  Example:
-
-              ```julia
-              using Example_jll
+              This is an autogenerated package constructed using [`BinaryBuilder.jl`](https://github.com/JuliaPackaging/BinaryBuilder.jl).""")
+        if from_yggdrasil
+            print(io, " The recipe can be found on [`Yggdrasil`](https://github.com/JuliaPackaging/Yggdrasil/), the community build tree.")
```
staticfloat, line 1303: Since we know the `name`, we can make an educated guess and link directly to the project folder.

src/AutoBuild.jl
staticfloat, line 1351:
```suggestion
                  The Julia source code within this repository (all files under `src/`) are released under the terms of the MIT "Expat" License, the text of which is included below.  This license does not apply to the binary package wrapped by this Julia package and automatically downloaded by the Julia package manager upon installing this wrapper package.  The binary package's license is shipped alongside the binary itself and can be found within the `share/licenses/$name` directory within its prefix.
```

src/AutoBuild.jl
giordano: In most cases it will work, but there are some notable exception, like [`Coin-OR`](https://github.com/JuliaPackaging/Yggdrasil/tree/4a49d243c6a211eb18de52abde1ac3fdfc6a48b0/C/Coin-OR), [`OpenBLAS`](https://github.com/JuliaPackaging/Yggdrasil/tree/4a49d243c6a211eb18de52abde1ac3fdfc6a48b0/O/OpenBLAS) and [`LLVM`](https://github.com/JuliaPackaging/Yggdrasil/tree/4a49d243c6a211eb18de52abde1ac3fdfc6a48b0/L/LLVM).  [`R/Librsvg`](https://github.com/JuliaPackaging/Yggdrasil/tree/4a49d243c6a211eb18de52abde1ac3fdfc6a48b0/R/Librsvg) is also an interesting case where the letter directory doesn't match the first letter of the builder name.  These are the reason why I didn't add a direct link to the directory.  If we're ok with having a few broken special cases, I can add the link.

staticfloat: I think all except `Librsvg` are actually fine, it's still more helpful than the overall Yggdrasil link.  For `Librsvg`, we may just move that to `L` sometime.  :P

src/AutoBuild.jl
giordano: > the text of which is included below

This is actually in the `README`.  I assume you're suggesting to move it to the `LICENSE` file, doing so.

giordano:
Unless I missed something, I should have addressed all comments from review.  A preview of repo is always at https://github.com/giordano/libsodium_jll.jl.  Note in particular the [`LICENSE`](https://github.com/giordano/libsodium_jll.jl/blob/master/LICENSE) and the [`README.md`](https://github.com/giordano/libsodium_jll.jl/blob/master/README.md).

src/AutoBuild.jl
fingolfin: Perhaps then a CI check should also be introduced which enforces that the builder name matches the directory name, to prevent future drift?

I guess `OpenBLAS` is indeed fine, but I don't understand why `Coin-OR` is?

src/AutoBuild.jl
```diff
@@ -1268,80 +1275,98 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
-              $(join(collect("    $(typeof(p))($(product_names(p)), :$(variable_name(p)))" for (p, _) in products_info), ",\n"))
-              ]
-              ```
-
-              """,
-              """
-              ## Usage example
-
-              For example purposes, we will assume that the following products were defined in the imaginary package `Example_jll`:
+              # `$(src_name)_jll.jl` (`v$(build_version)`)
```
fingolfin, line 1306: Why is the version in code quotes? The result looks a bit funny, IMHO. How about this as an alternative? (I realize this is a bit of bikeshedding; feel free to ignore)
```suggestion
              # `$(src_name)_jll.jl` $(build_version)
```

src/AutoBuild.jl
```diff
@@ -1268,80 +1275,98 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
-    product_names(p::LibraryProduct) = p.libnames
-    product_names(p::FrameworkProduct) = p.libraryproduct.libnames
+    is_yggdrasil = get(ENV, "YGGDRASIL", "false") == "true"
+    ygg_head = is_yggdrasil ? LibGit2.head("..") : ""
+    src_name_first = uppercase(first(src_name))
+    print_source(io, s::ArchiveSource) = println(io, "* compressed archive: ", s.url, " (SHA256 checksum: `", s.hash,"`)")
+    print_source(io, s::GitSource) =     println(io, "* git repository: ", s.url, " (revision: `", s.hash,"`)")
+    print_source(io, s::FileSource) =    println(io, "* remote file: ", s.url, " (SHA256 checksum: `", s.hash,"`)")
+    function print_source(io, s::DirectorySource)
+        print(io, "* files in local directory: ")
```
fingolfin, line 1285: This is IMHO confusing: if I look at https://github.com/giordano/libsodium_jll.jl and read it there, I'd wonder "but where is this `./bundled` directory, I don't see it?".

Of course it's a bit better if this is from yggdrasil, but even then, "local directory" just doesn't make sense in my head (which might just indicate my head is too soft ;-) ). But how about this
```suggestion
        print(io, "* files bundled with the script used to generate this JLL: ")
```

giordano: > "but where is this ./bundled directory, I don't see it?".

The link is broken because the example is totally fake

src/AutoBuild.jl
```diff
@@ -1268,80 +1275,98 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
-
-        THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-        """)
+    # Add before the license a note about to what files this applies
+    license = strip(read(joinpath(code_dir, "LICENSE"), String))
+    note = "The Julia source code within this repository (all files under `src/`) are released under the terms of the MIT \"Expat\" License, the text of which is included below.  This license does not apply to the binary package wrapped by this Julia package and automatically downloaded by the Julia package manager upon installing this wrapper package.  The binary package's license is shipped alongside the binary itself and can be found within the `share/licenses/$(src_name)` directory within its prefix."
```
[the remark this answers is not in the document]
giordano: Note that this will completely break the check below

src/AutoBuild.jl
```diff
@@ -1268,80 +1275,98 @@ function build_jll_package(src_name::String, build_version::VersionNumber, code_
...
+    # Since this function can be called multiple times, we must make sure that
+    # the note is written only once.  Do nothing it is already there.
```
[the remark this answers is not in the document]
giordano: > That seems really brittle.

Why?  It's just how our CI system on Yggdrasil works, which can call the `build_jll_package` function multiple times, there is nothing fundamental.

> Where does the original license come from?

https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/b2cd2e35cb34cea0983c3b5ca88220d98b2d8940/src/AutoBuild.jl#L854

giordano:
> it would be kinda great if all JLLs could be regenerated to have this

It would be even greater if we had a mechanism to do this, but we don't

src/AutoBuild.jl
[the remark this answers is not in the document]
giordano: I wrapped this text

src/AutoBuild.jl
giordano: I tried to clarify this, but with a slightly less verbose wording

src/AutoBuild.jl
giordano: Done

src/AutoBuild.jl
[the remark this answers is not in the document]
fingolfin: With "brittle" I meant that if the content of `note_lines` is changed (e.g. to reformat it ;-) ) then this code would prepend it once again, instead of replacing the old variant.

So, why not just specify the full content of the LICENSE file here, and write it out unconditionally? What is gained by prepending this text? One thing I can think of is that in principle this would allow somebody to manually edit the JLL repository, to replace the license. But is that even supported? Who even would have such access?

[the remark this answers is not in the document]
giordano: One point is that the line with the copyright owner is custom, as you noted, I didn't want to change that one

fingolfin:
> > it would be kinda great if all JLLs could be regenerated to have this
> 
> It would be even greater if we had a mechanism to do this, but we don't

Well if there is even interest in this (I had zero feedback on that issue so far), I's be happy to discuss potential approaches to handling this and work on implementing one, too. I guess. It's also a question about how often this would be necessary -- one kinda would hope the "format" of JLLs would eventually stabilize; but this isn't 100% clear. And updating all those JLLs would also involve version changes etc... so it's be perhaps best to wait and first apply other changes that would affect many JLLs (e.g. it might be a good idea to add a postprocessing step to BinaryBuilder which remove all
Libtool `PREFIX/lib/*.la` files; and one might think about taking steps towards reproducible builds; and more...

giordano:
I already did it yesterday: https://github.com/JuliaPackaging/Yggdrasil/pull/1091.  Need still to test it though

giordano:
> I guess. It's also a question about how often this would be necessary -- one kinda would hope the "format" of JLLs would eventually stabilize; but this isn't 100% clear. And updating all those JLLs would also involve version changes etc...

Yes, I'm not overly happy about do mass updates for _all_ packages every time we change something, because the users will get a new version of several dozens of JLL packages every now and then.

src/AutoBuild.jl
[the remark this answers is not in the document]
fingolfin: But for all JLLs on `JuliaBinaryWrappers` this is just set to `JuliaBinaryWrappers`... That doesn't seem very useful...

BTW, it also always sets the current year as copyright year, which also is not quite right. Ideally it'd be `<year the jll was created> - <current_year>`; folded together into a single line if this is the same year. Of course that might be annoying to implement (I guess one could query git for the creation date of the build file?), so I am not suggestion you waste your time on this -- just thought I should mention it for completeness. :-)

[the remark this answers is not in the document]
giordano: We don't currently have any of the information you're suggesting to use

> Ideally it'd be <year the jll was created> - <current_year>; folded together into a single line if this is the same year. Of course that might be annoying to implement (I guess one could query git for the creation date of the build file?)

We must assume that a file must checked in a git repository, which isn't a requirement anywhere.

I think you may want to add more metadata to the builder.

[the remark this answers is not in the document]
fingolfin: Yeah, I realize that this information is not available (which is why I wrote "Ideally it'd be", but I should have made that clearer). And in fact I do think there should be more metadata in the builder.

Anyway, I just thought about it some more, and I now think my "ideally..." suggestion was actually totally *wrong*: this license after all applies to the generated wrapper. And that wrapper is generated by BinaryBuilder. So there is a case to be made that its copyright belongs to the authors of BinaryBuilder resp. the AutoBuild code; the person who run build_tarballs just pressed a button, but the actual creation of the JLL content is not done by them ...

So I think I'd either remove that copyright line, or use something like `Copyright 2019-2020 The authors of the BinaryBuilder.jl package".

Anyway, this is of course not very important, it just irks me a little to know that I might soon see `Copyright (c) 2020 JuliaBinaryWrappers` in lots of LICENSE files -- JuliaBinaryWrappers isn't a legal entity and can't hold any copyrights :-)

src/AutoBuild.jl
giordano: Fixed in #807

PR #795, opened 2020-05-23
Clarify wizard prompt asking for file path

### LINKED ISSUE #784: [Wizard] prompt for filename unclear
The prompt  *Enter path to build_tarballs.jl to write to* is unclear. I thought it wanted the directory, but it needed the filename. 
```
Enter path to build_tarballs.jl to write to:
> /home/user/Yggdrasil/L/librsync

SystemError: opening file "/home/user/Yggdrasil/L/librsync": Is a directory
```

giordano:
Fix #784.  @aviks is this any better? :grimacing:

### DIFF
--- a/src/wizard/deploy.jl
+++ b/src/wizard/deploy.jl
@@ -246,7 +246,7 @@ function _deploy(state::WizardState)
             ins=state.ins,
             outs=state.outs,
         )
-        mkpath(directory))
+        mkpath(directory)
         open(joinpath(directory, "build_tarballs.jl"), "w") do io
             print_build_tarballs(io, state)
         end
--- a/src/wizard/deploy.jl
+++ b/src/wizard/deploy.jl
@@ -240,14 +240,14 @@ function _deploy(state::WizardState)
         )
         yggdrasil_deploy(state, yggdrasil_select == 1)
     elseif deploy_select == 2
-        filename = nonempty_line_prompt(
+        directory = nonempty_line_prompt(
             "filename",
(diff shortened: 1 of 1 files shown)


aviks:
probably not? 😞

giordano:
Any concrete suggestion?

giordano:
Maybe in the true wizard spirit we can check if the path is an existing string and if so create a file called `build_tarballs.jl` in there, otherwise assume it's the path to the file?

aviks:
So yeah, thinking about it, I think the problem here is that the computer knows that the filename it needs is `build_tarballs.jl`, yet aks the user to enter it. Computers should ask the user to enter only information that it does not already know.

So by this logic, the user should only be asked to enter the directory.

If you want the user to enter the filename, say something like 'Enter the full path to a file to save the build recipe. This file is traditionally called build_tarball.jl'

giordano:
@aviks ok, changed to prompt for the path of the directory

src/wizard/deploy.jl
```diff
@@ -240,14 +240,14 @@ function _deploy(state::WizardState)
...
+        directory = nonempty_line_prompt(
             "filename",
-            "Enter path to build_tarballs.jl to write to:";
+            "Enter directory where to write build_tarballs.jl to:";
             ins=state.ins,
             outs=state.outs,
         )
-        mkpath(dirname(abspath(filename)))
-        open(filename, "w") do io
+        mkpath(directory))
```
staticfloat, line 249:
```suggestion
        mkpath(directory)
```

PR #809, opened 2020-06-03
Split BinaryBuilder into multiple modules

giordano:
This is the beginning of a long journey.  For the time being, this is a proof-of-concept more than else.  Tests will fail very soon.

My long term plan is to restructure BinaryBuider to make its codebase more well-structured and easier to grasp for potential new contributors.  We can also probably take some functionalities completely out of this package, like `OutputCollector` and everything related to writing the JLL wrappers and registering the packages.  Elliot suggested that we could even move the sandboxing feature to its own package.

An advantage of this reform is that this package could basically do only what it supposed to do: build the binaries.  Moving some features outside of this package would make it easier to test them, as the test suite of this package is already largish.  The JLL wrappers are basically untested, having a package dedicated to generating them would be a huge improvement in this regard.


giordano:
I believe the Travis cache is a bit too aggressive.  Tests are failing with
```
ERROR: LoadError: UndefVarError: check_cxxstring_abi not defined
Stacktrace:
 [1] _precompile_() at /home/travis/build/JuliaPackaging/BinaryBuilder.jl/deps/SnoopCompile/precompile/precompile_BinaryBuilder.jl:104
 [2] top-level scope at /home/travis/build/JuliaPackaging/BinaryBuilder.jl/src/BinaryBuilder.jl:364
 [3] include at ./boot.jl:328 [inlined]
 [4] include_relative(::Module, ::String) at ./loading.jl:1105
 [5] include(::Module, ::String) at ./Base.jl:31
 [6] top-level scope at none:2
 [7] eval at ./boot.jl:330 [inlined]
 [8] eval(::Expr) at ./client.jl:425
 [9] top-level scope at ./none:3
in expression starting at /home/travis/build/JuliaPackaging/BinaryBuilder.jl/src/BinaryBuilder.jl:364
```
but line 104 of `deps/SnoopCompile/precompile/precompile_BinaryBuilder.jl` https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/e6560ed7edd13cc3f9df8865b5038706c6a721a5/deps/SnoopCompile/precompile/precompile_BinaryBuilder.jl#L104
doesn't have any reference to `check_cxxstring_abi`.

giordano:
Yay, it's all green now, including snoop precompilation!

My plan to go forward is to:

* create a new package called `OutputCollectors.jl`: https://github.com/giordano/BinaryBuilder.jl/blob/6a517cf9402509e3fa100d014aad96fee9190cc4/src/BinaryBuilder.jl#L16-L22
* create a new package called `BinaryBuilderBase.jl` or `Sandbox.jl` or any other meaningful name: https://github.com/giordano/BinaryBuilder.jl/blob/6a517cf9402509e3fa100d014aad96fee9190cc4/src/BinaryBuilder.jl#L24-L153
* create a new package called `JLLTools.jl`: https://github.com/giordano/BinaryBuilder.jl/blob/6a517cf9402509e3fa100d014aad96fee9190cc4/src/AutoBuild.jl#L317-L1412

src/AutoBuild.jl
```diff
@@ -314,7 +314,7 @@ function get_compilers_versions(; compilers = [:c])
     return output
 end
 
-function upload_to_github_releases(repo, tag, path; gh_auth=github_auth(;allow_anonymous=false), 
+function upload_to_github_releases(repo, tag, path; gh_auth=Wizard.github_auth(;allow_anonymous=false),
```
giordano, line 317: Importing stuff from `Wizard` isn't exactly something I wanted to do, but for the time being it does the job

src/wizard/obtain_source.jl
```diff
@@ -422,14 +421,14 @@ function step2(state::WizardState)
...
-        get_preferred_version(state, "LLVM", getversion.(reverse(available_llvm_builds)))
+        get_preferred_version(state, "LLVM", getversion.(reverse(BinaryBuilderBase.available_llvm_builds)))
     else
         state.compilers = [:c]
         # Default GCC version is the oldest one
-        state.preferred_gcc_version = getversion(available_gcc_builds[1])
+        state.preferred_gcc_version = getversion(BinaryBuilderBase.available_gcc_builds[1])
         # Default LLVM version is the latest one
-        state.preferred_llvm_version = getversion(available_llvm_builds[end])
+        state.preferred_llvm_version = getversion(BinaryBuilderBase.available_llvm_builds[end])
```
staticfloat, line 432: Can you just `using BinaryBuilderBase: available_gcc_builds, available_llvm_builds` so that you don't have to litter the module name all over the place?

test/auditing.jl
```diff
@@ -35,37 +90,81 @@ end
...
+                    install_license /usr/include/ltdl.h
+                    """,
+                    # Build for our platform
+                    [platform],
+                    # Ensure our executable products are built
+                    [product],
+                    # No dependencies
+                    Dependency[];
+                )
+            end
```
staticfloat, line 150: Can you add a test where `lock_microarchitecture == true` but you set a `-march` flag that causes the compiler wrapper to complain?

test/build_tests/isa_tests/main.c
```diff
@@ -6,9 +6,10 @@
 
 ELTYPE summation(ELTYPE * data, int length) {
     ELTYPE sum = 0;
-    for( int idx=0; idx<length/STRIDE; ++idx ) {
-        for( int jdx=0; jdx<STRIDE; ++jdx ) {
-            sum += data[idx + jdx];
+    int idx, jdx;
+    for( idx=0; idx<length/STRIDE; ++idx ) {
+        for( jdx=0; jdx<STRIDE; ++jdx ) {
+            sum += data[idx * STRIDE + jdx];
```
staticfloat, line 12: Not a fan of C99? ;)

src/wizard/obtain_source.jl
giordano: Yep, done :+1:

test/auditing.jl
giordano: This is a test I wanted to have in `BinaryBuilderBase`, I'll add it later, thanks for reminding me!

test/build_tests/isa_tests/main.c
giordano: Not me in particular, but a compiler I used at some point these days wasn't happy about the C99-style of the loop :smile:

PR #832, opened 2020-06-25
Use script to generate list of instructions

giordano:
This is based on #809.  Opening this as a draft just to keep the work done so far


contrib/generate_instructions_list.jl
```diff
@@ -0,0 +1,40 @@
...
+                    unique!(sort!(push!(all_instructions, instruction)))
+                    unique!(sort!(push!(instructions, instruction)))
+                end
+            end
+        end
+        dict[name] = instructions
+    end
+    free(xml)
+    # We're basically converting an XML to a JSOn, funny isn't it?
+    write(joinpath(@__DIR__, "..", "src", "auditor", "instructions.json"), JSON.json(dict))
```
staticfloat, line 38: Can we pretty-print the dict?  I know it's more bytes on disk, but it should help a lot, especially with diffs in the future.

src/auditor/instruction_set.jl
```diff
@@ -1,1105 +1,8 @@
...
-        "vp4dpwssds",
-        "vpopcntd",
-        "vpopcntq",
-    ],
+using JSON
 
-)
+## We start with definitions of instruction mnemonics, broken down by category:
+const instruction_categories = JSON.parsefile(joinpath(@__DIR__, "instructions.json");
+                                              dicttype=Dict{String,Vector{String}})
```
staticfloat, line 5: Nice, parsing happens at precompile time so no runtime overhead.  👍

src/auditor/instruction_set.jl
```diff
@@ -1177,18 +80,19 @@ whether the object file being pointed to is a 32-bit or 64-bit one:
 """
 function minimum_instruction_set(counts::Dict, is_64bit::Bool)
     if is_64bit
-        if counts["avx512"] > 0
+        if counts["PKU"] + counts["RDSEED"] + counts["ADCX"] + counts["CLFLUSH"] + counts["XSAVEC"] + counts["XSAVES"] + counts["CLWB"] + counts["AVX512EVEX"] + counts["AVX512VEX"] > 0
```
staticfloat, line 83:
```suggestion
        avx512_instruction_categories = ("PKU", "RDSEED", "ADCX", "CLFLUSH", "XSAVEC", "XSAVES", "CLWB", "AVX512EVEX", "AVEX512VEX")
        if any(get.(Ref(counts), avx512_instruction_categories, 0) .> 0)
```

staticfloat: I think this form would be a little clearer; and this way I don't have to wonder if the counts can be negative or whatever.

PR #838, opened 2020-06-30
Add basic tests for JLL wrappers

giordano:


test/jll.jl
```diff
@@ -0,0 +1,126 @@
...
+    @test_throws MethodError build_project_dict("LibFoo", v"1.3.5", [Dependency("Zlib_jll"), BuildDependency("Xorg_util_macros_jll")])
+end
+
+@testset "JLLs - building" begin
+    mktempdir() do build_path
+        name = "libfoo"
+        version = v"1.0.0"
+        sources = [DirectorySource(build_tests_dir)]
+        # Build for this platform and a platform that isn't this one for sure:
+        # FreeBSD.
```
giordano, line 29: @ararslan :heart:

test/jll.jl
```diff
@@ -0,0 +1,126 @@
...
+            "JULIA_PKG_DEVDIR" => joinpath(build_path, "devdir"),
+            # Let's pretend to be in Yggdrasil, set the relevant environment
+            # variables.
+            "YGGDRASIL" => "true",
+            "BUILD_SOURCEVERSION" => "0123456789abcdef0123456789abcdef01234567",
+            "PROJECT" => "L/$(name)",
+        ) do
+            # What follows loosely mimics what we have to do to build JLL packages
+            # in Yggdrasil.
+            buff = IOBuffer(strip(String(take!(buff))))
```
giordano, line 78: I'm pretty sure that this is overly complicated, but I didn't quickly understand why `buff` as it is caused errors in `JSON.parse` below

giordano: It looks like the `strip` is mandatory, without it (or simply with `seekstart(buff)` instead of this line) I get
```julia
JLLs - building: Error During Test at /home/mose/.julia/dev/BinaryBuilder/test/jll.jl:23
  Unexpected end of input
   ...when parsing byte with value '10'
  Stacktrace:
   [1] error(::String) at ./error.jl:33
   [2] _error(::String, ::JSON.Parser.StreamingParserState{Base.GenericIOBuffer{Array{UInt8,1}}}) at /home/mose/.julia/packages/JSON/d89fA/src/Parser.jl:150
   [3] byteat(::JSON.Parser.StreamingParserState{Base.GenericIOBuffer{Array{UInt8,1}}}) at /home/mose/.julia/packages/JSON/d89fA/src/Parser.jl:59
   [4] current at /home/mose/.julia/packages/JSON/d89fA/src/Parser.jl:72 [inlined]
   [5] chomp_space! at /home/mose/.julia/packages/JSON/d89fA/src/Parser.jl:117 [inlined]
   [6] parse_value(::JSON.Parser.ParserContext{Dict{String,Any},Int64,true,nothing}, ::JSON.Parser.StreamingParserState{Base.GenericIOBuffer{Array{UInt8,1}}}) at /home/mose/.julia/packages/JSON/d89fA/src/Parser.jl:160
   [7] parse(::Base.GenericIOBuffer{Array{UInt8,1}}; dicttype::Type{T} where T, inttype::Type{Int64}, allownan::Bool, null::Nothing) at /home/mose/.julia/packages/JSON/d89fA/src/Parser.jl:494
   [8] parse(::Base.GenericIOBuffer{Array{UInt8,1}}) at /home/mose/.julia/packages/JSON/d89fA/src/Parser.jl:492
```

staticfloat: If you want, you can replace `IOBuffer()` with `BufferStream()` from `SimpleBufferStream` and you probably don't have to do any of this.  Just use `buff` without a `seekstart()` at all.

test/jll.jl
```diff
@@ -0,0 +1,126 @@
...
+            end
+
+            tag = "$(name)-v$(build_version)"
+            upload_prefix = "https://github.com/$(repo)/releases/download/$(tag)"
+
+            # This loop over the unmerged objects necessary in the event that we have multiple packages being built by a single build_tarballs.jl
+            for (i,json_obj) in enumerate(objs_unmerged)
+                from_scratch = (i == 1)
+                BinaryBuilder.rebuild_jll_package(json_obj; download_dir=download_dir, upload_prefix=upload_prefix, verbose=false, lazy_artifacts=json_obj["lazy_artifacts"], from_scratch=from_scratch)
+            end
```
staticfloat, line 116: Can we drop this kind of stuff, since we know exactly what we're building?

test/jll.jl
giordano: For the record, it turned out `SimpleBufferStream` doesn't help here, it's probably a limitation of `JSON.jl`.  `JSON3.jl` might be a better option, but it also adds some complexity (we'd need to create a struct to wrap the entire data we want to write to disk) that might not be worth the extra work now, since `JSON.jl` isn't perfect but does the hob.

staticfloat:
Huzzah!

PR #841, opened 2020-07-07
Add to the JLLs a function to tell whether the current platform is supported

giordano:
Happy to receive better suggestions for the name of the function.

TODO:

* [ ] Mention this function in the documentation of the JLL wrappers

CC: @fingolfin

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1121,6 +1121,9 @@ function build_jll_package(src_name::String,
             end
 
             print(io, """
+            # Inform that the wrapper is available for this platform
+            is_available() = tue
+
             \"\"\"
             Open all libraries
             \"\"\"
@@ -1214,6 +1217,13 @@ function build_jll_package(src_name::String,
 
         using Pkg, Pkg.BinaryPlatforms, Pkg.Artifacts, Libdl
         import Base: UUID
+
+        \"\"\"
+            is_available()
+
+        Return whether the artifact is available for the current platform.
+        \"\"\"
+        is_available() = false
 
         # We put these inter-JLL-package API values here so that they are always defined, even if there
         # is no underlying wrapper held within this JLL package.
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1122,7 +1122,7 @@ function build_jll_package(src_name::String,
 
             print(io, """
             # Inform that the wrapper is available for this platform
-            is_available() = tue
+            is_available() = true
 
             \"\"\"
             Open all libraries
--- a/test/jll.jl
+++ b/test/jll.jl
@@ -134,6 +134,7 @@ end
             Pkg.develop(PackageSpec(path=code_dir))
             @eval TestJLL using libfoo_jll
             @test 6.08 ≈ @eval TestJLL ccall((:foo, libfoo), Cdouble, (Cdouble, Cdouble), 2.3, 4.5)
+            @test @eval TestJLL libfoo_jll.is_available()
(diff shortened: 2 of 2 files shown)


src/AutoBuild.jl
```diff
@@ -1121,6 +1121,9 @@ function build_jll_package(src_name::String,
             end
 
             print(io, """
+            # Inform that the wrapper is available for this platform
+            is_available() = tue
```
fingolfin, line 1125:
```suggestion
            is_available() = true
```

fingolfin: Isn't the usual Julia convention to not have an underscore after `is`, i.e., `isavailable`? Mind you, I am not complaining; personally, I prefer `is_...` but it doesn't seem to be used in Julia otherwise?

giordano: That's a convention in the Julia's code base, here we have a lot of functions with underscores.

Any other suggestion about the name? Is `is_available` clear enough? Alternatives?

src/AutoBuild.jl
```diff
@@ -1215,6 +1218,13 @@ function build_jll_package(src_name::String,
         using Pkg, Pkg.BinaryPlatforms, Pkg.Artifacts, Libdl
         import Base: UUID
 
+        \"\"\"
+            is_available()
+
+        Return whether the artifact is available for the current platform.
```
fingolfin, line 1224: Short version: I really think `is_available` is fine, thank you!

Long version: I personally find `is_available` clear enough. If I am asked to be super critical, I could point out that "the JLL is of course available, else Pkg would have refused to install it". But of course in your doc string, you make it clear that this about whether the *artifact* making up the bulk of any JLL is available (for this platform) or not. So once one reads that, it should be clear to any Julia *developer* what this is about.

For a *user*, I just hope that developers turn this into helpful error messages (if they choose to display an error at all, perhaps just some functionality get disabled). I.e., I hope they'll user the doc string as inspiration to produce checks like this:
```julia
if !some_JLL.is_available
  error("some_JLL is not available for the current platform")
end
```
and not something IMHO confusing like this:
```julia
if !some_JLL.is_available
  error("some_JLL is not available") # huh? but Pkg loaded it just fine...
end
```
I tried but I can't really think of a name that conveys that in a clearer fashion without being confusing in other ways or too bulky, e.g.
- `is_usable` ("that's a subjective question, isn't it? I don't really like your API")
- `is_available_on_current_platform` ("this ain't Java")
- `is_ready` (makes it sound as if there was a way to make it ready)
- `is_supported` ("so, do you provide support via email, too?")

giordano: > If I am asked to be super critical

You are, that's why I pinged you :wink:

> I could point out that "the JLL is of course available, else Pkg would have refused to install it".

Yes, that was my concern as well.  However, as you guessed, this function is supposed to be used mainly by the developers who are going to use the JLL package.  I'll also document this in https://juliapackaging.github.io/BinaryBuilder.jl/dev/jll/, to hopefully further reduce confusion.

> I tried but I can't really think of a name that conveys that in a clearer fashion without being confusing in other ways or too bulky, e.g.
> - `is_usable` ("that's a subjective question, isn't it? I don't really like your API")
> - `is_available_on_current_platform` ("this ain't Java")
> - `is_ready` (makes it sound as if there was a way to make it ready)
> - `is_supported` ("so, do you provide support via email, too?")

Love them :slightly_smiling_face:

fingolfin:
Any particular reason this is not yet merged?

PR #853, opened 2020-07-25
Fix documentation typo

BeastyBlacksmith:


staticfloat:
Believe it or not, this is actually intended.  :)

PR #855, opened 2020-07-26
Pass credentials to Libgit2.fetch in init_jll_package

mortenpi:
I am playing around with BinaryBuilder and private repos. Specifically, trying to update an existing _private_ jll repo. The `git fetch` step here asks for username & password, even though I am using `ENV["GITHUB_TOKEN"]`.

Passing the credentials to `fetch` (like is done for `clone` just a few lines before) would make sense to me. It seems to fix it for my use-case, but I haven't tested it beyond that.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1382,19 +1382,13 @@ function push_jll_package(name, build_version;
     wrapper_repo = LibGit2.GitRepo(code_dir)
     LibGit2.add!(wrapper_repo, ".")
     LibGit2.commit(wrapper_repo, "$(name)_jll build $(build_version)")
-    creds = LibGit2.UserPasswordCredential(
-        deepcopy(gh_username),
-        deepcopy(gh_auth.token),
-    )
-    try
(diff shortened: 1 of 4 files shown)


giordano:
The change looks good to me, thank you so much!  However, it's probably better to consolidate in a function, since this pattern is used 3 times in this file, as far as I can see?  Maybe as a function that takes a function as first argument, so that we can use the `do`-block, the the body of `try` can be different?

mortenpi:
> However, it's probably better to consolidate in a function

Agreed, I just didn't want the first iteration to touch too much code. I've created a function for it now.

I wasn't quite sure where to put it though. I also noticed that the same pattern was used in `Wizard`. Since `Wizard` has the `utils.jl` file, I figure it would fit there pretty well, but you need to call it as `Wizard.with_gitcreds` in the main `BinaryBuilder` module. Happy to move it around though if you think it would fit better somewhere else.

giordano:
Awesome, thanks also for all the tests!

Issue #521, opened 2019-11-27
local build of `foo_jll` package


ssfrr:
I'd like to locally generate the `foo_jll` package for testing but I can't figure out if that's possible. It looks like generating the package is pretty intertwined with actually pushing it to GitHub. I tried `--deploy=ssfrr/foo_jll` but got several `404` errors. Then when I tried again I got a `422` "validation failed" error with the "already exists" code.

edit: I got around the `422` error by going into my GitHub settings and manually deleting the "BinaryBuilder.jl generated token" access token.

fingolfin:
PR #861 takes care of this, essentially.

staticfloat:
You should now (on the latest `master` of `BinaryBuilder.jl`) be able to specify `--deploy=local`, and it will build a JLL package in your `~/.julia/dev` directory, which you can immediately use with the artifacts constructed during your actual build.

If you are developing a C library locally and want to skip BB itself completely, you can deploy binaries to a `override` directory within the dev'ed out JLL package, and it will load those instead of the artifact libraries (you can double-check by looking at `Libdl.dllist()` after loading your JLL package).  Because it may not be immediately obvious what the layout of the `override` directory should be, you can use `dev_jll()` within your JLL package to dev it out, then copy the contents of the artifact in as an example layout.  Then you can replace the library at will, and immediately test changes locally as quickly as you can compile.

PR #861, opened 2020-07-28
Improve local development workflow

staticfloat:
This adds two new capabilities:

* `--deploy=local`; for those who really just want to test out their new
JLLs locally, this makes it a cinch.  Just tack `--deploy=local` onto
the end of your `build_tarballs.jl` invocation, and you'll get the
artifacts installed and the JLL packages built within your
`~/.julia/dev` directory. (Note that you will still need to add that
directory into any environment you want to use it from)

* `override` directories for `dev`'ed out JLL packages.  This allows a
user to override the binaries within a JLL easily and cheaply.  At
compile time, the JLL package will look for an `override` directory in
its project root, and if it finds it, it will attempt to load binaries
from there instead of from an Artifact.  To aid in setting this up, JLLs
now have a `dev_jll()` method that will perform the checkout and copy
over the normal artifact's contents to the `override` directory,
allowing for the user to easily see what must be overwritten in order to
test out a local build of something.

### DIFF
--- a/docs/src/building.md
+++ b/docs/src/building.md
@@ -348,36 +348,41 @@ Examples of builders that depend on other binaries include:
 # Building and testing JLL packages locally
 
 As a package developer, you may want to test JLL packages locally, or as a binary dependency
-developer you may want to easily use custom binaries.  In both cases, the new `dev` and
-`overrides` directory support make it easy to get complete control over the local JLL package
-state.
+developer you may want to easily use custom binaries.  Through a combination of `dev`'ing out
+the JLL package and creating an `overrides` directory, it is easy to get complete control over
+the local JLL package state.
 
 ## Overriding a prebuilt JLL package's binaries
 
 After running `pkg> dev LibFoo_jll`, a local JLL package will be checked out to your depot's
-`dev` directory, (on most installations this is `~/.julia/dev`) and by default it will make
-use of binaries within your depot's `artifacts` directory.  However, if an `override`
-directory is present within that package directory, it will look within that directory for
-binaries, rather than in any artifact directory.  There is no mixing and matching of binaries
-within a single JLL package; if an `override` directory is present, all products defined within
-that JLL package must be found within the `override` directory, none will be sourced from an
-artifact.  Dependencies (e.g. found within another JLL package) may still be loaded from their
-respective artifacts, so dependency JLLs must themselves be `dev`'ed and have `override`
-directories created with files or symlinks created within them.
+`dev` directory (on most installations this is `~/.julia/dev`) and by default the JLL package
+will make use of binaries within your depot's `artifacts` directory.  If an `override`
+directory is present within the JLL package directory, the JLL package will look within that
+`override` directory for binaries, rather than in any artifact directory.  Note that there is
+no mixing and matching of binaries within a single JLL package; if an `override` directory is
+present, all products defined within that JLL package must be found within the `override`
+directory, none will be sourced from an artifact.  Dependencies (e.g. found within another
+JLL package) may still be loaded from their respective artifacts, so dependency JLLs must
+themselves be `dev`'ed and have `override` directories created with files or symlinks
+created within them.
 
-### Auto-populating `override` directory
+### Auto-populating the `override` directory
 
-To ease creation of an `override` directory, JLL packages now contain a `dev_jll()` function,
+To ease creation of an `override` directory, JLL packages contain a `dev_jll()` function,
 that will ensure that a `~/.julia/dev/<jll name>` package is `dev`'ed out, and will copy the
 normal artifact contents into the appropriate `override` directory.  This will result in no
 functional difference from simply using the artifact directory, but provides a template of
 files that can be replaced by custom-built binaries.
 
+Note that this feature is rolling out to new JLL packages as they are rebuilt; if a JLL
+package does not have a `dev_jll()` function, [open an issue on Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil/issues/new)
+and a new JLL version will be generated to provide the function.
+
 ## Building a custom JLL package locally
 
 When building a new version of a JLL package, if `--deploy` is passed to `build_tarballs.jl`
-then a JLL package will be deployed to a GitHub repository.  (Read the documentation given
-by passing `--help` to a `build_tarballs.jl` script for more on `--deploy` options)  If
-`--deploy=local` is passed, the JLL package will still be built in the `~/.julia/dev/`
-directory, but it will not be uploaded anywhere.  This is useful for local testing and
-validation that the built artifacts are working with your package.
+then a newly-built JLL package will be deployed to a GitHub repository.  (Read the
+documentation given by passing `--help` to a `build_tarballs.jl` script for more on
+`--deploy` options).  If `--deploy=local` is passed, the JLL package will still be built
+in the `~/.julia/dev/` directory, but it will not be uploaded anywhere.  This is useful
+for local testing and validation that the built artifacts are working with your package.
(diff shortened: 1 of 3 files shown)


src/AutoBuild.jl
```diff
@@ -1219,18 +1228,55 @@ function build_jll_package(src_name::String,
...
         import Base: UUID
 
+        wrapper_available = false
         \"\"\"
             is_available()
 
         Return whether the artifact is available for the current platform.
         \"\"\"
-        is_available() = false
+        is_available() = wrapper_available
```
giordano, line 1237: Maybe call the function `wrapper_available`?  I wasn't 100% happy with `is_available`

staticfloat: I don't care about the name; I just care about the compiler warnings about precompilation being broken when we redefine a top-level method.  :P

If you want a different name, you can pick whatever you want. :)

docs/src/building.md
```diff
@@ -344,3 +344,40 @@ Examples of builders that depend on other binaries include:
 * [`Xorg_libX11`](https://github.com/JuliaPackaging/Yggdrasil/blob/eb3728a2303c98519338fe0be370ef299b807e19/X/Xorg_libX11/build_tarballs.jl#L36-L42)
   depends on `Xorg_libxcb_jll`, and `Xorg_xtrans_jll` at build- and run-time,
   and on `Xorg_xorgproto_jll` and `Xorg_util_macros_jll` only at build-time.
+
+# Building and testing JLL packages locally
+
+As a package developer, you may want to test JLL packages locally, or as a binary dependency
+developer you may want to easily use custom binaries.  In both cases, the new `dev` and
+`overrides` directory support make it easy to get complete control over the local JLL package
+state.
```
fingolfin, line 353: The way this is phrased (esp. the use of "the new ... support") makes it well suited for a release announcement, but IMHO somewhat less so for a permanent entry in the manual. I don't claim to know the perfect way to do this, but in an attempt to be constructive and not just destructive, here is an attempt for an alternative phrasing
```suggestion
developer you may want to easily use custom binaries. To support these use cases, we provide
means to make it easy to get complete control over the local JLL package state, which are
described in the following subsections.
```

docs/src/building.md
```diff
@@ -344,3 +344,40 @@ Examples of builders that depend on other binaries include:
...
+
+## Overriding a prebuilt JLL package's binaries
+
+After running `pkg> dev LibFoo_jll`, a local JLL package will be checked out to your depot's
+`dev` directory, (on most installations this is `~/.julia/dev`) and by default it will make
```
fingolfin, line 358: "it" = the JLL?
```suggestion
`dev` directory (on most installations this is `~/.julia/dev`), and by default the JLL will make
```

docs/src/building.md
```diff
@@ -344,3 +344,40 @@ Examples of builders that depend on other binaries include:
...
+use of binaries within your depot's `artifacts` directory.  However, if an `override`
+directory is present within that package directory, it will look within that directory for
```
fingolfin, line 360: Again, "it".

Also, when reading this I was unsure what "that package directory" refers to -- presumably the one in `~/.julia/dev`? I didn't see mention of another, but "that" sounds (to me, as a non-native speaker, so this may very well be entirely be my own fault!) as if there was some alternate dir to consider? (And perhaps there is and I am just oblivious to it). Perhaps this could be made clearer; I am not quite sure how so; but for me, the following already is a bit clearer.

```suggestion
directory is present within its package directory, the JLL will look within that directory for
```

docs/src/building.md
```diff
@@ -344,3 +344,40 @@ Examples of builders that depend on other binaries include:
...
+binaries, rather than in any artifact directory.  There is no mixing and matching of binaries
+within a single JLL package; if an `override` directory is present, all products defined within
+that JLL package must be found within the `override` directory, none will be sourced from an
+artifact.  Dependencies (e.g. found within another JLL package) may still be loaded from their
+respective artifacts, so dependency JLLs must themselves be `dev`'ed and have `override`
+directories created with files or symlinks created within them.
+
+### Auto-populating `override` directory
+
+To ease creation of an `override` directory, JLL packages now contain a `dev_jll()` function,
```
fingolfin, line 370: I am a bit torn of the use of "now" here, for the reason I already outlined above: how will this read to somebody in one year? One way to go about this wold be to insert a date in here, e.g. "since August 2020". But that doesn't seem great, also it depends a bit on when the patch is merged and Yggdrasil updated to use it? 

So I guess instead I would suggest this: drop "now" here, and instead insert a comment somewhere along the lines of: "Note that older JLLs which were generated before this feature was introduced may lack a `dev_jll`
function." 
```suggestion
To ease creation of an `override` directory, JLL packages contain a `dev_jll()` function,
```

docs/src/building.md
```diff
@@ -344,3 +344,40 @@ Examples of builders that depend on other binaries include:
...
+that will ensure that a `~/.julia/dev/<jll name>` package is `dev`'ed out, and will copy the
+normal artifact contents into the appropriate `override` directory.  This will result in no
+functional difference from simply using the artifact directory, but provides a template of
+files that can be replaced by custom-built binaries.
```
fingolfin, line 374:
```suggestion
files that can be replaced by custom-built binaries.

Note that JLLs which were generated before this feature was introduced to
BinaryBuilder may lack a `dev_jll` function.
```

docs/src/building.md
```diff
@@ -344,3 +344,40 @@ Examples of builders that depend on other binaries include:
...
+
+## Building a custom JLL package locally
+
+When building a new version of a JLL package, if `--deploy` is passed to `build_tarballs.jl`
+then a JLL package will be deployed to a GitHub repository.  (Read the documentation given
```
fingolfin, line 379:
```suggestion
then the newly-built JLL package will be deployed to a GitHub repository.  (Read the documentation given
```

docs/src/building.md
```diff
@@ -344,3 +344,40 @@ Examples of builders that depend on other binaries include:
...
+by passing `--help` to a `build_tarballs.jl` script for more on `--deploy` options)  If
```
fingolfin, line 380:
```suggestion
by passing `--help` to a `build_tarballs.jl` script for more on `--deploy` options).  If
```

docs/src/building.md
```diff
@@ -344,3 +344,40 @@ Examples of builders that depend on other binaries include:
...
+`--deploy=local` is passed, the JLL package will still be built in the `~/.julia/dev/`
+directory, but it will not be uploaded anywhere.  This is useful for local testing and
+validation that the built artifacts are working with your package.
```
fingolfin, line 383:
```suggestion
for validating that the built artifacts are working with your package.
```

docs/src/jll.md
```diff
@@ -209,3 +209,15 @@ defined for it are:
...
   is kept for consistency with all other product types.
+
+
+## Overriding `dev`'ed JLL packages
+
+In the event that a user wishes to override the binaries within a JLL package with
+their own binaries, the user may use the `dev_jll()` method inside of JLL packages
+to check out a mutable copy of the package to their `~/.julia/dev` directory.  An
+`override` directory will be created within the package directory, providing a
+convenient location for the user to copy in their own libraries over the typical
```
fingolfin, line 220:
```suggestion
convenient location for the user to copy in their own files over the typical
```

I'd also consider changing "binaries" at the start of this paragraph to "files", as in principle one might also want to tweak, say, a header or config file, right? Wrong?

docs/src/jll.md
```diff
@@ -209,3 +209,15 @@ defined for it are:
 
 * `data_txt_path`: this unexported variable is actually equal to `data_txt`, but
   is kept for consistency with all other product types.
+
+
+## Overriding `dev`'ed JLL packages
+
+In the event that a user wishes to override the binaries within a JLL package with
+their own binaries, the user may use the `dev_jll()` method inside of JLL packages
```
fingolfin, line 217:
```suggestion
In the event that a user wishes to override the binaries within a JLL package with
their own binaries, the user may use the `dev_jll()` method provided by that JLL package
```

fingolfin:
I think this resolves #521

PR #867, opened 2020-07-31
Include build_tarballs --help text in the manual

mortenpi:
Moves the `build_tarballs` docstring a bit down in the manual and puts the `--help` text right after it. It will render as a code block, so unfortunately won't be searchable, but still better than not having it at all I think.

### DIFF
--- a/docs/src/FAQ.md
+++ b/docs/src/FAQ.md
@@ -32,7 +32,7 @@ Remember also that you should use the standard environment variables like `CC`, 
 
 ### I love the wizard, but now I want to break free: can I build the tarballs without it?
 
-The `build_tarballs.jl` script can be used as a command line utility, it takes a few options and as argument the list of triplets of the targets.  You can find more information about the syntax of the script with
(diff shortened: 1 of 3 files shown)


giordano:
This is hinted in this section: https://juliapackaging.github.io/BinaryBuilder.jl/dev/FAQ/#I-love-the-wizard,-but-now-I-want-to-break-free:-can-I-build-the-tarballs-without-it? It would be great to have a reference to the new one from there :slightly_smiling_face: 

PS: a tip: to avoid running the tests for BB, which take forever and frequently fail, for documentation-only PRs, add `[skip ci]` or `[skip travis]` in the commit message :wink:

mortenpi:
I added in some cross-references and also slightly edited the `build_tarballs` docstring. Good to go on my part.

> PS: a tip: to avoid running the tests for BB, which take forever and frequently fail, for documentation-only PRs, add `[skip ci]` or `[skip travis]` in the commit message

Whoops, sorry for the CI noise. Will keep that in mind in the for the future :slightly_smiling_face:

giordano:
Perfect, thanks!

Issue #856, opened 2020-07-27
Add a way to adjust the compat entry for Julia in JLL's `Package.toml`


fingolfin:
See also the discussion in issue #511.

In addition to the use case described there, this would also be useful for e.g. GMP.jl (see https://github.com/JuliaRegistries/General/pull/16517) and MPFR.jl (see @giordano's commit https://github.com/JuliaRegistries/General/pull/18342/commits/0495de0dc77381e6d0ccb791a0b28cfe09c26de6 in https://github.com/JuliaRegistries/General/pull/18342/commits).

giordano:
For whoever wants to work on this, the function to look at is [`build_project_dict`](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/2bfca64a84bda4fa5d7535542d86c395e5bcfaf3/src/AutoBuild.jl#L1420-L1460).

fingolfin:
I guess it would make sense to pass the Julia compat to `build_project_dict` via its existing `dependencies` argument.

Hmmm... how about adding a generic `RuntimeDependency` type, kinda dual to `BuildDependency`: it is only used to generate the `Project.toml` of the JLL but has no effect during build time. Then one could use this for Julia compat, but also for other things. Not that I have any good example in mind, so this may be pointless. However, it seems to be very easy to implement. 

One alternative is to add a new optional keyword argument to `build_tarball` which then has to be threaded through the code until it finally reaches `build_project_dict `. See also https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/511#issuecomment-634974371. That requires modifying quite a few places, and one also has to be careful to make sure it reaches `rebuild_jll_package`; all of this is "for free" if it was just another type of dependency.

giordano:
While I agree `RuntimeDependency` can be useful in some situations (none that we've seen so far though), note that `julia` must be treated specially anyway: it won't go into the `deps` section of `Project.toml`, but only in `compat`, and even if not specified explicitly it must have a compat entry, supporting v1.0 by default.  So I don't think having `julia` as `RuntimeDependency` would be less work.

giordano:
> While I agree `RuntimeDependency` can be useful in some situations (none that we've seen so far though)

There we go: https://github.com/JuliaPackaging/Yggdrasil/pull/2418#issuecomment-763536147

PR #872, opened 2020-08-04
Move part of autobuild() into new get_meta_json() helper

fingolfin:
The two new functions have similar but not identical signature; this change
means that we don't have to pass things to autobuild() which are not needed
for building, but rather just for the meta JSON; and vice versa.

I came up with this while trying to write a patch for issue #856, and trying to understand what data I need to pass where.

The big diff in `test` is mostly due to the removal of `mktempdir() do ... end` blocks; if one uses the "ignore whitespace" diff mode, one can see the actual (comparatively small) changes.

### DIFF
--- a/test/declarative.jl
+++ b/test/declarative.jl
@@ -3,122 +3,114 @@ using JSON, BinaryBuilder, Test
 import BinaryBuilder.BinaryBuilderBase: sourcify, dependencify, major, minor, patch, version
 
 @testset "Meta JSON" begin
-    mktempdir() do tmpdir
+    meta_json_buff = IOBuffer()
+
+    # Run autobuild() a few times to generate a moderately complex `meta.json`:
+    dict = get_meta_json(
+        "libfoo",
+        v"1.0.0",
+        [FileSource("https://julialang.org", "123123"), DirectorySource("./bundled")],
+        "exit 1",
+        [Linux(:x86_64)],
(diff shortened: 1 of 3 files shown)


staticfloat:
I'm generally in favor of this change, but I'll let @giordano have final approval.

test/declarative.jl
```diff
@@ -3,122 +3,114 @@ using JSON, BinaryBuilder, Test
...
+    dict = get_meta_json(
+        "libfoo",
+        v"1.0.0",
+        [FileSource("https://julialang.org", "123123"), DirectorySource("./bundled")],
+        "exit 1",
+        [Linux(:x86_64)],
+        Product[LibraryProduct("libfoo", :libfoo), FrameworkProduct("fooFramework", :libfooFramework)],
+        [Dependency("Zlib_jll")];
+    )
+    println(meta_json_buff, JSON.json(dict))
```
giordano, line 18: Isn't there anything interesting we can test on the dictionary?

fingolfin: I am sure one could. For example, by comparing it with the decoded JSON data in `objs` down below -- although that'd essentially test the JSON encoder/decoder; not sure if that's so useful?

However, I deliberately tried to restrict the changes in this PR to refactoring, not adding new stuff.

giordano: Sure, but there was a minimal test before :wink:

fingolfin: OK, I must have missed that, what test was there that I accidentally removed? It still serializes and deserializes the dict to JSON and back as before.

Issue #511, opened 2019-11-21
Creating a binary that links against Julia


fingolfin:
For various reasons, I need to create a binary build for a binary that links against Julia resp. `libjulia` (to interact with the Julia GC, but that should be irrelevant). This requires Julia >= 1.2

My problem is that in the build environment, there is no Julia installed, and in particular no Julia headers, so I can't do that. Is there any way around this? I realize this is probably a relatively rare thing to do, so perhaps BinaryBuilder does not cover it yet? If it does not, is there are at least principle interest in supporting this? (If so, I'd be inclined to try and help implement it).

Not sure if it is relevant, but note that I am currently using BinaryBuilder 0.1.4, because that's the version being shipped. But I see that there are tons of changes in this repository compared to that, which makes me wonder what I should be using.

giordano:
You'd probably need to use Julia as dependency.  There is a [builder in Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil/blob/master/J/Julia/build_tarballs.jl), but it probably works for a handful of platforms and it's for Julia 1.0.3.

In general, if you're looking for a builder you should checkout [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil), and contribute there a new one

fingolfin:
That Julia builder looks like it builds another copy of Julia, though? That's not sufficient; it is vital to link against the same Julia that is loading "my" binary. 

Yggdrasil looks like it doesn't play with the current release of BinaryBuilder (0.1.4) though, or does it?

giordano:
> Yggdrasil looks like it doesn't play with the current release of BinaryBuilder (0.1.4) though, or does it?

Yeah, sorry, forgot to reply on this.  Yggdrasil is using `master` version of BinaryBuilder.jl

barche:
> That Julia builder looks like it builds another copy of Julia, though? That's not sufficient; it is vital to link against the same Julia that is loading "my" binary.

Why is that, exactly? We're using those binaries to build the C++ part of CxxWrap, and since it is using dynamic linking it works fine with libjulia 1.0 - 1.3, even if at compile time only the 1.0 version was available.

fingolfin:
Sorry, with "link" I was referring to "dynamic link at runtime", not "creating the binary during the build stage".

[...]

PR #875, opened 2020-08-06
Add julia_compat keyword argument to build_tarballs

### LINKED ISSUE #856: Add a way to adjust the compat entry for Julia in JLL's `Package.toml`

fingolfin:
Resolves #856

I ended up using a plain string for the `julia_compat` argument, which is written unchanged into the `Project.toml` (however, it gets validated before, so the user can't insert invalid version specifiers). My first prototype used `Pkg.Types.VersionSpec`, but I moved away from that for several reasons:

1. As far as I can tell, `Pkg.Types.VersionSpec` is completely undocumented
2. Julia package developers are used to compat strings like `"~1.4"`; but they can't use that directly with `Pkg.Types.VersionSpec`; not everybody will now (or find intuitive) that  "~1.4" needs to be entered as `"1.4.0 - 1.4"`, or `"^1.4"` as `"1.4.0 - 1"`
3. One can ease this by using `Pkg.Types.semver_spec` but again that's completely undocumented (I still use it in this PR to validate the version specifier provided by the user, though)
4. Even if one uses `Pkg.Types.semver_spec`, after serializing and deserializing something like `Pkg.Types.semver_spec("~1.4")` one ends up with a `[compat]` entry of the form `julia = "1.4.0 - 1.4"`, and I also dislike it there, as people reading the `Project.toml` may not be used to it. Certainly this point can be argued about most (the package registry seems to prefer these "normalized" version ranges).

Regarding the name `julia_compat`, here are some alternatives I considered:
1. `julia_dependency` is what I originally used (still appears in the branch name for this PR) and was originally [suggested by @staticfloat in a comment](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/511#issuecomment-634974371); in that original suggestion, the name made sense, because the idea was to pass in something like `Pkg.Types.PackageSpec(name="julia", version=Pkg.Types.VersionSpec("1.0"))`. But since then I decided to change away from `PackageSpec` (as that seemed needlessly redundant). In the end, the content of the new parameter is sufficiently different from what's stored in `dependencies`, so I felt that calling it `julia_dependency` would be potentially confusing and misleading
2. `julia_version` -- this is not just a version in general, but a version *spec*
3. `julia_versionspec` -- would make sense if the content was really a `Pkg.Types.VersionSpec`, but I feel it's a bit misleading if it's just a string
4. `julia_compat` is what I finally settled on, as this is the Julia entry in the `[compat]` section
5. `compat_julia` would certainly be another alternative, but I wanted to get the bikeshed ready; it's easy to repaint it now if people want to :-)

I've decided this with `--deploy=local` and it worked there.

Of course now that I've written it, I am not completely sure anymore this is the way to move forward: for my usecase, and likely also that of @barche for CxxWrap.jl, it would actually be better if the Julia version was part of the "platform triplet" ("platform tuple"???) because we both really want a single JLL, just with different binaries for different Julia platforms. 
However, I have no idea how to get started on that...

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -6,6 +6,8 @@ using RegistryTools, Registrator
 import LibGit2
 import PkgLicenses
 
+const DEFAULT_JULIA_VERSION_SPEC = "1.0"
+
 const BUILD_HELP = (
     """
     Usage: build_tarballs.jl [target1,target2,...] [--help]
@@ -84,6 +86,10 @@ see what it can do, you can call it with `--help` in the `ARGS` or see the
 
 The `kwargs` are passed on to [`autobuild`](@ref), see there for a list of
 supported ones. A few additional keyword arguments are accept:
+
+* `julia_compat` can be set to a version string which is used to set the
+  supported Julia version in the `[compat]` section of the `Project.toml` of
+  the generated JLL package. The default value is `"1.0"`.
 
 * `lazy_artifacts` sets whether the artifacts should be lazy.
 
@@ -224,7 +230,7 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
         # Dependencies that must be downloaded
         dependencies,
     )
-    extra_kwargs = extract_kwargs(kwargs, (:lazy_artifacts, :init_block))
+    extra_kwargs = extract_kwargs(kwargs, (:julia_compat, :lazy_artifacts, :init_block))
 
     if meta_json_stream !== nothing
         # If they've asked for the JSON metadata, by all means, give it to them!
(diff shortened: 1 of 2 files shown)


staticfloat:
Looks good to me!

staticfloat:
Regarding the actual use of this; I agree that we want platform objects that have Julia versions embedded within them.  [We'll get those in 1.6](https://github.com/JuliaLang/julia/blob/sf/artifacts_rework/base/binaryplatforms.jl#L758-L759).

PR #880, opened 2020-08-12
WIP 1.6 Artifacts JLL generation

staticfloat:
Changing JLL generation to use the new 1.6 `Artifacts` stdlib, instead of Pkg, and optimizing things to do more work at compile time.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1151,15 +1151,15 @@ function build_jll_package(src_name::String,
 
                 global PATH_list, LIBPATH_list
             """)
-            if !isempty(dependencies)
-                print(io, """
-                # Initialize PATH and LIBPATH environment variable listings.
-                # From the list of our dependencies, generate a tuple of all the PATH and LIBPATH lists,
-                # then append them to our own.
-                foreach(p -> append!(PATH_list, p), ($(join(["$(getname(dep)).PATH_list" for dep in dependencies], ", ")),))
-                foreach(p -> append!(LIBPATH_list, p), ($(join(["$(getname(dep)).LIBPATH_list" for dep in dependencies], ", ")),))
-                """)
-            end
+            # if !isempty(dependencies)
+            #     print(io, """
+            #     # Initialize PATH and LIBPATH environment variable listings.
+            #     # From the list of our dependencies, generate a tuple of all the PATH and LIBPATH lists,
+            #     # then append them to our own.
+            #     foreach(p -> append!(PATH_list, p), ($(join(["$(getname(dep)).PATH_list" for dep in dependencies], ", ")),))
+            #     foreach(p -> append!(LIBPATH_list, p), ($(join(["$(getname(dep)).LIBPATH_list" for dep in dependencies], ", ")),))
+            #     """)
+            # end
             #println(io, "    initialize_path_list!(PATH_list, $(repr(tuple(["$(getname(dep)).PATH_list" for dep in dependencies]))))")
             #println(io, "    initialize_path_list!(LIBPATH_list, $(repr(tuple(["$(getname(dep)).LIBPATH_list" for dep in dependencies]))))")
 
@@ -1298,17 +1298,16 @@ function build_jll_package(src_name::String,
         const PATH_list = String[]
         const LIBPATH_list = String[]
 
-        # We determine, here, at compile-time, whether our JLL package has been dev'ed and overridden
-        override_dir = joinpath(dirname(@__DIR__), "override")
-        if isdir(override_dir)
-            function find_artifact_dir()
-                return override_dir
-            end
-        else
-            function find_artifact_dir()
+        function find_artifact_dir()
+            # We determine at compile-time whether our JLL package has been dev'ed and overridden
+            @static if isdir(joinpath(dirname(@__DIR__), "override"))
+                return joinpath(dirname(@__DIR__), "override")
+            else
                 return artifact"$(src_name)"
             end
+        end
 
+<<<<<<< HEAD
             \"\"\"
                 dev_jll()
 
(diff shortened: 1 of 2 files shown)


src/AutoBuild.jl
```diff
@@ -1166,53 +1150,98 @@ function build_jll_package(src_name::String,
 
                 global PATH_list, LIBPATH_list
             """)
+            if !isempty(dependencies)
```
giordano, line 1153: This is done already 20 lines above, before init, #844 :slightly_smiling_face:

staticfloat: You're right!  I need to get rid of the previous one, because I don't think it will work. :P

We can't do that initialization at compile-time because if we do, we bake in the values of `dep_jll.PATH_list` that existed at compile-time; then if we move artifacts (like as part of a PackageCompiler bundle) we end up with `PATH_list`'s that are out of sync. :(

staticfloat: I think I previously convinced you that this could be done at compile time, and now I realize it can't.

giordano: Yes, you did :joy: ok, we should revert #844, we can't wait for this PR to be merged

giordano: Done in #882

src/AutoBuild.jl
```diff
@@ -1239,91 +1236,72 @@ function build_jll_package(src_name::String,
...
-                @info("$(src_name)_ll dev'ed out to $(joinpath(Pkg.devdir(), "$(src_name)_jll")) with pre-populated override directory")
-            end
+        \"\"\"
+            dev_jll()
+        
+        Check this package out to the dev package directory (usually ~/.julia/dev),
+        copying the artifact over to a local `override` directory, allowing package
+        developers to experiment with a locally-built binary.
+        \"\"\"
+        function dev_jll()
```
KristofferC, line 1278: Can't we add something to Pkg or BinaryBuilder to not have to put this into every jll? Like `BinaryBuilder.dev("...")`?

src/AutoBuild.jl
```diff
@@ -1172,59 +1149,79 @@ function build_jll_package(src_name::String,
...
-        if VERSION < v"1.3.0-rc4"
+    # Generate target-demuxing main source file.
+    jll_jl = """
+        # Use baremodule to shave a few pieces of data off
+        baremodule $(src_name)_jll
+        using Base
+        if VERSION < v"1.6.0-DEV"
             # We lie a bit in the registry that JLL packages are usable on Julia 1.0-1.2.
             # This is to allow packages that might want to support Julia 1.0 to get the
             # benefits of a JLL package on 1.3 (requiring them to declare a dependence on
```
KristofferC, line 1227: Still some 1.3 references in this paragraph.

src/AutoBuild.jl
giordano: https://github.com/JuliaPackaging/JLLWrappers.jl/pull/2#issuecomment-675571311

KristofferC: That's loaded by every jll files as well though? It just feels like the wrong place to put it, tools for developing the package shouldn't have to be compiled into the package itself imo. It should be handled by some external thing like `JLLDeveloperTools.jl` or just BinaryBuilder.

giordano: That's an interesting point of view.

KristofferC: I mean, packages don't need an `update_me()` function bundled with them and compiled into their .ji file nor do they need to depend on Pkg to be updatable. Because the process of updating a package is handled externally (and is independent) to the code the package itself. This feels like kinda the same thing. If we want to make the jlls (and the common package they depend on) small, then moving everything dev like that the package won't run itself out to a developer package seems reasonable.

giordano: To be clear, I think you're probably right that the function doesn't need to be here at all :sweat_smile:

src/AutoBuild.jl
```diff
@@ -1522,8 +1500,9 @@ function build_project_dict(name, version, dependencies::Array{Dependency})
     end
     # Always add Libdl and Pkg as dependencies
     stdlibs = isdefined(Pkg.Types, :stdlib) ? Pkg.Types.stdlib : Pkg.Types.stdlibs
-    project["deps"]["Libdl"] = first([string(u) for (u, n) in stdlibs() if n == "Libdl"])
-    project["deps"]["Pkg"] = first([string(u) for (u, n) in stdlibs() if n == "Pkg"])
+    project["deps"]["Libdl"] = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
```
KristofferC, line 1503: Maybe there should be compat info added on JLLWrappers?

src/AutoBuild.jl
```diff
@@ -1522,8 +1500,9 @@ function build_project_dict(name, version, dependencies::Array{Dependency})
     end
     # Always add Libdl and Pkg as dependencies
```
KristofferC, line 1501: Still mentions Pkg

src/AutoBuild.jl
```diff
@@ -1065,287 +1068,87 @@ function build_jll_package(src_name::String,
...
-                filter!(!isempty, unique!(PATH_list))
-                filter!(!isempty, unique!(LIBPATH_list))
-                global PATH = join(PATH_list, $(repr(pathsep)))
-                global LIBPATH = join(vcat(LIBPATH_list, [$(init_libpath)]), $(repr(pathsep)))
+                    $(init_block)
+                """)
+            end
 
-                $(init_block)
+            print(io, """
```
giordano, line 1133:
```suggestion
            print(io, """
```

src/AutoBuild.jl
staticfloat: Yep, we're still going to express a dependency on `Pkg` because if it's an older version of Julia running, it will need to import `Pkg.Artifacts`, not `Artifacts`.

src/AutoBuild.jl
staticfloat: We always want the latest version of JLLWrappers, I think, because we're going to use that as the new way to "regenerate" JLL packages; since JLLWrappers is a bunch of macros that generate the JLLs at compile time, when we tag a new version of JLLWrappers, we will upgrade all the JLLs.

giordano (approved):

PR #883, opened 2020-08-14
Fix OpenLDAPClient_jll problems

staticfloat:
We weren't finding all of the libraries correctly, and even when we did, we didn't always complete all steps properly.

Fixes https://github.com/JuliaPackaging/Yggdrasil/issues/1438

### DIFF
--- a/src/Auditor.jl
+++ b/src/Auditor.jl
@@ -64,7 +64,7 @@ function audit(prefix::Prefix, src_name::AbstractString = "";
     translate_symlinks(prefix.path; verbose=verbose)
 
     # Inspect binary files, looking for improper linkage
-    predicate = f -> (filemode(f) & 0o111) != 0 || valid_dl_path(f, platform)
+    predicate = f -> (filemode(f) & 0o111) != 0 || valid_library_path(f, platform)
     bin_files = collect_files(prefix, predicate; exclude_externalities=false)
     for f in collapse_symlinks(bin_files)
         # If `f` is outside of our prefix, ignore it.  This happens with files from our dependencies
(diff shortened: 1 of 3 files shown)


staticfloat:
To elaborate on the issues fixed here:

First, `bin_files` is collected with `exclude_externalities=false`.  This includes files outside of the prefix, (such as the libraries from dependencies).  This is on purpose, because we use `bin_files` as "the list of all libraries we know about" (including dependencies) when passing it to `check_dynamic_linkage()`, but it's bad when we iterate over it, because we don't want to try and fix dependency libraries; we only want to fix our own.  Hence, we need to do `startswith(f, prefix.path)` and things like that, after passing `bin_files` through `collapse_symlinks` so that we're not fixing multiple symlinks that all point to the same file.

Secondly, we need to ensure that `shlib_files` actually finds all the libraries.  In this case, because we had a file named things like `foo.1.2.3-10.20.24.dylib`, that got rejected by the function that was supposed to tell us if that's a valid library name.  This rejection was wrong, because the method was foolishly using the `parse_dl_name_version()` function, which was originally written to find the `libgfortran` version that Julia is linked against; e.g. it is expecting much more regular style names.  That's fine; we'll just write our own library name validator.

aviks:
Thanks Elliot. Idle curiousity, what's special about this OpenLDAP build that triggered this, rather than the thousands of builds before this?

staticfloat:
The naming of the library `libldap-2.4.2.10.13.dylib`; it's not a legal version number, and that tripped up an overly-strict parser, causing the library to not get audited.

PR #889, opened 2020-08-18
Add comma to separate dlopen arguments

mortenpi:
I'm running into `ERROR: LoadError: InitError: UndefVarError: libhelfem_pathRTLD_LAZY not defined` when loading the _jll because the generated wrapper file looks like this:

```julia
    # Manually `dlopen()` this right now so that future invocations
    # of `ccall` with its `SONAME` will find this path immediately.
    global libhelfem_handle = dlopen(libhelfem_pathRTLD_LAZY | RTLD_DEEPBIND)
```

Comparing with [the definition of `dlopen_flags_str`](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/bed945d7bc51fd8e17336f9e3b65b1432680db7c/src/Products.jl#L127-L134), I'm pretty sure this should have a comma, but I am not sure why this hasn't become an error before (git blame points to #809 and I have been running on master for a while).

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1183,7 +1183,7 @@ function build_jll_package(src_name::String,
                     println(io, """
                         # Manually `dlopen()` this right now so that future invocations
                         # of `ccall` with its `SONAME` will find this path immediately.
-                        global $(vp)_handle = dlopen($(vp)_path$(BinaryBuilderBase.dlopen_flags_str(p)))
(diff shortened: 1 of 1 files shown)


mortenpi:
Actually, sorry for the noise, I should've done a bit more research. `dlopen_flags_str` was changed in JuliaPackaging/BinaryBuilderBase.jl#20, and I should probably run #880, where this is fixed, if I want to run the BinaryBuilderBase master.

PR #898, opened 2020-08-27
Use GitHub Actions for CI

giordano:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+              - BINARYBUILDER_USE_SQUASHFS: false
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/julia-buildpkg@latest
+      # - name: Setup git user
+      #   run: |
+      #     git config --global user.name "Binary Builder"
+      #     git config --global user.email "binary.builder@julialang.org"
+      - uses: julia-actions/julia-runtest@latest
+      - uses: julia-actions/julia-uploadcodecov@latest
```
SaschaMann, line 41: That action is deprecated. It's better to use julia-actions/julia-processcoverage and codecov/codecov-action instead. That also doesn't require the token anymore. See pkgtemplates for an example: https://github.com/invenia/PkgTemplates.jl/blob/5cf93cc4d8ed5b0b6887982046066fec238774c2/test/fixtures/AllPlugins/.github/workflows/ci.yml#L40

staticfloat:
Okay the latest error should be fixed by this: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/26

Mosè, what's your preferred way to deal with bumping BBB in a BB PR?  Should we `add` it at a particular branch in this PR's manifest?

giordano:
> Should we add it at a particular branch in this PR's manifest?

Sounds good

PR #900, opened 2020-08-31
Faster yet startup of JLL packages

PallHaraldsson:


PallHaraldsson:
The failed CI checks seem unrelated, as while working on 1.4, and not on 1.5, the code for 1.5 worked before, and the new code for 1.6 shouldn't be triggered on 1.5.

KristofferC:
See https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/880/files#diff-fa7ad01ce6e00f424f899e0521c2d27aR1246

PallHaraldsson:
I think Kristoffer is pointing out that my PR is redundant with the other Draft one (I'll lokk better at it later). I think, not sure Draft is a (new?) feature for what used to be warned with "WIP", so not ready? As far as my PR is, I think it shouldn't hurt, it seems the gain isn't too much.

Please merge either (or both?). We could unmerge mine later, if redundant (or conflicting with the other).

staticfloat:
We've completely revamped how JLL packages work, I think you will be pleased with the performance improvements that will be heading down the pipe very quickly.  :)

PallHaraldsson:
Hi @staticfloat,

On my just downloaded new master, it is a tiny bit faster:

```
julia> @time using GTK3_jll
  1.784891 seconds (2.40 M allocations: 157.035 MiB, 1.42% gc time)

is there some faster way? This does something, should it be the way?

julia> using Artifacts
julia> @time artifact"GTK3_jll"
ERROR: LoadError: Cannot locate '(Julia)Artifacts.toml' file when attempting to use artifact 'GTK3_jll' in 'Main'
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] @artifact_str(__source__::LineNumberNode, __module__::Module, name::Any)
   @ Artifacts /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Artifacts/src/Artifacts.jl:578
in expression starting at REPL[6]:1
```

Is this the way forward, after a file added (by some prcedure, do we need to regenerate packages here?).

I however noticed precompiling got really slow (might be acceptable for (a larger gain later):
```
julia> @time using GTK3_jll
[ Info: Precompiling GTK3_jll [77ec8976-b24b-556a-a1bf-49a033a670a6]
 84.284408 seconds (2.66 M allocations: 172.517 MiB, 0.04% gc time)

compared to older master:

julia> @time using GTK3_jll
[ Info: Precompiling GTK3_jll [77ec8976-b24b-556a-a1bf-49a033a670a6]
 12.869754 seconds (2.72 M allocations: 175.623 MiB)
```

giordano:
To be clear, (almost) _nothing_ changed since you rebuilt all packages related to GTK3, so don't expect to see _any_ improvement now.

Please, be patient.

PallHaraldsson:
Yes, just asking (it wasn't clear if rebuilding needed for (some of the) changes, theoretically could be done faster without rebuilding?), and if rebuilding needed curious about what needs changing first.

PR #902, opened 2020-09-01
Add Auditor arch mapping form Darwin AArch64

Keno:
Depends on https://github.com/staticfloat/ObjectFile.jl/pull/26


giordano:
Need to require ObjectFile v0.3.6

PR #905, opened 2020-09-03
Add support for parallel universes

Keno:
This adds the concept of a "universe" for autobuilds. A universe
is essentially a depot that has its own copy of the General registry.
When building packages using `--deploy=universe` the built jll gets
registered into this copy of the registry and subsequent builds can
pick it up from there as if it had been registered with the real
general registry. This allows successive builds of dependent packages
without first having to go register the dependency. The universe
to use is specified using `--universe=<name>`. If a universe by
that name does not exist, it is created. Otherwise the previously
created universe is re-used.

cc @KristofferC @fredrikekre


src/AutoBuild.jl
```diff
@@ -856,47 +936,70 @@ end
...
+        try
+            with_gh_auth() do creds
+                LibGit2.fetch(repo; credentials=creds)
+            end
+            origin_master_oid = LibGit2.GitHash(LibGit2.lookup_branch(repo, "origin/master", true))
+            LibGit2.reset!(repo, origin_master_oid, LibGit2.Consts.RESET_HARD)
+            if string(LibGit2.head_oid(repo)) != string(origin_master_oid)
+                LibGit2.branch!(repo, "master", string(origin_master_oid); force=true)
+            end
+        catch
```
staticfloat, line 1002: I don't really like big `try/catch` chunks like this with no error handling; what is the error you're trying to ignore?

src/AutoBuild.jl
```diff
@@ -912,7 +1015,7 @@ function rebuild_jll_package(obj::Dict;
                              lazy_artifacts::Bool = false,
                              from_scratch::Bool = true)
     if build_version === nothing
-        build_version = BinaryBuilder.get_next_wrapper_version(obj["name"], obj["version"])
+        build_version = BinaryBuilder.get_next_wrapper_version(nothing, obj["name"], obj["version"])
```
staticfloat, line 1018: Let's thread a `universe` kwarg through `rebuild_jll_package` and default it to `nothing`.  This will be useful on Yggdrasil.

staticfloat (changes requested):
This is really nice, and I'm a big fan of melodramatic names. :)

src/AutoBuild.jl
```diff
@@ -160,8 +164,57 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
...
+    if universe
+        if deploy
+            if deploy_repo != "universe"
+                error("When deploying from a universe, must use --deploy=universe")
+            end
+            deploy_universe = true
+            deploy_bin_repo = deploy_jll_repo = "JuliaBinaryWrappers/$(src_name)_jll.jl"
+        end
+
+        # Find the universe directory by ascending from pwd() until we find a "universes" directory
```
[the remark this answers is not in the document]
Keno: I thought about it, but I do kinda want it explicitly available in Yggdrasil in case you need to go in and edit something. Putting it in Scratch makes it really hard to find.

src/AutoBuild.jl
Keno: The repo may not have an upstream.

src/AutoBuild.jl
Keno: ok

src/AutoBuild.jl
```diff
@@ -380,37 +440,57 @@ if VERSION < v"1.4-"
     Pkg.Operations.registered_paths(ctx::Pkg.Types.Context, uuid::UUID) = Pkg.Operations.registered_paths(ctx.env, uuid)
 end
 
-function get_next_wrapper_version(src_name, src_version)
+function in_universe(f, universe; update = false)
```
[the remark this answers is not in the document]
Keno: Clever, but I think `in_universe` is clearer

src/AutoBuild.jl
```diff
@@ -13,7 +13,7 @@ const BUILD_HELP = (
     Usage: build_tarballs.jl [target1,target2,...] [--help]
                              [--verbose] [--debug]
                              [--deploy] [--deploy-bin] [--deploy-jll]
-                             [--register] [--meta-json]
+                             [--register] [--meta-json] [--universe=<universe>]
```
giordano, line 16: May document the universe flag further below in the docstring

src/AutoBuild.jl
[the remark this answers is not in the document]
Keno: It's also the same thing that `git` does to find its `.git` folders.

src/AutoBuild.jl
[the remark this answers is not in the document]
staticfloat: Heh, yeah.

src/AutoBuild.jl
staticfloat: You mean in the event that it was previously created during a build with `--deploy=universe`?  Can we test for the existence of the `origin/master` branch and wrap this in an `if` then?  I don't want to swallow unrelated errors.

Keno: sure

src/AutoBuild.jl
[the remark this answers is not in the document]
staticfloat: I'd like to have a way for these universes to get GC'ed.  I think Fredrik and I are going to add optional user-defined time limits to scratch spaces so that you can e.g. `@get_scratch("Universe-$(universe)"; lifetime=Days(3))` and if you don't use it again in three days, it gets auto-cleaned up.

That's the main reason I want to use scratch spaces.  Perhaps we can add a symlink in `build`, so that if you do `julia build_tarballs.jl --universe=SS_rebuild --deploy=universe`, it then creates not only `build/SuiteSparse*.tar.gz`, but also `build/universes/SS_rebuild` which points to the universe where that was built?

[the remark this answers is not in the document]
Keno: Why should universes get explicitly gc'ed? Seems like something that should be very much up to the user.

src/AutoBuild.jl
```diff
@@ -160,8 +172,57 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
...
+            deploy_universe = true
+            deploy_bin_repo = deploy_jll_repo = "JuliaBinaryWrappers/$(src_name)_jll.jl"
+        end
+
+        # Find the universe directory by ascending from pwd() until we find a "universes" directory
+        dir = pwd()
+        while !isdir(joinpath(dir, "universes"))
+            dir = dirname(dir)
+            if dir == "/"
+                error("Could not find `universes` diretory in any parent of pwd")
```
fredrikekre, line 190:
```suggestion
                error("Could not find `universes` directory in any parent of pwd")
```

src/AutoBuild.jl
```diff
@@ -160,8 +172,57 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
...
+                @warn "Universe does not exist, but deploy not specified. Skipping Universe creation."
+                universe_path = nothing
+            end
+        end
+    end
+
     if deploy_bin || deploy_jll
-        code_dir = joinpath(Pkg.devdir(), "$(src_name)_jll")
+        if universe
+            code_dir = joinpath(universe_path, "clones", string(jll_uuid("$(src_name)_jll")))
```
fredrikekre, line 222: What's the reason for this?

Keno: Makes the commit available to the regular logic to just check it out from `clones`

fredrikekre: Okay. Could alternatively put in the devdir path as the url in the registry, right? I guess this avoids one layer of indirection since Pkg would cache it here anyway.

Keno: Yes, this seemed easier.

staticfloat: I'm confused; the `clones` directory in my normal depot is a bunch of bare repositories, and the names are all numeric, not UUIDs.

src/AutoBuild.jl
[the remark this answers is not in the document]
staticfloat: I think these should be auto-GC'ed because the usecase for universes is for experimentation, not for users to use this as their main source of JLLs.  I want to use these within Yggdrasil to allow us to open a PR that builds multiple, interdependent packages, and use the builds from the bottom of the stack as dependencies for the top of the stack, without having to register in-between.  If we're creating a universe per-PR, I want to be sure that they will be removed after a while.

src/AutoBuild.jl
fredrikekre: Yea, cold be stored wherever. I don't think Pkg will pick this up.

PR #907, opened 2020-09-12
Add multiple options to enter the debug shell when building

### LINKED ISSUE #879: Drop to environment after successful build
With the `--debug` flag, running `build_tarballs.jl` will drop down into the cross compilation environment if the build fails. However, it does not do that when the build succeds, but audit fails. 

I often find myself adding a `false` statement to the end of the build script when trying to troubleshoot audit failures, or needing to inspect the built libraries for any other reason. As you'd expect, I've sometimes forgotten to remove it before making a PR. 

I think it would be useful to have a `--wait` (or some such) option to `build_tarballs.jl` to ensure that the build drops down to the cross environment even after a successful build.

giordano:
Fix #879.  CC: @aviks


src/AutoBuild.jl
```diff
@@ -212,6 +220,15 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
         end
     end
 
+    # Modify script for debugging
+    if debug
+        if debug_mode == "begin"
+            script = "false\n" * script
```
giordano, line 226: In this case it could simply be
```suggestion
            script = "false"
```
I don't have a strong opinion on either way

staticfloat: Let's keep the script, we may eventually want to add a "continue" option.

staticfloat (approved):

staticfloat:
I love it!

PR #911, opened 2020-09-19
Fix permissions before cleaning-up directory during rebuild of JLLs

giordano:
Some packages, like `ICU_jll`, have files with pesky permissions which prevent
cleaning-up the temporary directory.


staticfloat:
Included as part of https://github.com/JuliaPackaging/BinaryBuilder.jl/commit/7f242da7085bc87bf43a95a3e957c2f7f86355ea

PR #914, opened 2020-09-20
Add compat entry for JLLWrappers in generated wrappers

giordano:


staticfloat:
I always forget; will this limit us to `1.1.X` or to `1.X`?

giordano:
The latter

PR #918, opened 2020-09-24
Massive upgrade to 1.6+

staticfloat:
Oh, and also, test using AZP.  :P


Project.toml
```diff
@@ -45,7 +45,7 @@ ProgressMeter = "1"
 Registrator = "1.1"
 RegistryTools = "1.3.4"
 ghr_jll = "0.13.0"
-julia = "1.4"
+julia = "1.6"
```
giordano, line 48: Bump version of BinaryBuilder to 0.3.0?

deps/SnoopCompile/precompile/precompile_BinaryBuilder.jl
```diff
@@ -1,205 +0,0 @@
-const __bodyfunction__ = Dict{Method,Any}()
```
giordano, line 1: :scream:

src/Auditor.jl
```diff
@@ -226,79 +227,72 @@ function audit(prefix::Prefix, src_name::AbstractString = "";
...
+
+    # Search for this platform's march in the march list
+    idx = findfirst(m -> m == march(platform), march_list)
+    if idx === nothing
+        throw(ArgumentError("Microarchitecture \"$(march(platform))\" not valid for architecture \"$(arch(platform))\""))
     end
-    return [first(list)]
+
+    # Return all up to that index
+    return march_list[1:idx]
```
giordano, line 258:
```suggestion
    return march_list[begin:idx]
```
:grin:

src/BinaryBuilder.jl
```diff
@@ -4,15 +4,14 @@ using BinaryBuilderBase
 using ObjectFile
 using GitHub
 import InteractiveUtils
-using Pkg, Pkg.BinaryPlatforms, Pkg.PlatformEngines, Pkg.Artifacts
+using Pkg, Base.BinaryPlatforms, Pkg.Artifacts
```
giordano, line 7: `Pkg.Artifacts` or `Artifacts`?

src/auditor/extra_checks.jl
```diff
@@ -1,20 +1,20 @@
...
+                """, '\n' => ' ')
+                @warn(strip(msg))
+            end
+            return false
         end
-        return false
     end
     return true
-end
+end
```
giordano, line 20: :sob:

src/BinaryBuilder.jl
staticfloat: We actually need to use `Pkg.Artifacts` because we do things like `bind_artifact!()` and such, which exist only in the `Pkg` version.

src/Auditor.jl
staticfloat: TIL about `begin`

deps/SnoopCompile/precompile/precompile_BinaryBuilder.jl
staticfloat: We can add it back in the future, but yeah, it was too out of date.  :P

giordano:
Since this now requires v1.6, can we integrate #911 with the correct way to cleanup the temp directory?

macd:
Does the compat for BinaryBuilderBase need to be bumped to 0.5?

staticfloat:
It already was.

macd:
I missed that! Sorry!

src/auditor/instruction_set.jl
```diff
@@ -36,12 +36,14 @@ function instruction_mnemonics(path::AbstractString, platform::Platform)
...
     # Run objdump to disassemble the input binary
-    if platform isa MacOS || platform isa FreeBSD
+    if Sys.isbsd(platform)
         objdump_cmd = "llvm-objdump -d $(basename(path))"
     else
         objdump_cmd = "\${target}-objdump -d $(basename(path))"
     end
+    @info("Fetching mnemonics for $(path)")
     run_interactive(ur, `/bin/bash -c "$(objdump_cmd)"`; stdout=output, stderr=devnull)
+    @info("done!", readdir(dirname(path)))
```
giordano, line 46: Remember to remove this before merging

staticfloat: I was just editing that, heh.  I guess we're both watching the logs. :P

src/wizard/obtain_source.jl
```diff
@@ -168,6 +168,8 @@ function download_source(state::WizardState)
             source_path = joinpath(state.workspace, "$(name)_$n$ext")
         end
 
+        # Initialise `gen_download_cmd`
+        Pkg.probe_platform_engines!()
         download_cmd = Pkg.PlatformEngines.gen_download_cmd(url, source_path)
```
giordano, line 173: `gen_download_cmd` was crashing badly without a call to `Pkg.probe_platform_engines!()`, did anything change in Pkg?

Can we just use `Downloads.download` instead?

giordano:
Something is odd with the storage setup again, [this build](https://dev.azure.com/JuliaPackaging/BinaryBuilder.jl/_build/results?buildId=5912&view=results) is downloading again all the shards for the non-docker jobs

src/wizard/obtain_source.jl
staticfloat: Yes, we should use that everywhere.

staticfloat:
Until all three workers have both the unpacked and squashfs shards, that will continue to happen every now and then.

giordano:
Right, makes sense!

PR #919, opened 2020-09-27
replace_includes

bramtayl:
I thought this script was useful to replace headers like "#include <fcntl.h>" with headers like "#include <linux/fcntl.h>". I have no idea how to make it executable from within BinaryBuilder, so this PR just a sketch. I'm not sure where to put docs but I've written up an idea below:

```
"""
    replace_includes old_header new_header

Replace all C includes with another include. For example, you might want to replace `fcntl` with `linux/fcntl` to use newer libc features. Note that `old_header` and `new_header` will be passed directly to `grep` and `sed`, and so should be escaped accordingly (so, for example, put two backslashes before a forward slash).
"""
replace_headers
```


staticfloat:
Hey, this is neat!

The best place for this to go would be in our library of functions that are defined in the rootfs, [here](https://github.com/JuliaPackaging/Yggdrasil/tree/master/0_RootFS/Rootfs/bundled/utils).  This is where things like `flagon` and `install_license` are stored. :)

Secondly, I suggest using some regex goodness to make this as resilient as possible.  So instead of `"#include <$1.h>"`, you can do something like `"^#\s*include\s*<$1\.h>"`, so as to catch things like `# include  <stdio.h>`.

bramtayl:
Ok! https://github.com/JuliaPackaging/Yggdrasil/pull/1714

PR #932, opened 2020-10-04
Teach audit to check multiple CSL libraries

giordano:
In addition to `libgomp`, it's possible that a library links against libatomic,
so we need to check also that one.

### DIFF
--- a/src/auditor/compiler_abi.jl
+++ b/src/auditor/compiler_abi.jl
@@ -66,7 +66,7 @@ end
 function check_csl_libs(oh::ObjectHandle, platform::AbstractPlatform; verbose::Bool=false,
                         has_csl::Bool=true, csl_libs::Vector{String}=["libgomp", "libatomic"])
     if has_csl
-        # No need to do any check, CompilerSupportLibraries_jll si already a dependency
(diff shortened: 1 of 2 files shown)


src/auditor/compiler_abi.jl
```diff
@@ -63,24 +63,33 @@ function check_libgfortran_version(oh::ObjectHandle, platform::AbstractPlatform;
...
-function check_libgomp(oh::ObjectHandle, platform::AbstractPlatform; verbose::Bool = false,
-                       has_csl::Bool = true)
-    has_libgomp = false
-    try
-        libs = basename.(path.(DynamicLinks(oh)))
-        has_libgomp = length(filter(l -> occursin("libgomp", l), libs)) >= 1
+function check_csl_libs(oh::ObjectHandle, platform::AbstractPlatform; verbose::Bool=false,
+                        has_csl::Bool=true, csl_libs::Vector{String}=["libgomp", "libatomic"])
+    if has_csl
+        # No need to do any check, CompilerSupportLibraries_jll si already a dependency
```
staticfloat, line 69:
```suggestion
        # No need to do any check, CompilerSupportLibraries_jll is already a dependency
```

staticfloat (approved):

Issue #727, opened 2020-04-02
[Verifier] Have the verifier extract and check the GlibC/libc++ version required by the package


vchuravy:
While we provide a Glibc version, the actual version needed might be lower. In general it would be good to record some metadata about the dependency assumption
a package makes. As an example : `nm -D --with-symbol-versions LLVM.so` yields entries like  `U _ZTVNSt7__cxx1115basic_stringbufIcSt11char_traitsIcESaIcEEE@GLIBCXX_3.4.21`

giordano:
More generally, we should think about a way to record the metadata, I'm not sure where the glibc information could go as it is now.

The glibcxx version may actually be already available by looking at the gcc shard used: https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/e53faa3629eebd9492a91cea1132e7da6599ea83/src/Rootfs.jl#L342-L349  But this is probably an upper bound, looking at the symbols in the library products, as suggested by Valenin, can give a more accurate requirement.

giordano:
Before I venture again into detecting the glibcxx version, we already have a function for this: https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/4a0e367f0ceb49b8fce8b0bafa96cf9c7da98153/src/auditor/compiler_abi.jl#L87-L113

PR #934, opened 2020-10-04
Add function to detect GLIBCXX

giordano:
This would like to be a step towards #727, but currently doesn't work at all, as
`Symbols(oh)` ignores the GLIBCXX annotations.


giordano:
We already have this function :facepalm:

PR #944, opened 2020-10-13
compat bounds: fix version range test for 'x.y.0-x.y'

benlorenz:
Without this change we have the following behaviour (using the `exactly_this_version` function from `AutoBuild.jl`):
```julia
julia> exactly_this_version(VersionSpec("8.3"))
"=8.3"
julia> exactly_this_version(VersionSpec("8.3.0-8.3"))
"=8.3"
julia> exactly_this_version(VersionSpec("8.3.1-8.3"))
"8.3.1-8.3"
```
Similarly `8.0-8` or `8` returns `=8` instead if just `8`.
In these cases the lower bound equals the upper bound but not all three version components are fixed.

I tried looking into Pkg.jl (`src/versions.jl`) but could not think of any way to do this without accessing that field directly.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1354,6 +1354,7 @@ function build_project_dict(name, version, dependencies::Array{Dependency}, juli
     end
     function exactly_this_version(v::Pkg.Types.VersionSpec)
         if length(v.ranges) == 1 &&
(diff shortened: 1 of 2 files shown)


giordano:
The CI issue is hopefully fixed by #950

Issue #929, opened 2020-10-04
Add the ability to have multiple in-progress wizard builds


Keno:
I was doing one wizard build, but then realized it needed a dependency that was not in Yggdrasil yet, so I wanted to interrupt the wizard to build that one. At the moment it's inconvenient to interrupt one wizard execution to start another one. We should ideally be able to save a wizard execution for later. We already have most of the support for this, it just needs a UI.

PR #962, opened 2020-11-02
Update state serialization

staticfloat:
* Use `Scratch` instead of `MutableArtifacts.toml`

* Add tests to ensure that the state is serialized faithfully

Close #901


giordano:
Does it help with #929?

test/wizard.jl
```diff
@@ -414,6 +414,18 @@ end
...
+@testset "Wizard - state serialization" begin
+    for state_generator in (WizardState, step2_state, step3_state, step7_state)
+        mktempdir() do dir
+            state = state_generator()
+
+            Wizard.save_wizard_state(state, dir)
+            @test Wizard.load_wizard_state(dir) == state
+        end
+    end
+end
```
giordano, line 426: :pray:

staticfloat:
A little; you can serialize things out to different directories easier now, but this was primarily aimed at https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/901

giordano (approved):
Looks good as far as I can tell :smile:

test/wizard.jl
```diff
@@ -414,6 +414,18 @@ end
     end
 end
 
+@testset "Wizard - state serialization" begin
+    for state_generator in (Wizard.WizardState, step2_state, step3_state, step7_state)
```
giordano, line 418: Unrelated to this PR, but I wonder if we can rename `Wizard.WizardState` to `Wizard.State`

staticfloat: I think it might be confusing to see `State` objects floating around within the `Wizard` module.  More likely, we can just export `WizardState` from `Wizard`, and then `using BinaryBuilder.Wizard` in this test suite.

giordano:
Still failing for no good reason :disappointed:

giordano:
The docker job hangs at
```
Test Summary:       | Pass  Total
Wizard - Deployment |    8      8
┌ Warning: Legacy `RadioMenu` interface is deprecated, set a configuration option such as `RadioMenu(options; charset=:ascii)` to trigger the new interface.
│   caller = load_wizard_state(wizard_state_dir::String; as_is::Bool) at Wizard.jl:60
└ @ BinaryBuilder.Wizard /data/staticfloat/bb_azp/rootfs/agent_3/_work/1/s/src/Wizard.jl:60
##[warning]Legacy `RadioMenu` interface is deprecated, set a configuration option such as `RadioMenu(options; charset=:ascii)` to trigger the new interface.
Would you like to resume the previous incomplete wizard run?
 > Resume previous run
```
this probably needs to be fixed

giordano:
> The docker job hangs at

This is still an issue, I guess it was never fixed

test/wizard.jl
```diff
@@ -414,6 +414,18 @@ end
...
+        mktempdir() do dir
+            state = state_generator()
+
+            Wizard.save_wizard_state(state, dir)
+            @test Wizard.load_wizard_state(dir) == state
```
giordano, line 423: This is untestable as it is now, `Wizard.load_wizard_state` requires interactive input, but this function doesn't take a state as argument.  I take you didn't try it, I don't see how it could possibly work :stuck_out_tongue:

giordano: [Past Mosè to the rescue](https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/761): we can use the `as_is` keyword argument to skip questions :grin:

PR #964, opened 2020-11-07
Properly plumb kwarg

staticfloat:


giordano:
Now someone please explain me why tests here pass and fail in #962 because of the hashes :confused:

PR #967, opened 2020-11-20
Update TagBot integration

fingolfin:
Cf. https://discourse.julialang.org/t/ann-required-updates-to-tagbot-yml/49249 and https://github.com/marketplace/actions/julia-tagbot.


giordano:
Thanks!

PR #970, opened 2020-11-25
Fix valid_library_path to reject .sot extension

### LINKED ISSUE #857: Weird warning for files whose name ends in "so" (but not ".so")
When building a JLL which uses Ncurses_jll, one sees warnings like that:
```
##[warning]Could not probe /agent/_work/1/s/G/GAP/build/aarch64-linux-gnu-libgfortran4-cxx11/sjigd7D9/destdir/share/terminfo/c/cons60-iso for an SONAME!
```
My guess here is that some code somewhere assumes that all files whose name ends in "so" are shared libraries. If that's correct, that code ought to be changed to instead check for a suffix of ".so".

Alas, I couldn't find any code doing that, so perhaps my guess is completely wrong? But then why does it print this warning?

fingolfin:
Fixes #857 (again)

### DIFF
--- a/test/auditing.jl
+++ b/test/auditing.jl
@@ -1,5 +1,5 @@
 using BinaryBuilder.Auditor
-using BinaryBuilder.Auditor: compatible_marchs
+using BinaryBuilder.Auditor: compatible_marchs, valid_library_path
 
 # Tests for our auditing infrastructure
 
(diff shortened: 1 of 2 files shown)


test/basic.jl
```diff
@@ -240,3 +241,29 @@ end
     dict = build_project_dict("Clang", v"9.0.1+2", dependencies)
     @test dict["compat"]["libLLVM_jll"] == "8.3"
 end
+
+
+@testset "valid_library_path" begin
```
giordano, line 246: Why not testing in [`test/auditing.jl`](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/d0437c4ed649fa335ac825f04d5e2410ea1f04c7/test/auditing.jl)?

fingolfin: I actually looked at `test/auditing.jl` first, but didn't see where to fit it in. I am completely neutral on this, though. If you tell me where you want it, I'll move it there

giordano: It's just that this is "audit" stuff anyway (and so you don't need to specifically load `Auditor` at the beginning), I don't think this function is used anywhere else (but I din't actually check this claim)

giordano:
By the way, it'd be great if this function was actually accessible from `Base` so we don't have to apply these fixes twice all the time :stuck_out_tongue:

fingolfin:
Not sure what you mean with "accessible from `Base`"?

In `Base` there is already `parse_dl_name_version` but it serves a slightly different purpose. Of course one could move `valid_library_path` there but then it wouldn't be called by `Base`. I don't really see a good way to unify the two functions; well I guess one could share the regex, by adding a function which returns them. But that seems artificial?

test/basic.jl
fingolfin: moved it

giordano:
> Not sure what you mean with "accessible from Base"?

The same functionality which is used in `Base` is inside a slightly unrelated function, as you said.

test/auditing.jl
```diff
@@ -516,3 +516,28 @@ end
         @test readlink(joinpath(testdir, "lib", "libfoo.so")) == "libfoo.so.1.0.0"
     end
 end
+
+@testset "valid_library_path" begin
+    linux = Platform("x86_64", "linux")
+    macos = Platform("x86_64", "macos")
+    windows = Platform("x86_64", "windows")
+
+    @test valid_library_path("/usr/libc.dylib", macos)
```
giordano, line 525: You might still need to import the function?

giordano:
Tests exceptionally didn't time out for unknown reasons! :tada: 

Thanks for the PR!

PR #972, opened 2020-11-29
Update BinaryBuilderBase

giordano:


giordano:
The failure in the Rust toolchain for 32-bit Windows is unrelated to https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/80 or https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/82, I can reproduce it with https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/79 but not the commit before it.  So it looks like the new Rust toolchain has a problem for that platform

giordano:
All other tests are now passing.  I'm going to merge this anyway.

PR #976, opened 2020-12-12
[Auditor] Make sure libraries for eabihf match this ABI

### LINKED ISSUE #975: Improve audit checks for the ABI of armv7l binaries
The armv7l binaries of [`ICU_jll` v67.1.0+2](https://github.com/JuliaBinaryWrappers/ICU_jll.jl/releases/tag/ICU-v67.1.0%2B2) are built for the wrong ABI:
```
% readelf -h libicudata.so.67.1
ELF Header:
  Magic:   7f 45 4c 46 01 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF32
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           ARM
  Version:                           0x1
  Entry point address:               0x1e8
  Start of program headers:          52 (bytes into file)
  Start of section headers:          28397040 (bytes into file)
  Flags:                             0x5000200, Version5 EABI, soft-float ABI
  Size of this header:               52 (bytes)
  Size of program headers:           32 (bytes)
  Number of program headers:         4
  Size of section headers:           40 (bytes)
  Number of section headers:         11
  Section header string table index: 8
```
We should have the hard-float ABI, ~@staticfloat also suggested that the ABI version 5 might be wrong~ _Edit_: the ABI version is probably OK.  BinaryBuilder builds [ICU](https://github.com/JuliaPackaging/Yggdrasil/blob/f38d35d685bbba5291fd888c385a5d37791e9fbb/I/ICU/build_tarballs.jl) for `armv7l-linux-gnueabihf-cxx11` with zero warnings, audit should catch all these issues

giordano:
Fix #975.

### DIFF
--- a/src/auditor/extra_checks.jl
+++ b/src/auditor/extra_checks.jl
@@ -9,9 +9,19 @@ function check_os_abi(oh::ObjectHandle, p::AbstractPlatform, rest...; verbose::B
             if verbose
                 msg = replace("""
                 $(basename(path(oh))) has an ELF header OS/ABI value that is not set to FreeBSD
-                ($(ELF.ELFOSABI_FREEBSD)), this may be an issue at link time"
+                ($(ELF.ELFOSABI_FREEBSD)), this may be an issue at link time
                 """, '\n' => ' ')
                 @warn(strip(msg))
+            end
+            return false
+        end
+    elseif call_abi(p) == "eabihf"
+        # Make sure the object file has the hard-float ABI.  See Table 4-2 of
+        # "ELF for the ARM Architecture" document
+        # (https://developer.arm.com/documentation/ihi0044/e/).
+        if iszero(header(oh).e_flags & 0x400)
+            if verbose
+                @error("$(basename(path(oh))) does not match the hard-float ABI")
             end
             return false
         end
--- a/test/auditing.jl
+++ b/test/auditing.jl
@@ -284,8 +284,8 @@ end
                     """,
                     # Build for our platform
                     [platform],
-                    # Ensure our executable products are built
-                    Product[LibraryProduct("libfoo", :libfoo)],
+                    # Ensure our library product is built
+                    [LibraryProduct("libfoo", :libfoo)],
                     # No dependencies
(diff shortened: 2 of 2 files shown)


src/auditor/extra_checks.jl
```diff
@@ -9,12 +9,22 @@ function check_os_abi(oh::ObjectHandle, p::AbstractPlatform, rest...; verbose::B
...
                 """, '\n' => ' ')
                 @warn(strip(msg))
             end
             return false
         end
+    elseif compiler_abi(p) == "eabihf"
+        # Make sure the object file has the hard-float ABI.  See Table 4-2 of
+        # "ELF for the ARM Architecture" document
+        # (https://developer.arm.com/documentation/ihi0044/e/).
+        if iszero(header(oh).e_flags & 0x400)
```
giordano, line 22: Not sure there are more specific ways to check these flags, I couldn't find anything else in `ObjectFile.jl`

src/auditor/extra_checks.jl
```diff
@@ -9,12 +9,22 @@ function check_os_abi(oh::ObjectHandle, p::AbstractPlatform, rest...; verbose::B
...
+            if verbose
+                @error("$(basename(path(oh))) does not match the hard-float ABI")
```
giordano, line 24: Maybe `@error` is a bit excessive, but we're basically building for a wrong platform, seems quite important to flag this issue.

Please do suggest better messages!

giordano (commented):
Not sure how to produce a useful binary to test this (besides downloading the broken `ICU_jll` tarballs), `libicudata.so` seems to be built in a rather contrived way.

src/auditor/extra_checks.jl
staticfloat: I think erroring is correct.

staticfloat (approved):

PR #984, opened 2020-12-30
Re-export nbits() from BinaryBuilderBase

fingolfin:
This is useful in the same way as libc, arch, etc. are.

See https://github.com/JuliaPackaging/Yggdrasil/pull/2323#discussion_r549897586


giordano:
Looks same failure as in #983, no idea what's going on

fingolfin:
Hmm, where is "At line 322, ABORTED (2: No such file or directory)!" coming from?

giordano:
Failure is clearly unrelated, I'll try to investigate separetely

Issue #342, opened 2018-08-31
comment out rather than totally removing platforms in wizard-generated build_tarballs.jl


ssfrr:
If some of my builds failed while running the wizard, I might want to get through the wizard to get my progress saved to disk before I sit down to fix each build. In this case it would be nice if the generated `build_tarballs.jl` just commented out the builds that weren't working in the `platforms` list, so I can easily see what worked and what didn't, and work on the broken ones one at a time.

E.g., from [lpoptBuilder](https://github.com/staticfloat/IpoptBuilder/blob/master/build_tarballs.jl):

```
platforms = [
    Windows(:i686),
    Windows(:x86_64),
    Linux(:i686, :glibc),
    Linux(:x86_64, :glibc),
    Linux(:aarch64, :glibc),
    Linux(:armv7l, :glibc),
    # ppc64le isn't working right now....
    #Linux(:powerpc64le, :glibc),
    MacOS()
]
```

PR #989, opened 2021-01-09
comment out failing platforms

matbesancon:
Attempt at https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/342
If a platform has failed, comment it out instead of removing it.

### DIFF
--- a/src/wizard/deploy.jl
+++ b/src/wizard/deploy.jl
@@ -20,7 +20,16 @@ function print_build_tarballs(io::IO, state::WizardState)
     else
         platforms_string = """
         [
-            $(strip(join(repr.(state.platforms),",\n    ")))
+        """
+        for platform in state.platforms
+            rp = repr(platform)
+            if platform in state.failed_platforms
+                platforms_string *= "# $rp,\n"
+            else
+                platforms_string *= "  $rp,\n"
+            end
+        end
+        platforms_string *= """
         ]
         """
     end
--- a/src/wizard/interactive_build.jl
+++ b/src/wizard/interactive_build.jl
@@ -871,20 +871,20 @@ function step6(state::WizardState)
     println(state.outs)
 
     if choice == 1
-        filter!(p->!(p in state.failed_platforms), state.platforms)
+        # filter!(p->!(p in state.failed_platforms), state.platforms)
         state.step = :step7
     elseif choice == 2
         plats = collect(state.failed_platforms)
         if length(plats) > 1
-            choice = request(terminal,
+            plat_choice = request(terminal,
                 "Which platform would you like to revisit?",
                 RadioMenu(map(repr, plats); charset=:ascii))
             println(state.outs)
         else
-            choice = 1
+            plat_choice = 1
         end
-        if step5_internal(state, plats[choice])
-            delete!(state.failed_platforms, plats[choice])
+        if step5_internal(state, plats[plat_choice])
+            delete!(state.failed_platforms, plats[plat_choice])
         end
         # Will wrap back around to step 6
     elseif choice == 3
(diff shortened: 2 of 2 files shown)


matbesancon:
Test Docker_Unpacked and Test Privileged_SquashFS were failing on first try, will see if it was just error from non-related things

giordano:
I have no idea what's wrong with the registry, but it picks up old versions of JLLWrappers

giordano:
I assume you tested this locally?

matbesancon:
Partially but artifact download was freezing this weekend, not sure if my connexion, laptop or the registries issues were the reason.
I'm trying now with HiGHS. The package tests from BB itself are passing

matbesancon:
The tests are freezing again here:
```
 Downloading artifact: GCCBootstrap-x86_64-apple-darwin14.v4.8.5.x86_64-linux-musl.unpacked
    Downloading [============================>            ]  67.5 %
```
I believe it was at the same point the last times, not sure if there is something special with the GCCBootstrap artifacts?

matbesancon:
The last tests to pass were
```
Test Summary:    | Pass  Total
Shard testsuites |   13     13
```
and then a bunch of downloads and the freeze mentioned above

giordano:
I was more wondering if you tried out the new functionality yourself.  It isn't exercised directly in the tests (you didn't add any test for it), and running locally the full test suite isn't a good idea (I did it once or twice, ever)

matbesancon:
> I was more wondering if you tried out the new functionality yourself

I tried it partially, I will re-run a complete package locally to see if it behaves as intended on the full wizard

matbesancon:
Me debugging some make errors is the new generative art network:
```
	make -j${nproc} highs
	make install libhighs
	make install
	make install highs
	vim Makefile 
	make package
	make install/fast
	make list
	make help
	make FortranHighs
	make install
	make help
	make libhighs
	make install help
	vim Makefile 
	make install/strip
	make install/strip fast
	make help
	make capi_unit_tests
	make install/local
	make install/fast
```

matbesancon:
I was two calls away from `make install/prettyplease`

giordano:
Even if you disable PkgServer with
```sh
export JULIA_PKG_SERVER=""
```
?

matbesancon:
Trying it. Will the wizard be able to complete the build without access to the Pkg server?

giordano:
Files would be downloaded directly from GitHub

matbesancon:
Even with `export JULIA_PKG_SERVER=""`, the wizard fails at the same point locally

giordano:
That's... surprising

matbesancon:
(on Julia 1.6)

matbesancon:
1.6 is the only way to make BB master run now right?

giordano:
If you're brave enough, you can also try Julia `master`.  Some stuff may not work though (in particular anything related with JLLs which are now standard libraries)

matbesancon:
I'm still getting these download freezes when downloading artifacts:
```
  Downloaded artifact: PlatformSupport-x86_64-linux-gnu.v2021.1.28.x86_64-linux-musl.unpacked
 Downloading artifact: GCCBootstrap-x86_64-linux-gnu.v4.8.5.x86_64-linux-musl.unpacked
    Downloading [============>                            ]  29.8 %
```

matbesancon:
an alternative would be to add a test to the wizard, but I don't see the final output being tested already, not sure how to approach it

giordano:
> I'm still getting these download freezes when downloading artifacts:

Should hopefully be fixed by https://github.com/JuliaLang/Downloads.jl/pull/102, which should make it for v1.6 (otherwise many downloads would freeze...)

PR #991, opened 2021-01-13
Add dependency on `Preferences` to JLLs

staticfloat:
Also bump generated JLLWrappers compat entry to `v1.3.0`

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1379,7 +1379,7 @@ function build_project_dict(name, version, dependencies::Array{Dependency}, juli
         "deps" => Dict{String,Any}(),
         # We require at least Julia 1.3+, for Pkg.Artifacts support, but we only claim
         # Julia 1.0+ by default so that empty JLLs can be installed on older versions.
-        "compat" => Dict{String,Any}("JLLWrappers" => "1.2.0",
+        "compat" => Dict{String,Any}("JLLWrappers" => "1.3.0",
                                      "julia" => "$(julia_compat)")
     )
     for dep in dependencies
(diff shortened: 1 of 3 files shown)


giordano:
This makes the packages fundamentally Julia v1.6-and-later-only

staticfloat:
Hmmm, that's a good point.  We can do the ol' "lie about compat" trick and have `Preferences.jl` say it can be installed on 1.0+, then have `JLLWrappers` only emit  `Preferences.jl` lookups on 1.6+.

src/AutoBuild.jl
```diff
@@ -1391,11 +1391,11 @@ function build_project_dict(name, version, dependencies::Array{Dependency}, juli
     end
     # Always add Libdl, Pkg and Artifacts as dependencies
     # Once we stop supporting Julia 1.5-, we can drop the `Pkg` requirement.
-    stdlibs = isdefined(Pkg.Types, :stdlib) ? Pkg.Types.stdlib : Pkg.Types.stdlibs
     project["deps"]["Libdl"] = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
     project["deps"]["Pkg"] = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
     project["deps"]["Artifacts"] = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
     project["deps"]["JLLWrappers"] = "692b3bcd-3c85-4b1f-b108-f13ce0eb3210"
+    project["deps"]["Preferences"] = "21216c6a-2e73-6563-6e65-726566657250"
```
giordano, line 1398: With https://github.com/JuliaPackaging/JLLWrappers.jl/pull/29, this will not be necessary anymore
```suggestion
```

test/jll.jl
```diff
@@ -12,12 +12,13 @@ module TestJLL end
     @test project["deps"] == Dict("JLLWrappers" => "692b3bcd-3c85-4b1f-b108-f13ce0eb3210",
                                   "Artifacts"   => "56f22d72-fd6d-98f1-02f0-08ddc0907c33",
                                   "Pkg"         => "44cfe95a-1eb2-52ea-b672-e2afdf69b78f",
+                                  "Preferences" => "21216c6a-2e73-6563-6e65-726566657250",
```
giordano, line 15: Like above, after https://github.com/JuliaPackaging/JLLWrappers.jl/pull/29 this isn't necessary
```suggestion
```

src/AutoBuild.jl
```diff
@@ -1379,7 +1379,7 @@ function build_project_dict(name, version, dependencies::Array{Dependency}, juli
         "deps" => Dict{String,Any}(),
         # We require at least Julia 1.3+, for Pkg.Artifacts support, but we only claim
         # Julia 1.0+ by default so that empty JLLs can be installed on older versions.
-        "compat" => Dict{String,Any}("JLLWrappers" => "1.2.0",
+        "compat" => Dict{String,Any}("JLLWrappers" => "1.3.0",
```
giordano, line 1382: Technically, I don't think we even need to bump the required version, as the API will be the same, which makes this pull request useless

PR #992, opened 2021-01-16
[Auditor] Add pass to make all binary files executable

giordano:
Related to https://github.com/JuliaLang/julia/issues/38993.  I verified locally
it works as I expect and will add tests later, after getting some feedback.

### DIFF
--- a/src/Auditor.jl
+++ b/src/Auditor.jl
@@ -85,6 +85,13 @@ function audit(prefix::Prefix, src_name::AbstractString = "";
                     all_ok &= check_isa(oh, platform, prefix; verbose=verbose, silent=silent)
                     # Check that the OS ABI is set correctly (often indicates the wrong linker was used)
                     all_ok &= check_os_abi(oh, platform, verbose = verbose)
+                    # Make sure all binary files are executables, if libraries aren't
+                    # executables Julia may not be able to dlopen them:
+                    # https://github.com/JuliaLang/julia/issues/38993.  In principle this
+                    # should be done when autofix=true, but we have to run this fix on MKL
+                    # for Windows, for which however we have to set autofix=false:
+                    # https://github.com/JuliaPackaging/Yggdrasil/pull/922.
+                    all_ok &= ensure_executability(oh; verbose=verbose, silent=silent)
 
                     # If this is a dynamic object, do the dynamic checks
                     if isdynamic(oh)
--- a/src/auditor/filesystems.jl
+++ b/src/auditor/filesystems.jl
@@ -43,3 +43,27 @@ function check_absolute_paths(prefix::Prefix, all_files::Vector; silent::Bool = 
 
     return true
 end
+
+function ensure_executability(oh::ObjectHandle; verbose::Bool=false, silent::Bool=false)
+    old_mode = filemode(path(oh))
+    # Execution permissions only for users who can read the file
+    read_mask = (old_mode & 0o444) >> 2
+    # Check whether the file has executable permission for all
+    if old_mode & read_mask != read_mask
+        if verbose
+            @info "Making $(path(oh)) executable"
+        end
+        try
+            # Add executable permission for all users that can read the file
+            chmod(path(oh), old_mode | read_mask)
+        catch e
+            if isa(e, InterruptException)
+                rethrow(e)
+            end
+            if !silent
+                @warn "$(path(oh)) could not be made executable!" exception=(e, catch_backtrace())
+            end
+        end
+    end
+    return true
+end
--- a/test/auditing.jl
+++ b/test/auditing.jl
@@ -517,6 +517,52 @@ end
     end
 end
 
+@testset "Auditor - execution permission" begin
+    mktempdir() do build_path
+        build_output_meta = nothing
+        product = LibraryProduct("libfoo", :libfoo)
+        @test_logs (:info, r"Making .*libfoo.* executable") match_mode=:any begin
+            build_output_meta = autobuild(
+                build_path,
+                "exec",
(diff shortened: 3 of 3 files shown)


src/Auditor.jl
```diff
@@ -85,6 +85,13 @@ function audit(prefix::Prefix, src_name::AbstractString = "";
                     all_ok &= check_isa(oh, platform, prefix; verbose=verbose, silent=silent)
                     # Check that the OS ABI is set correctly (often indicates the wrong linker was used)
                     all_ok &= check_os_abi(oh, platform, verbose = verbose)
+                    # Make sure all binary files are executables, if libraries aren't
+                    # executables Julia may not be able to dlopen them:
+                    # https://github.com/JuliaLang/julia/issues/38993.  In principle this
+                    # should be done when autofix=true, but we have to run this fix on MKL
+                    # for Windows, for which however we have to set autofix=false:
```
staticfloat, line 92: Another reason why we should not hide such complex functionality behind a simple `autofix` flag, and we should support enabling/disabling of each audit pass separately. :)

src/auditor/filesystems.jl
```diff
@@ -43,3 +43,22 @@ function check_absolute_paths(prefix::Prefix, all_files::Vector; silent::Bool =
 
     return true
 end
+
+function ensure_executability(oh::ObjectHandle; verbose::Bool=false, silent::Bool=false)
+    if filemode(path(oh)) & 0o755 != 0o755
```
staticfloat, line 48: Instead of checking for exactly `0o755`, I think we might instead want to check `filemode(path(oh)) & 0o111 == 0`, and then we probably want to set `+x` for any groups that have `+r`; so we want to set `new_mode = old_mode | (old_mode & 0o444 >> 2)`.

src/Auditor.jl
giordano: Agreed!

src/auditor/filesystems.jl
giordano: I knew I was doing it wrong, I was too tired yesterday night :grimacing: Thanks!

giordano: I'm not sure I understand why `old_mode | (old_mode & 0o444 >> 2)`:
```julia
julia> old_mode = 0o644
0x01a4

julia> string(old_mode | (old_mode & 0o444 >> 2) ; base=8)
"644"
```
doesn't look right.  I think it's just
```julia
julia> string(old_mode | 0o111; base=8)
"755"
```
no?

staticfloat: you need more parens:

```
julia> old_mode = 0o640
0x01a0

julia> string(old_mode | ((old_mode & 0o444) >> 2); base=8)
"750"
```

staticfloat: My bad.  :P

giordano: `750` still doesn't have exec bit for others... or am I missing something?  Do you want to require readability in order to add the exec permission?  `chmod +x` just adds `x` everywhere

giordano: As weird as it is, it doesn't look like a file need to be readable, in order to execute it:
```
% cat foo.c 
#include <stdio.h>

int main()
{
    printf("Hello World!\n");
    return 0;
}
% gcc -o foo foo.c
% chmod 100 foo   
% ./foo 
Hello World!
```

staticfloat: Yeah, that is weird.

That being said, what I want is for us to catch obvious mistakes (e.g. `mode & 0o111 == 0`) and to fix them without accidentally giving permissions to users that it wasn't intended for.  For instance, if we have SETUID binaries in an `/sbin` folder that accidentally didn't have permissions applied (very niche scenario, I know) we wouldn't want to turn `0o600` into `0o711`; we'd want to turn it into `0o700`.

giordano: All right, I applied the change suggested!

test/auditing.jl
```diff
@@ -517,6 +517,49 @@ end
...
+        tarball_path, tarball_hash = build_output_meta[platform][1:2]
+        @test isfile(tarball_path)
+
+        # Unpack it somewhere else
+        @test verify(tarball_path, tarball_hash)
+        testdir = joinpath(build_path, "testdir")
+        mkdir(testdir)
+        unpack(tarball_path, testdir)
+        libfoo_path = joinpath(testdir, build_output_meta[platform][4][product]["path"])
+        @test_broken filemode(libfoo_path) & 0o777 == 0o750
```
giordano, line 559: I believe we have an issue when packing the tarball: the file has permissions `0x750` as expected right after `ensure_executability`, but in the tarball it becomes `0o755`.  However this is orthogonal to this PR

giordano: It seems to be a deliberate decision in `Tar.jl`:

> files that are executable by the owner are archived with mode `0o755`;

I can reproduce this with
```julia
julia> using Tar

julia> mktempdir() do dir
           # Create the root of the tarball
           root = joinpath(dir, "root")
           mkpath(root)
           # Create the file and give permissions 750
           path = joinpath(root, "testfile")
           touch(path)
           chmod(path, 0o750)
           # Package the root
           tarball = Tar.create(root, joinpath(dir, "test.tar"))
           files = Tar.list(tarball)
           # Get the permissions of the file
           string(filter(f -> f.type === :file, files)[1].mode; base=8)
       end
"755"
```

staticfloat: Yeah; I think it’s good for BB to be correct, and Tar is fine to do it’s thing as well.

PR #1011, opened 2021-02-13
[Docs] More extensive description of different workflows

giordano:
In particular

* be slightly less enthusiastic and more realistic about the wizard
* encourage users to modify `build_tarballs.jl` for existing recipes instead
  of starting from scratch with the wizard
* move information about how to run the `build_tarballs.jl` script from the FAQ
  to the first page (frankly, I started using BB seriously when I learned this)
* briefly mention GitHub Codespaces
* update some doctests


docs/src/index.md
```diff
@@ -10,13 +10,44 @@ Suppose that you have a Julia package `Foo.jl` which wants to use a compiled `li
...
+```
+julia build_tarballs.jl --debug --verbose aarch64-linux-musl,arm-linux-musleabihf
+```
+you'll run the build script only for the `aarch64-linux-musl` and `arm-linux-musleabihf` target platforms.
+
+If you decide to use this workflow, however, you will need to manually open pull requests for [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil/).
+
+### GitHub Codespaces
+
+If you already have access to the [GitHub Codespaces](https://github.com/features/codespaces) service, you can use use BinaryBuilder and all the workflows described above in your browser or with Visual Studio Code, on any operating system, including those not natively supported by the package!  Head to [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil/) and create a new Codespace.
```
giordano, line 50: I don't know if we want to be more extensive here, for example with screenshots, but I feel like this page is already getting too long.

staticfloat (approved):

PR #1022, opened 2021-03-01
host vs target dependencies

bramtayl:
I thought this would be good to clarify because it took me a long time to figure it out (and I'm still learning)


giordano:
The content is useful and the table is nice, but I'm not convinced this is the right place where to have this information.  Another thing is that I'd like to start suggesting `HostBuildDependency` over installing packages with `apk`

bramtayl:
Yay! Sure, maybe there is a better place to put it. Maybe it would also be a good idea to be able to pick what kind of dependency you want in the wizard?

Issue #987, opened 2021-01-08
[CI] Test randomly failing


giordano:
We have somewhat frequently jobs failing on CI: https://dev.azure.com/JuliaPackaging/BinaryBuilder.jl/_build/results?buildId=9355&view=logs&j=2aca1793-ba52-57cd-8cb0-0d670b1a1d3e&t=917821bb-0400-5814-653b-f7fbfea4c593&s=d654deb9-056d-50a2-1717-90c08683d50a
```
Wizard - Downloading: Test Failed at /data/staticfloat/bb_azp/rootfs/agent_2/_work/1/s/test/wizard.jl:148
  Expression: getfield.(state.source_files, :hash) == [libfoo_tarball_hash]
   Evaluated: ["9354b78c0e5a9eadadd8d3250294766c5047d94b9e8e0a2387e6c728894078b7"] == ["e6374d589ce3a75aadd86f195a7ac27ce65bb43281806c20c75e8cd6cd158b3b"]
Stacktrace:
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1144
 [3] top-level scope
   @ /data/staticfloat/bb_azp/rootfs/agent_2/_work/1/s/test/wizard.jl:128
Wizard - Downloading: Test Failed at /data/staticfloat/bb_azp/rootfs/agent_2/_work/1/s/test/wizard.jl:174
  Expression: getfield.(state.source_files, :hash) == [libfoo_tarball_hash, libfoo_tarball_hash]
   Evaluated: ["9354b78c0e5a9eadadd8d3250294766c5047d94b9e8e0a2387e6c728894078b7", "9354b78c0e5a9eadadd8d3250294766c5047d94b9e8e0a2387e6c728894078b7"] == ["e6374d589ce3a75aadd86f195a7ac27ce65bb43281806c20c75e8cd6cd158b3b", "e6374d589ce3a75aadd86f195a7ac27ce65bb43281806c20c75e8cd6cd158b3b"]
Stacktrace:
 [1] top-level scope
   @ /data/staticfloat/bb_azp/rootfs/agent_2/_work/1/s/test/wizard.jl:174
 [2] top-level scope
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1144
 [3] top-level scope
[...]

PR #1023, opened 2021-03-03
Produce compat entries in Project.toml

fingolfin:
... using the optional meta data provided by users in BBB's Dependency type.

This is companion to PR https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/123, and part of a new effort to address https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/89. My previous attempt, https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/98, was not sufficient, as it failed to address the core issue: that while a compat string can be turned into a VersionSpec, the reverse is in general not possible; so specifying the compat bounds using a VersionSpec isn't too useful. The solution chosen here is to instead simply store it as a string (validated by calling `semver_spec` on it).

Unfortunately, this change is completely untested! The reason is that I am not sure at this point how I would adequately test it. Perhaps I can figure that out tomorrow, but I wanted to already open the PRs now to give people a chance to look at them.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -133,6 +133,18 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
     sources = coerce_source.(sources)
     dependencies = coerce_dependency.(dependencies)
 
+    # Reject user supplied dependencies using a VersionSpec: these should
+    # either use compat, or build_version, or both (depending on what they are
+    # trying to achieve). We cannot check for this in the Dependency
+    # constructor, as there are several valid situations in which we *do* want
+    # to store versions here (e.g. after running the dependency through the
+    # package resolver).
+    for dep in dependencies
+        if dep isa Dependency && dep.pkg.version != Pkg.Types.VersionSpec("*")
+            error("Dependency $(dep.pkg.name) specifies a version, use build_version and/or compat instead")
+        end
+    end
+
     # Do not clobber caller's ARGS
     ARGS = deepcopy(ARGS)
 
@@ -1369,24 +1381,6 @@ const uuid_package = UUID("cfb74b52-ec16-5bb7-a574-95d9e393895e")
 # "_jll" to the name of the new package before computing its UUID.
 jll_uuid(name) = bb_specific_uuid5(uuid_package, "$(name)_jll")
 function build_project_dict(name, version, dependencies::Array{Dependency}, julia_compat::String=DEFAULT_JULIA_VERSION_SPEC; lazy_artifacts::Bool=false, kwargs...)
-    function has_compat_info(d::Dependency)
-        r = Pkg.Types.VersionRange()
-        return isa(d.pkg.version, VersionNumber) ||
-               length(d.pkg.version.ranges) != 1 ||
-               d.pkg.version.ranges[1] != r
-    end
-    function exactly_this_version(v::VersionNumber)
-        return string("=", VersionNumber(v.major, v.minor, v.patch))
-    end
-    function exactly_this_version(v::Pkg.Types.VersionSpec)
-        if length(v.ranges) == 1 &&
-           v.ranges[1].lower.n == 3 &&
-           v.ranges[1].lower == v.ranges[1].upper
-           return string("=", v)
-       end
-       return string(v)
-    end
-    exactly_this_version(v) = v
     Pkg.Types.semver_spec(julia_compat) # verify julia_compat is valid
     project = Dict(
         "name" => "$(name)_jll",
@@ -1401,8 +1395,9 @@ function build_project_dict(name, version, dependencies::Array{Dependency}, juli
     for dep in dependencies
         depname = getname(dep)
         project["deps"][depname] = string(jll_uuid(depname))
-        if has_compat_info(dep)
-            project["compat"][depname] = string(exactly_this_version(dep.pkg.version))
+        if dep isa Dependency && length(dep.compat) > 0
+            Pkg.Types.semver_spec(dep.compat) # verify dep.compat is valid
+            project["compat"][depname] = dep.compat
         end
     end
     # Always add Libdl, Pkg and Artifacts as dependencies
--- a/test/basic.jl
+++ b/test/basic.jl
@@ -221,21 +221,21 @@ end
 
     # Ensure passing compat bounds works
     dependencies = [
-        Dependency(PackageSpec(name="libLLVM_jll", version=v"9")),
+        Dependency(PackageSpec(name="libLLVM_jll"), compat="=9.0.0"),
     ]
     dict = build_project_dict("Clang", v"9.0.1+2", dependencies)
     @test dict["compat"]["julia"] == "1.0"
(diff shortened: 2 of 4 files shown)


giordano:
Until https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/123 is merged, you may want to activate the package environment and do
```
]add https://github.com/fingolfin/BinaryBuilderBase.jl.git#mh/compat
```
This would make CI use the other PR

src/AutoBuild.jl
```diff
@@ -1076,7 +1076,7 @@ function build_jll_package(src_name::String,
     mkpath(joinpath(code_dir, "src", "wrappers"))
 
     # Drop BuildDependency objects
-    dependencies = Dependency[d for d in dependencies if is_runtime_dependency(d)]
+    dependencies = filter(is_runtime_dependency, dependencies)
```
giordano, line 1079: The reason why I did the comprehension with `Dependency[...]` is because
```
julia> dep = [Dependency("Foo_jll"), BuildDependency("Bar_jll")]
2-element Vector{AbstractDependency}:
 Dependency(PackageSpec(
  name = Foo_jll
  version = *
), nothing)
 BuildDependency(PackageSpec(
  name = Bar_jll
  version = *
))

julia> filter(is_runtime_dependency, dep)
1-element Vector{AbstractDependency}:
 Dependency(PackageSpec(
  name = Foo_jll
  version = *
), nothing)
```
but I wanted the output to be `Vector{Dependency}` (there may be tests making sure of this)

src/AutoBuild.jl
```diff
@@ -1401,8 +1383,9 @@ function build_project_dict(name, version, dependencies::Array{Dependency}, juli
     for dep in dependencies
         depname = getname(dep)
         project["deps"][depname] = string(jll_uuid(depname))
-        if has_compat_info(dep)
-            project["compat"][depname] = string(exactly_this_version(dep.pkg.version))
+        if dep isa Dependency && length(dep.compat) > 0
+            Pkg.Types.semver_spec(dep.compat) # verify dep.compat is valid
```
giordano, line 1387: Isn't this enforced in [the constructor of `Dependency`](https://github.com/fingolfin/BinaryBuilderBase.jl/blob/e2f1baefa710b5d3e44e61fa30a1259e34fd3f58/src/Dependencies.jl#L77)?  I guess double checking isn't a big deal though

staticfloat (approved):

src/AutoBuild.jl
fingolfin: Fixed, oops

src/AutoBuild.jl
fingolfin: Thing is, this data is produced in another package, then serialized, deserialized, then processed here, possibly even on another machine / VM. So I prefer to code defensively.

giordano: Fair enough

test/basic.jl
```diff
@@ -221,21 +221,21 @@ end
...
     @test dict["compat"]["libLLVM_jll"] == "=9.0.0"
 
     dependencies = [
-        Dependency(PackageSpec(name="libLLVM_jll", version="8.3-10")),
+        Dependency(PackageSpec(name="libLLVM_jll"), compat="8.3 - 10"),
     ]
     dict = build_project_dict("Clang", v"9.0.1+2", dependencies)
     @test dict["compat"]["julia"] == "1.0"
-    @test dict["compat"]["libLLVM_jll"] == "8.3-10"
+    @test dict["compat"]["libLLVM_jll"] == "8.3 - 10"
```
fingolfin, line 235: Note that the old compat string was invalid, and hence so was this test.

Also note that the *new* compat string only is supported in Julia >= 1.4 (IIRC), so I think my PRs should either outright reject it, or at least reject it if the `julia_compat` strings allows for older Julia versions. Thoughts?

fingolfin:
@giordano @staticfloat the test "Test Unprivileged_Unpacked" failed after running for 6 hours. Is that "normal", or is this something I need to investigate (but how?)

fingolfin:
I tested this successfully with a slightly tweaked `Arb_jll`, namely with this patch:
```patch
diff --git a/A/Arb/build_tarballs.jl b/A/Arb/build_tarballs.jl
index 9eb826bb..fbd8df82 100644
--- a/A/Arb/build_tarballs.jl
+++ b/A/Arb/build_tarballs.jl
@@ -61,7 +61,7 @@ products = [
 
 dependencies = [
-    Dependency(PackageSpec(name="FLINT_jll", version=VersionSpec("200.700"))),
+    Dependency(PackageSpec(name="FLINT_jll"), compat = "~200.700"),
     Dependency("GMP_jll", v"6.1.2"),
     Dependency("MPFR_jll", v"4.0.2"),
 ]
```
I then invoked BB (pointed at the two branches) like this:
```
julia-1.6.0-rc1 --color=yes --project=proj-compat ....../Yggdrasil/A/Arb/build_tarballs.jl --verbose --deploy=local
```
This produced this `~/.julia/dev/Arb_jll/Project.toml`:
```toml
name = "Arb_jll"
uuid = "d9960996-1013-53c9-9ba4-74a4155039c3"
version = "200.1900.0+1"

[deps]
JLLWrappers = "692b3bcd-3c85-4b1f-b108-f13ce0eb3210"
Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
GMP_jll = "781609d7-10c4-51f6-84f2-b8444358ff6d"
MPFR_jll = "3a97d323-0669-5f0c-9066-3539efd106a3"
Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
FLINT_jll = "e134572f-a0d5-539d-bddf-3cad8db41a82"

[compat]
JLLWrappers = "1.2.0"
julia = "1.0"
FLINT_jll = "~200.700"
```

staticfloat (approved):

giordano:
> the test "Test Unprivileged_Unpacked" failed after running for 6 hours. Is that "normal", or is this something I need to investigate (but how?)

Random failures described in #987 should be fixed by #1026 (and also worked around by #1025).  Tests may time out if there is an error inside a wizard test wrapped in a `try`/`catch` block.  Most of the errors I've seen are related to failures to download the sources of the wizard tests.  Perhaps these will also be fixed by #1026.

fingolfin:
So how do we proceed with this? I run all tests I could think of and also executed parts of the Yggdrasil scripts manually as suggested by @giordano. What else to do?

Of course the conflict in the Manifest needs to be resolved (once the BBB PR has been merged). perhaps the version of BBB increased to indicate a breaking change following semcver rules. What else?

giordano:
> perhaps the version of BBB increased to indicate a breaking change following semcver rules

We haven't released a new version of BBB in months :wink:

fingolfin:
Ok, but that doesn't really answer my question...

PR #1025, opened 2021-03-06
Use `Tar.jl` to create tarball to use for Wizard tests

giordano:
In
https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/937c7af962af6745809eb3edaca201b88f1ddc3e/test/wizard.jl#L48-L57
we create a tarball, compute its hash and serve it via HTTP, for use during the tests.  The wizard functions in the tests would download this file, compute the hash and compare with the hash of the original file.  The problem is that we're often observing that the hash of the original file is different from that of the downloaded files, causing a failure in the tests.

Switching to `Tar.jl` to create the tarballs works around #987.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -28,7 +28,7 @@ uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
 [[BinaryBuilderBase]]
 deps = ["CodecZlib", "Downloads", "InteractiveUtils", "JSON", "LibGit2", "Libdl", "Logging", "OutputCollectors", "Pkg", "Random", "SHA", "Scratch", "SimpleBufferStream", "TOML", "Tar", "UUIDs", "p7zip_jll", "pigz_jll"]
-git-tree-sha1 = "557b497af843c3f48fe4678c4b47b13db81805aa"
+git-tree-sha1 = "87236ea2e2cf83b4a5bbeb29bc76c94f4a3252d4"
 repo-rev = "master"
 repo-url = "https://github.com/JuliaPackaging/BinaryBuilderBase.jl.git"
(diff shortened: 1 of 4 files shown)


staticfloat:
I don't see how this can possibly effect anything.  We're just serving the same bits over and over, we're not calculating it again and again, right?  And we only compare against the hash that we calculate after calling `tar`?

giordano:
> I don't see how this can possibly effect anything.

Me neither :slightly_smiling_face: 

> We're just serving the same bits over and over, we're not calculating it again and again, right? And we only compare against the hash that we calculate after calling `tar`?

That's all correct.  And in my multiple debug attempts I also verified Julia computes the right hash (compared with `sha256sum`).  We know that 

As I mentioned in Slack, the only advantage of `Tar.jl` is that creates reproducible tarballs (because the timestamps are all the same), which makes it easier to check where something is breaking (we know what's the expected hash, as we can run the same code also locally).  I wanted to use `Tar` for easier debugging but after running the tests 16 times I got no failures (compared to 1 in 3, on average), soooo :shrug:

giordano:
For the record, we believe that the issue was that we run the tests in parallel on the same machine.  Each job would try to set up an HTTP server on the same port, but only the first one would succeed (we don't see that the others fail because those are `@async` tasks) and serve the files to all the other jobs, even if the files they wanted to serve were different.  This issue is addressed by #1026, which properly fixes #987.

Switching to `Tar.jl`, as done in this pull request, works around #987 because `Tar.jl` generates reproducible tarballs normalising their content (e.g., all timestamps are reduced to the Unix epoch): in the end it doesn't matter which job succeeds to set up the server, they'd all serve the same content.

PR #1036, opened 2021-04-18
Add platform argument to cleanup_dependencies

barche:
Use the extra `platform` argument introduced in PR https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/135

This will of course fail until that PR is merged.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -33,7 +33,7 @@ ghr_jll = "07c12ed4-43bc-5495-8a2a-d5838ef8d533"
 
 [compat]
 ArgParse = "1.1"
-BinaryBuilderBase = "0.5, 0.6"
+BinaryBuilderBase = "0.6.1"
 GitHub = "5.1"
 HTTP = "0.8, 0.9"
 JLD2 = "0.1.6, 0.2, 0.3, 0.4"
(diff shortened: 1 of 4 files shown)


giordano:
After https://github.com/JuliaRegistries/General/pull/34640/ is merged, require BinaryBuilderBase v0.6.0 in `Project.toml`, and `]update` the `Manifest.toml`

giordano:
For the record, I initially went with passing a `platform` argument to `cleanup_dependencies`, but then changed my mind because there were some places in the wizard where the concrete platforms wasn't known: https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/995#issuecomment-767214109

Project.toml
```diff
@@ -33,7 +33,7 @@ ghr_jll = "07c12ed4-43bc-5495-8a2a-d5838ef8d533"
 
 [compat]
 ArgParse = "1.1"
-BinaryBuilderBase = "0.5, 0.6"
+BinaryBuilderBase = "0.6"
```
giordano, line 36:
```suggestion
BinaryBuilderBase = "0.6.1"
```

giordano: Need also to update the manifest 😁

Issue #1033, opened 2021-04-08
Improve experience when `GITHUB_TOKEN` is invalid


omus:
I was deploying a BB build and I saw this error:
```
ERROR: LoadError: Error found in GitHub reponse:
	Status Code: 401
	Message: Bad credentials
	Docs URL: https://docs.github.com/rest
	Errors:
```
I've used BB in the past normally I would get a prompt telling me to enter a token in a website to authenticate so this message confused me.

I ended up looking into the issue and isolated it to:
```julia
julia> BinaryBuilder.Wizard.github_auth(;allow_anonymous=false)
ERROR: Error found in GitHub reponse:
	Status Code: 401
	Message: Bad credentials
	Docs URL: https://docs.github.com/rest
	Errors:
Stacktrace:
  [1] error(::String, ::String, ::String, ::String, ::String)
    @ Base ./error.jl:42
  [2] handle_response_error(r::HTTP.Messages.Response)
    @ GitHub ~/.julia/packages/GitHub/Em3bH/src/utils/requests.jl:177
  [3] github_request(api::GitHub.GitHubWebAPI, request_method::Function, endpoint::String; auth::GitHub.OAuth2, handle_error::Bool, headers::Dict{Any, Any}, params::Dict{Any, Any}, allowredirects::Bool)
    @ GitHub ~/.julia/packages/GitHub/Em3bH/src/utils/requests.jl:68
  [4] #gh_get#4
    @ ~/.julia/packages/GitHub/Em3bH/src/utils/requests.jl:72 [inlined]
  [5] #authenticate#28
    @ ~/.julia/packages/GitHub/Em3bH/src/utils/auth.jl:56 [inlined]
  [6] authenticate
    @ ~/.julia/packages/GitHub/Em3bH/src/utils/auth.jl:55 [inlined]
  [7] #authenticate#29
    @ ./none:0 [inlined]
  [8] authenticate
    @ ./none:0 [inlined]
  [9] github_auth(; allow_anonymous::Bool)
    @ BinaryBuilder.Wizard ~/.julia/packages/BinaryBuilder/HOnlf/src/wizard/github.jl:11
 [10] top-level scope
    @ REPL[3]:1
```
Eventually I realized I had set `GITHUB_TOKEN` in my startup.jl and I also rotated my PATs meaning the token stored there was no longer valid. I did briefly think the problem could be this but as `env` didn't show `GITHUB_TOKEN` being set I thought the issue was something else.

[...]

PR #1041, opened 2021-04-21
Catch error from stale GitHub token

Wimmerer:
This seems like the best way to catch the error. Should I be checking if it's actually an auth error?

### DIFF
--- a/src/wizard/github.jl
+++ b/src/wizard/github.jl
@@ -8,7 +8,12 @@ function github_auth(;allow_anonymous::Bool=true)
     if !isassigned(_github_auth) || !allow_anonymous && isa(_github_auth[], GitHub.AnonymousAuth)
         # If the user is feeding us a GITHUB_TOKEN token, use it!
         if length(get(ENV, "GITHUB_TOKEN", "")) == 40
-            _github_auth[] = GitHub.authenticate(ENV["GITHUB_TOKEN"])
+            try
+                _github_auth[] = GitHub.authenticate(ENV["GITHUB_TOKEN"])
+            catch e
+                    println(stdout, "The GitHub Token found in \"GITHUB_TOKEN\" has likely expired. We will now attempt to obtain a fresh token for this session.")
+                    _github_auth[] = GitHub.authenticate(obtain_token())
+            end
         else
             if allow_anonymous
                 _github_auth[] = GitHub.AnonymousAuth()
--- a/src/wizard/github.jl
+++ b/src/wizard/github.jl
@@ -5,32 +5,21 @@ using Registrator
 const _github_auth = Ref{GitHub.Authorization}()
 
 function github_auth(;allow_anonymous::Bool=true)
-    if !isassigned(_github_auth) || !allow_anonymous && isa(_github_auth[], GitHub.AnonymousAuth)
-        # If the user is feeding us a GITHUB_TOKEN token, use it!
-        if length(get(ENV, "GITHUB_TOKEN", "")) == 40
-            try 
-                _github_auth[] = GitHub.authenticate(ENV["GITHUB_TOKEN"])
-            catch e
-                    println(stdout, "The GitHub token found in \"GITHUB_TOKEN\" is incorrect or has expired.")
-                    if yn_prompt("Would you like to attempt to obtain a new token from GitHub?") == :y
-                        _github_auth[] = GitHub.authenticate(obtain_token())
-                    else
-                        error("No GitHub token available. Please provide one in the \"GITHUB_TOKEN\" environment variable, or reply yes to the previous prompt.")
-                    end
-            end
-        else
-            if allow_anonymous
-                _github_auth[] = GitHub.AnonymousAuth()
-            else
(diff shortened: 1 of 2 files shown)


Wimmerer:
Hopefully resolves #1033

giordano:
I wonder whether this function should rethrow the error but prompt a more informative message, suggesting to use `obtain_key`.  CC: @omus

staticfloat:
Yeah, I'm not sure that `obtain_token()` is the right thing to automatically try here, since it will create a persistent token in your GitHub profile I think.  Probably best to only do that when explicitly requested.

Wimmerer:
I don't beleive it creates a permanent token. Or at least it doesn't against my account. There's a println in line 102 of github.jl that indicates it is only a termporary token used for the session. Am I missing something?﻿

Wimmerer:
Oh I found it. It's not in personal access tokens (duh), it's in Authorized Apps->Authorized OAuth Apps. I can change it to a prompt, which will error out with a friendly message if the answer is no, if the answer is yes proceed with `obtain_token`? The PR matches what we do if there is no token already, rather than a stale token. Should I prompt for both?

src/wizard/github.jl
```diff
@@ -8,14 +8,27 @@ function github_auth(;allow_anonymous::Bool=true)
...
                 # If we're not allowed to return an anonymous authorization,
                 # then let's create a token.
-                _github_auth[] = GitHub.authenticate(obtain_token())
+                if yn_prompt("Would you like to attempt to obtain a new token from GitHub?") == :y
+                    _github_auth[] = GitHub.authenticate(obtain_token())
+                else
+                    error("No GitHub token available. Please provide one in the \"GITHUB_TOKEN\" environment variable, or reply yes to the previous prompt.")
+                end
             end
         end
```
omus, line 33: How about doing the following to reduce code reuse:
```julia
if (!isassigned(_github_auth) || !allow_anonymous && isa(_github_auth[], GitHub.AnonymousAuth)) && length(get(ENV, "GITHUB_TOKEN", "")) == 40
    try 
        _github_auth[] = GitHub.authenticate(ENV["GITHUB_TOKEN"])
    catch e
        e isa ... || rethrow()
    end
end

if !isassigned(_github_auth) && allow_anonymous
    _github_auth[] = GitHub.AnonymousAuth()
end

if !isassigned(_github_auth)
    _github_auth[] = GitHub.authenticate(obtain_token())
end
```

src/wizard/github.jl
```diff
@@ -8,14 +8,27 @@ function github_auth(;allow_anonymous::Bool=true)
...
+                    end
+            end
         else
             if allow_anonymous
                 _github_auth[] = GitHub.AnonymousAuth()
             else
                 # If we're not allowed to return an anonymous authorization,
                 # then let's create a token.
-                _github_auth[] = GitHub.authenticate(obtain_token())
+                if yn_prompt("Would you like to attempt to obtain a new token from GitHub?") == :y
```
omus, line 27: As `obtain_token()` already requires user interaction to proceed I don't really see the point of adding an additional prompt.

Wimmerer: @staticfloat are you ok with no additional prompt? I can revert back to no extra prompt if so. This would be the original behavior with the GitHub prompt but no extra "would you like to proceed" step.

staticfloat: Yeah, @omus knows this part better than I do; let's do what he suggests.

Wimmerer:
I chose `@warn` to inform the user that the token may be stale and that we'll request to obtain a temporary token.  Also checked to be sure this is a 401 error, otherwise rethrow.

src/wizard/github.jl
```diff
@@ -5,20 +5,26 @@ using Registrator
...
+
+    if (!isassigned(_github_auth) || !allow_anonymous && isa(_github_auth[], GitHub.AnonymousAuth)) && length(get(ENV, "GITHUB_TOKEN", "")) == 40
+        try 
             _github_auth[] = GitHub.authenticate(ENV["GITHUB_TOKEN"])
-        else
-            if allow_anonymous
-                _github_auth[] = GitHub.AnonymousAuth()
+        catch e
+            if occursin("401", e.msg)
+                @warn "GitHub was unable to authenticate using the token from ENV[\"GITHUB_TOKEN\"], it may be stale. We will now request a temporary token."
```
omus, line 14: Possibly anonymous auth may be attempted
```suggestion
                @warn "GitHub was unable to authenticate using the token from `ENV[\"GITHUB_TOKEN\"]`, it may be stale. Falling back on alternate authentication methods."
```

src/wizard/utils.jl
```diff
@@ -200,7 +200,6 @@ function yn_prompt(state::WizardState, question::AbstractString, default = :y)
         end
     end
 end
-
 """
```
omus, line 203:
```suggestion

"""
```

omus (approved):

omus:
@staticfloat anything else to be done here?

staticfloat:
Thanks @Wimmerer and @omus! :)

PR #1044, opened 2021-04-29
Add `* commit: mycommithash123` to the PR body when making PRs to the General registry

DilumAluthge:
This will be required for the upcoming breaking release of RegistryCI.jl.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -479,6 +479,7 @@ function register_jll(name, build_version, dependencies, julia_compat;
     end
     # Calculate tree hash of wrapper code
     wrapper_tree_hash = bytes2hex(Pkg.GitTools.tree_hash(code_dir))
+    wrapper_commit_hash = strip(read(`git -C $(code_dir) rev-parse HEAD`, String))
 
     # Use RegistryTools to push up a new `General` branch with this JLL package registered within it
     # TODO: Update our fork periodically from upstream `General`.
@@ -513,6 +514,7 @@ function register_jll(name, build_version, dependencies, julia_compat;
         * Registering JLL package $(basename(deploy_repo))
         * Repository: https://github.com/$(deploy_repo)
         * Version: v$(build_version)
+        * Commit: $(wrapper_commit_hash)
         """
         if is_yggdrasil()
             body *= """
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -479,7 +479,7 @@ function register_jll(name, build_version, dependencies, julia_compat;
     end
     # Calculate tree hash of wrapper code
     wrapper_tree_hash = bytes2hex(Pkg.GitTools.tree_hash(code_dir))
-    wrapper_commit_hash = strip(read(`git -C $(code_dir) rev-parse HEAD`, String))
+    wrapper_commit_hash = LibGit2.head(code_dir)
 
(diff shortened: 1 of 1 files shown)


src/AutoBuild.jl
```diff
@@ -479,6 +479,7 @@ function register_jll(name, build_version, dependencies, julia_compat;
     end
     # Calculate tree hash of wrapper code
     wrapper_tree_hash = bytes2hex(Pkg.GitTools.tree_hash(code_dir))
+    wrapper_commit_hash = strip(read(`git -C $(code_dir) rev-parse HEAD`, String))
```
giordano, line 482:
```suggestion
    wrapper_commit_hash = readchomp(`git -C $(code_dir) rev-parse HEAD`)
```
is simpler, but why not using LibGit2 as we do everywhere else here?

DilumAluthge: Done, take a look.

DilumAluthge:
@giordano @staticfloat Any idea what is causing these test failures? They don't seem to be related to this PR.

staticfloat:
@giordano somehow we lost the `os_version` in the triplets being generated by the `rustc` wrappers, I think.

giordano:
> @giordano somehow we lost the os_version in the triplets being generated by the rustc wrappers, I think.

You mean https://github.com/JuliaPackaging/Yggdrasil/pull/2148#issuecomment-732477268 (and https://github.com/JuliaPackaging/Yggdrasil/pull/2168 and https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/80)?  What I don't understand is why now :confused:

giordano:
>  What I don't understand is why now confused

I believe this was broken by https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/43 which moved some directories around, in particular the binaries directories that were previously in `/opt/bin`.  But that was merged 3 months ago, I still don't understand why it didn't pop up so far.

Anyway, fix in https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/141

DilumAluthge:
Still failing :(

giordano:
If you don't change the manifest nothing changes.  But

1. I updated the manifest in #1046
2. there is a different error on FreeBSD (an object file, `crt1.o`, doesn't seem to have been built with `-fPIC`)

DilumAluthge:
Ah you're using a manifest.

BTW I recently made an automatic Manifest Updater action for the General registry (https://github.com/JuliaRegistries/General/blob/master/.github/workflows/update_manifests.yml) that you could use here if you wanted.

It supports having multiple different manifests, one for each minor version of Julia.

giordano (approved):

DilumAluthge:
Btw, it would be nice to get this "deployed" in the sense that the `@jlbuild` user account will start using this version of the code.

DilumAluthge:
Is there a way for me to check which version of BinaryBuilder.jl is being run by the `jlbuild` bot? I'm going to release RegistryCI 7.0.0 soon and deploy it on the General registry, and this will require the feature from this PR.

DilumAluthge:
From looking at https://github.com/JuliaPackaging/BinaryBuilder.jl/commit/5f2614b434966ca0762ea183ce5a892afc1650e1, it looks like the first version of BinaryBuilder.jl that has this PR in it is `v0.3.3`.

PR #1046, opened 2021-04-30
Use `github_release` instead of `ghr` to upload tarballs

giordano:


giordano:
Ok, not really necessary anymore as `ghr` is creating again tags with the right name

PR #1048, opened 2021-05-01
A simpler way of disabling Codecov commit statuses

DilumAluthge:
Based on https://docs.codecov.io/docs/commit-status#disabling-a-status

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,11 +2,5 @@ comment: false
 
 coverage:
   status:
-    project:
-      default:
-        threshold: 1%
-        target: 1%
-    patch:
-      default:
-        threshold: 1%
-        target: 1%
+    project: off
+    patch: off


giordano (commented):
I don't think the goal was to completely disable codecov commit statuses (at least that wasn't my intention and I added those lines: #550) but rather have lower thresholds

DilumAluthge:
Is a threshold of 1% useful in practice? Would it make sense to pick slightly higher thresholds?

giordano:
The problem is more that I don't really understand how the thresholds work (allowing something like ~ -5% would be ok, but still good to report larger drops) and settings of codecov have always been very confusing to me.

DilumAluthge:
Okay I see what you want to do. Let me make a new PR.

DilumAluthge:
My earlier comment was mistaken. The problem is not the threshold, it's the target. You basically never want to set the target to something as low as 1%. See https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1049

PR #1049, opened 2021-05-02
Improve the Codecov configuration

DilumAluthge:
Okay, I think this accomplishes what you want. Here's an explanation of what it does.

There will be three Codecov statuses:
- `codecov/minimum`
- `codecov/project`
- `codecov/patch`

## `codecov/minimum`

If, on the PR merge commit, the total coverage of the entire BinaryBuilder.jl package is greater than or equal to `50%`, this status will pass. Otherwise, it will fail.

## `codecov/project`

Let `M` denote the total coverage of the entire BinaryBuilder.jl package on the `master` branch. For example, here is the current value of `M`:

[codecov.io](http://codecov.io/github/JuliaPackaging/BinaryBuilder.jl?branch=master)

If, on the PR merge commit, the total coverage of the entire BinaryBuilder package is greater than or equal to `M - 3%`, this status will pass. Otherwise, it will fail.

## `codecov/patch`

Let `M` denote the total coverage of the entire BinaryBuilder.jl package on the `master` branch.

If, on the PR merge commit, the coverage on **only the lines changed by the PR** is greater than or equal to `M - 3%`, this status will pass. Otherwise, it will fail.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,9 +4,11 @@ coverage:
   status:
     project:
       default:
-        threshold: 1%
-        target: 1%
+        threshold: 3%
+        target: auto
+      minimum:
+        target: "50%"
     patch:
       default:
-        threshold: 1%
-        target: 1%
+        threshold: 3%
+        target: auto
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,11 +4,11 @@ coverage:
   status:
     project:
       default:
-        threshold: 3%
-        target: auto
+        threshold: "3%"
+        target: "auto"
       minimum:
         target: "50%"
     patch:
       default:
-        threshold: 3%
-        target: auto
+        threshold: "3%"
+        target: "auto"


DilumAluthge:
(We can remove the `codecov/minimum` status if you like. I just put it in there as a suggestion. I find it useful for enforcing an "absolute minimum" level of coverage below which we never want our package to fall.

DilumAluthge:
Also, we can increase or decrease those `3%` thresholds if you like.

And we can have different threshold values for `codecov/project` vs `codecov/patch`.

giordano:
Am I right to say that the threshold was set correctly but what was wrong was the target?  With `auto` the reference coverage is the current one, instead at the moment we have a useless `1%`?

DilumAluthge:
> Am I right to say that the threshold was set correctly but what was wrong was the target? With `auto` the reference coverage is the current one, instead at the moment we have a useless `1%`?

Yeah that's correct.

giordano (approved):
Ok, using target = auto is what I wanted to do, thanks!

giordano:
image

I'm not entirely sure this worked as expected, but this confirms Codecov settings are just meaningless

PR #1057, opened 2021-05-12
Ping the Yggdrasil committer when deploying

### LINKED ISSUE #1056: Ping author on PR to Yggdrasil
In case registration fails, the author should be aware of the progress through registration.

https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/801f58c3f7e0091e2a52cb7b14f9dbbc0fb41a5d/src/AutoBuild.jl#L519-L523

Pangoraw:
Hi!

This adds a mention to the Yggdrasil committer when deploying creating PRs to the registry.
This fails silently because some commits can have no github author if the email address used for the commit doesn't match any GitHub user.

Closes #1056

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -517,9 +517,16 @@ function register_jll(name, build_version, dependencies, julia_compat;
         * Commit: $(wrapper_commit_hash)
         """
         if is_yggdrasil()
+            commit_hash = yggdrasil_head()
             body *= """
-                    * Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/$(yggdrasil_head())
+                    * Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/$commit_hash
                     """
+            commit_author_login = get_github_author_login("JuliaPackaging/Yggdrasil", commit_hash; gh_auth=gh_auth)
+            if commit_author_login !== nothing
+                body *= """
+                        * Created by: @$commit_author_login
+                        """
+            end
         end
         params = Dict(
             "base" => "master",
@@ -883,6 +890,15 @@ function download_github_release(download_dir, repo, tag; gh_auth=Wizard.github_
         download(asset["browser_download_url"], joinpath(download_dir, asset["name"]))
     end
     return assets
+end
+
+function get_github_author_login(repository, commit_hash; gh_auth=Wizard.github_auth())
+    try
+        commit = GitHub.commit(repository, commit_hash; auth=gh_auth)
+        commit.author.login
+    catch
+        nothing
+    end
 end
 
 
--- a/test/jll.jl
+++ b/test/jll.jl
@@ -1,6 +1,7 @@
 using JSON
 using UUIDs
-using BinaryBuilder: jll_uuid, build_project_dict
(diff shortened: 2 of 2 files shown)


src/wizard/github.jl
```diff
@@ -115,6 +115,15 @@ function obtain_token(; outs=stdout, github_api=GitHub.DEFAULT_API)
...
 
+function find_yggdrasil_commit_author_login(commit_hash; auth=github_auth(; allow_anonymous=true))
+  try
+    commit = GitHub.commit("JuliaPackaging/Yggdrasil", commit_hash; auth=auth)
+    commit.author.login
+  catch
+    nothing
+  end
+end
+
```
giordano, line 126: I don't think this function belongs to this module, it isn't used anywhere here.  The function can also take the repository in input.  In the future we may want to make it easy to use something different from hard-coded `JuliaPackaging/Yggdrasil`.  Also, please use 4 spaces for indentation, like the rest of the package.

src/AutoBuild.jl
```diff
@@ -517,9 +517,16 @@ function register_jll(name, build_version, dependencies, julia_compat;
         * Commit: $(wrapper_commit_hash)
         """
         if is_yggdrasil()
+            commit_hash = yggdrasil_head()
             body *= """
-                    * Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/$(yggdrasil_head())
+                    * Revision on Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/commit/$commit_hash
                     """
+            commit_author_login = Wizard.find_yggdrasil_commit_author_login(commit_hash; auth=gh_auth)
+            if commit_author_login != nothing
```
giordano, line 525:
```suggestion
            if commit_author_login !== nothing
```

src/wizard/github.jl
Pangoraw: I moved it close to `download_github_release()` in `src/AutoBuild.jl`

Pangoraw:
The code coverage is complaining, but I couldn't find tests for `register_jll(...)`. Am I missing something or should I add a test only for the login function ?

giordano:
Yeah, the registration function isn't tested at all because we'd need a fake registry to push to.  We should really have it because registration gets broken from time to time.

To test `get_github_author_login` you can do something like this:
```julia
julia> gh_auth = GitHub.AnonymousAuth()
GitHub.AnonymousAuth()

julia> @test get_github_author_login("JuliaPackaging/Yggdrasil", "gggggggg"; gh_auth) === nothing
Test Passed

julia> gh_auth = Wizard.github_auth();

julia> if gh_auth != GitHub.AnonymousAuth()
           @test get_github_author_login("JuliaPackaging/Yggdrasil", "dea7c3fadad16281ead2427f7ab9b32f1c8cb664") == "Pangoraw"
       end
Test Passed
```
and let's hope you won't change your GitHub name :wink:

Pangoraw:
Ok!

I added a test for a commit without author: https://github.com/juliapackaging/yggdrasil/commit/815de56a4440f4e05333c5295d74f1dc9b73ebe3
The tests can be changed when a test registry is created

giordano:
Looks good now, thank you so much!

Issue #1082, opened 2021-07-20
More Rust tests failing


giordano:
This is different from #1061.  In CI I'm [seeing](https://dev.azure.com/JuliaPackaging/BinaryBuilder.jl/_build/results?buildId=12369&view=logs&j=2aca1793-ba52-57cd-8cb0-0d670b1a1d3e&t=917821bb-0400-5814-653b-f7fbfea4c593) Rust tests for aarch64-linux-musl and i686-w64-mingw32 failing:
```
..rustcgo -o /tmp/testsuite/aarch64-linux-musl/rust/hello_world/hello_world hello_world.rs 
build -o /tmp/testsuite/aarch64-linux-musl/go/hello_world/hello_world hello_world.go
cargo install --verbose --path . --root /tmp/testsuite/aarch64-linux-musl/rust/cargo_build
cc -o /tmp/testsuite/aarch64-linux-musl/c/hello_world/hello_world -g -O2 hello_world.c
          /build/musl-1.1.22/src/stdio/vfprintf.c:246: undefined reference to `__subtf3'
          /build/musl-1.1.22/src/stdio/vfprintf.c:246: undefined reference to `__multf3'
          /build/musl-1.1.22/src/stdio/vfprintf.c:268: undefined reference to `__multf3'
          /build/musl-1.1.22/src/stdio/vfprintf.c:275: undefined reference to `__subtf3'
          /build/musl-1.1.22/src/stdio/vfprintf.c:275: undefined reference to `__multf3'
          /build/musl-1.1.22/src/stdio/vfprintf.c:331: undefined reference to `__addtf3'
          /opt/x86_64-linux-musl/toolchains/1.43.0-x86_64-unknown-linux-musl/lib/rustlib/aarch64-unknown-linux-musl/lib/liblibc-7c3c4e3d691d5338.rlib(frexpl.lo): In function `frexpl':
          /build/musl-1.1.22/src/math/frexpl.c:16: undefined reference to `__multf3'
          collect2: error: ld returned 1 exit status
          

error: aborting due to previous error

error: failed to compile `cargo_build v0.1.0 (/usr/share/testsuite/rust/cargo_build)`, intermediate artifacts can be found at `/usr/share/testsuite/rust/cargo_build/target`

Caused by:
  could not compile `cargo_build`.

Caused by:
[...]

PR #1084, opened 2021-08-11
Update to BinaryBuilderBase 1.0

ararslan:
This allows both BinaryBuilderBase 0.6 and 1.0 but updates the manifest to use 1.0. Allowing both means we should be able to make a non-breaking release. I've changed the Project.toml version accordingly in a separate commit.


giordano (approved):
Yes, I agree allowing both (API didn't really change) and bumping only patch version number is fine

giordano:
Tests failure in the Rust toolchain are #1082, which are unrelated and predate this change.  Unfortunately I have little clue of what's going on there and can't reproduce the failure locally.  However I suspected something did happen between BinaryBuilderBase 0.6.11 (current version in `Manifest.toml`) and 1.0.0, but no idea of what specifically could cause those errors

PR #1087, opened 2021-08-21
add newline at end of print_build_tarballs string body

Crghilardi:
My workflow with BB is typically `run_wizard()` -> write to local file -> copy file and submit PR manually.

When submitting files directly from the wizard output, the addition of a newline at the end has been noted in review a few times so it seemed reasonable to patch here. I don't think this is an issue with the auto-Yggdrasil PR option or the other option?


giordano:
There is already a newline at the end, no? :thinking:

Crghilardi:
I have always been copy/pasting from the default text editor (xed?) on Linux mint which apparently doesn't show blank lines at EOF. Looks like there is even an open issue on the xed repo from ~2016...

Everything looks as expected when viewing files in VScode or Sublime.

PR #1119, opened 2021-11-30
In `bb add`, unlink artifacts, but move rm to BBBase

Keno:
BBBase changed the layout of the $prefix/artifacts directory, such that
the fix in #931 was no longer effective. Instead of closely coupling
BB to BBBase just move the rm part to BBBase[1] and do a proper
cleanup/setup cycle here.

[1] https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/183


giordano:
Looks great, thanks a lot!

PR #1123, opened 2021-12-08
[Docs] Add installation tips about Rust and Go

giordano:
We may eventually want to add Go and Rust hints to the wizard, but at the moment we don't support those toolchains at all in that environment.

### DIFF
--- a/docs/src/build_tips.md
+++ b/docs/src/build_tips.md
@@ -83,6 +83,40 @@ Examples of builds performed with Meson include:
   this script modifies `c_args` in the Meson cross file in order to add an include directory;
 * [xkbcommon](https://github.com/JuliaPackaging/Yggdrasil/blob/2f3638292c99fa6032634517f8a1aa8360d6fe8d/X/xkbcommon/build_tarballs.jl#L26-L30).
 
+## Go builds
+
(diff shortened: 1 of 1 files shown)


giordano:
> but at the moment we don't support those toolchains at all in that environment.

Wait, we do!  I did that: #553 :facepalm:

docs/src/build_tips.md
```diff
@@ -83,6 +83,40 @@ Examples of builds performed with Meson include:
   this script modifies `c_args` in the Meson cross file in order to add an include directory;
 * [xkbcommon](https://github.com/JuliaPackaging/Yggdrasil/blob/2f3638292c99fa6032634517f8a1aa8360d6fe8d/X/xkbcommon/build_tarballs.jl#L26-L30).
 
+## Go builds
+
+The Rust toolchain provided by BinaryBuilder can be requested by adding `:go` to the `compilers` keyword argument to [`build_tarballs`](@ref): `compilers=[:c, :go]`.  Go-based packages can usually be built and installed with `go`:
```
fredrikekre, line 88:
```suggestion
The Go toolchain provided by BinaryBuilder can be requested by adding `:go` to the `compilers` keyword argument to [`build_tarballs`](@ref): `compilers=[:c, :go]`.  Go-based packages can usually be built and installed with `go`:
```

PR #1126, opened 2021-12-14
[docs] explain use of Preferences.jl to override JLL

odow:
Instead of https://github.com/JuliaLang/Pkg.jl/pull/2889

### DIFF
--- a/docs/src/jll.md
+++ b/docs/src/jll.md
@@ -282,3 +282,24 @@ Some comments about how to write this file:
   you want to use to override the products of the JLL package must have the same
   tree structure as the artifact.  In our example we need to use `/usr` to
   override Fontconfig and `/usr/local` for Bzip2.
+
+### Overriding specific products
+
+Instead of overriding the entire artifact, you can override a particular library
+within a JLL using [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl).
+
+For example, to override our `libbz2` example:
+```julia
+using Preferences
+set_preferences!(
+    "LocalPreferences.toml",
+    "Bzip2_jll",
+    "libbzip2_path" => "/usr/local/lib/libbz2.so",
+)
+```
+Note that the product name is `libbzip2`, but we use `libbzip2_path`.
+
+!!! warning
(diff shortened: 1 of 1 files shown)


docs/src/jll.md
```diff
@@ -282,3 +282,24 @@ Some comments about how to write this file:
   you want to use to override the products of the JLL package must have the same
   tree structure as the artifact.  In our example we need to use `/usr` to
   override Fontconfig and `/usr/local` for Bzip2.
+
+### Overriding specific products
+
+Instead of overriding the entire artifact, you can override a particular library
```
giordano, line 288: I'd say something like
```suggestion
Instead of overriding the entire artifact, you can override a particular product (library, executable, etc...)
```

docs/src/jll.md
```diff
@@ -282,3 +282,24 @@ Some comments about how to write this file:
...
+    "Bzip2_jll",
+    "libbzip2_path" => "/usr/local/lib/libbz2.so",
+)
+```
+Note that the product name is `libbzip2`, but we use `libbzip2_path`.
+
+!!! warning
+    There are two common cases where this will not work:
+    1. The JLL is part of the [Julia stdlib](https://github.com/JuliaLang/julia/tree/master/stdlib), for example `Zlib_jll`
+    2. The JLL has not been compiled with [JLLWrappers.jl](https://github.com/JuliaPackaging/JLLWrappers.jl) as a dependency, for example, `Gzip_jll`.
```
giordano, line 305: I wouldn't hardcode the name of packages here, in the future they may be regenerated to use JLLWrappers

giordano (commented):
It may be worth pointing out that this requires Julia v1.6. Yes, v1.6 is now the LTS and it's increasingly unlikely there will be users on older versions, but this is still relevant for completeness

docs/src/jll.md
vchuravy: Maybe even add a call to action?

odow: How's this?

giordano: Looks good, thanks!

PR #1127, opened 2021-12-14
Prepend binary name to wizard PRs

jeremiahpslewis:
Proposal: Make wizard PR titles include binary name in brackets, e.g. "[jemalloc] Wizard recipe: jemalloc-v5.2.1 (#4064)".

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -538,9 +538,9 @@ function register_jll(name, build_version, dependencies, julia_compat;
         upstream_registry_url = "https://github.com/JuliaRegistries/General"
         name_jll = "$(name)_jll"
         if _package_is_registered(upstream_registry_url, name_jll)
-            pr_title = "New version: $(name_jll) v$(build_version)"
+            pr_title = "[$(name)] New version: $(name_jll) v$(build_version)"
         else
-            pr_title = "New package: $(name_jll) v$(build_version)"
+            pr_title = "[$(name)] New package: $(name_jll) v$(build_version)"
         end
         # Open pull request against JuliaRegistries/General
         body = """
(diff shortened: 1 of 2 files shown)


src/AutoBuild.jl
```diff
@@ -538,9 +538,9 @@ function register_jll(name, build_version, dependencies, julia_compat;
         upstream_registry_url = "https://github.com/JuliaRegistries/General"
         name_jll = "$(name)_jll"
         if _package_is_registered(upstream_registry_url, name_jll)
-            pr_title = "New version: $(name_jll) v$(build_version)"
+            pr_title = "[$(name)] New version: $(name_jll) v$(build_version)"
```
giordano, line 541: The name is already there, no?

jeremiahpslewis:
```suggestion
            pr_title = "[$(name)] New version: v$(build_version)"            
```

Yes, just not at beginning of title, would you remove it from the second part like this?

giordano: I mean, I don't understand the need to change this just for the sake of changing? What problem is this meant to solve?

jeremiahpslewis: The PR is about taking a syntax / pattern which has become the (manual) dominant pattern and making it the automatic default pattern. It's a non-technical change and one which is a minor tweak, so feel free to close this PR, but I was checking out the BinaryBuilder codebase and it was easy to see how it might be implemented. Here are a few screen shots to compare; perhaps it would be good to get further community feedback before implementing though?
Screen Shot 2021-12-14 at 16 07 06

Screen Shot 2021-12-14 at 16 06 49
Screen Shot 2021-12-14 at 16 06 21

Screen Shot 2021-12-14 at 16 06 37

PR #1133, opened 2021-12-18
Allow for non jll dependencies

### LINKED ISSUE #1132: Fictional UUID created for non JLL package
In a local version of https://github.com/JuliaPackaging/Yggdrasil/pull/4073 I am experimenting with adding a non `jll` dependency.
Thanks to @giordano who pointed me to the right place. We should look up the UUID instead of synthesizing a new one.

https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/084d75b38aa9d0a0240d328d9ff3adbc9c7d9068/src/AutoBuild.jl#L1442

cc: @staticfloat

vchuravy:
Fixes #1132

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1425,6 +1425,29 @@ const uuid_package = UUID("cfb74b52-ec16-5bb7-a574-95d9e393895e")
 # For even more interesting historical reasons, we append an extra
 # "_jll" to the name of the new package before computing its UUID.
 jll_uuid(name) = bb_specific_uuid5(uuid_package, "$(name)_jll")
+
+function find_uuid(ctx, pkg)
+    if Pkg.Types.has_uuid(pkg)
+        return pkg.uuid
+    end
+    depname = getname(pkg)
+    @static if VERSION >= v"1.7"
+        uuids = Pkg.Types.registered_uuids(ctx.registries, depname)
+    else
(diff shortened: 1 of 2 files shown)


src/AutoBuild.jl
```diff
@@ -1425,6 +1425,29 @@ const uuid_package = UUID("cfb74b52-ec16-5bb7-a574-95d9e393895e")
...
+        return pkg.uuid
+    end
+    depname = getname(pkg)
+    @static if VERSION >= v"1.7"
+        uuids = Pkg.Types.registered_uuids(ctx.registries, depname)
+    else
+        uuids = Pkg.Types.registered_uuids(ctx, depname)
+    end
+    if isempty(uuids)
+        return nothing
```
giordano, line 1440: I was thinking of using `jll_uuid` here, but maybe it's better to have this function returning the real UUID, if any, only.

giordano (approved):
Looks good to me, as long long as tests pass.  Thanks!

giordano:
Yay, more internal and undocumented Pkg functions, to make future upgrades more fun 🥳

src/AutoBuild.jl
vchuravy: Yeah the first version did that, but then I was like: Maybe I want to use that function elsewhere xD

PR #1138, opened 2021-12-28
update docs on debug flag

dpo:
--debug by itself no longer drops the user to a shell.
Replace --debug with --debug=error in the documentation.


giordano:
> --debug by itself no longer drops the user to a shell.

What? I use `--debug` all the time

dpo:
Well sorry. I must be doing something wrong. Neither `--debug` nor `--debug=error` drop me to a shell.

PR #1139, opened 2021-12-29
make a global variable const

KristofferC:
Was just browsing and saw this. I guess this can be `const`.


giordano:
Sounds good, thanks!

PR #1140, opened 2021-12-29
improve performance of instruction set checker

KristofferC:
Running this on the product from OpenBlas in yggdrasil takes this part from 

```
5.385089 seconds (48.11 M allocations: 4.965 GiB, 11.82% gc time)
```

to

```
1.505047 seconds (13.24 M allocations: 1.082 GiB, 8.13% gc time)
```

It is still not super significant because running the ´objdump` above takes like 15 seconds, but it is a start.

### DIFF
--- a/src/auditor/instruction_set.jl
+++ b/src/auditor/instruction_set.jl
@@ -44,12 +44,15 @@ function instruction_mnemonics(path::AbstractString, platform::AbstractPlatform)
     run_interactive(ur, `/bin/bash -c "$(objdump_cmd)"`; stdout=output, stderr=devnull)
     seekstart(output)
 
-    for line in eachline(output)
+    @time for line in eachline(output)
         isempty(line) && continue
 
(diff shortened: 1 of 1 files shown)


giordano:
Nice to see there are some low-hanging fruits!

KristofferC:
I think there are quite a lot of low hanging fruits. The next one is to read the output from objdump as it is being produced instead of waiting for it to finish before starting the string analysis.

src/auditor/instruction_set.jl
```diff
@@ -44,15 +44,22 @@ function instruction_mnemonics(path::AbstractString, platform::AbstractPlatform)
     run_interactive(ur, `/bin/bash -c "$(objdump_cmd)"`; stdout=output, stderr=devnull)
     seekstart(output)
 
-    for line in eachline(output)
+    @time for line in eachline(output)
```
KristofferC, line 47: 😭 😭 😭 😭

PR #1152, opened 2022-01-09
[Autobuild] Add log file prompt

jeremiahpslewis:
Resolves #1149

If *.log files are present, yields:

```bash
┌ Warning: Build failed, the following log files were generated:
│  - build/x86_64-linux-musl/4Rb7HiTe/srcdir/curl-7.81.0/config.log
│ 
│ Launching debug shell:
└ @ BinaryBuilder ~/.julia/dev/BinaryBuilder/src/AutoBuild.jl:816
```

Otherwise (in case of no log files), no change from current version.

PR #1154, opened 2022-01-14
[Autobuild] Add log file prompt

### LINKED ISSUE #1149: Feature: Autodiscover and add `*.log` build log file paths when debug shell starts
In the belief that I'm not the only one that makes this mistake *constantly*, and seeing as it's strongly encouraged in the [docs](https://docs.binarybuilder.org/stable/troubleshooting/#General-comments), I wonder whether we might highlight the file paths to common log files like `config.log` to encourage the user to check them out when debugging builds and make it easy for them to option click on the paths and get to the file.

I.e. extending this like https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/a56e405b1c4a1a8888c92328310fd800c54f77de/src/AutoBuild.jl#L794

To something like:

```bash
Build failed, the following log files were generated:
    x86_64-linux-musl/Jjl19OI7/srcdir/curl-7.81.0/config.log

Launching debug shell now:
```

jeremiahpslewis:
Supersedes https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1152

Resolves #1149

If *.log files are present, yields:

```bash
┌ Warning: Build failed, the following log files were generated:
│   - ${WORKSPACE}/srcdir/wget-1.21.1/config.log
│ 
│ Launching debug shell:
└ @ BinaryBuilder ~/.julia/dev/BinaryBuilder/src/AutoBuild.jl:822
```

Otherwise (in case of no log files), no change from current version.

### DIFF
--- a/test/basic.jl
+++ b/test/basic.jl
@@ -102,7 +102,7 @@ end
 
 
         # Test that debug prompt generation works
-        @testset "Check debug prompt logic"
+        @testset "Check debug prompt logic" begin
             @test "Build failed, launching debug shell:" == BinaryBuilder.compose_debug_prompt(build_path, temp_path)
             logfile_path = joinpath(build_path, "srcdir", "errors.log")
             open(logfile_path, "w") do io
--- a/test/basic.jl
+++ b/test/basic.jl
@@ -108,7 +108,7 @@ end
             open(logfile_path, "w") do io
                 write(io, "log message")
             end
-            @test "Build failed, the following log files were generated:\n    $(replace(logfile_path, "$temp_path/" => ""))\n\nLaunching debug shell:\n" == BinaryBuilder.compose_debug_prompt(build_path, temp_path)
+            @test "Build failed, the following log files were generated:\n    - $(replace(logfile_path, "$temp_path/" => ""))\n\nLaunching debug shell:\n" == BinaryBuilder.compose_debug_prompt(build_path, temp_path)
             rm(logfile_path)
         end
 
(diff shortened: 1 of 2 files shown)


jeremiahpslewis:
@giordano Updated and hopefully ready to go. The file paths are also 'option-click-able' in VSCode, so you don't just see the files, it's also easy to pull up the logs while debugging.

test/basic.jl
```diff
@@ -100,6 +100,18 @@ end
...
         @test isfile(env_file)
 
+
+        # Test that debug prompt generation works
+        @testset "Check debug prompt logic" begin
+            @test "Build failed, launching debug shell:" == BinaryBuilder.compose_debug_prompt(build_path)
+            logfile_path = joinpath(build_path, "srcdir", "errors.log")
+            open(logfile_path, "w") do io
+                write(io, "sample log message")
+            end
```
giordano, line 110:
```suggestion
            write(logfile_path, "sample log message")
```

test/basic.jl
```diff
@@ -100,6 +100,18 @@ end
         @test isfile(hist_file)
         @test isfile(env_file)
 
+
```
giordano, line 103:
```suggestion
```

test/basic.jl
```diff
@@ -100,6 +100,18 @@ end
...
+        # Test that debug prompt generation works
+        @testset "Check debug prompt logic" begin
```
giordano, line 105: I don't think this test has anything to do with "environment and history saving", so this might just be its own testset?

jeremiahpslewis: Good point; simplified it as well.

giordano (approved):
This looks much better now, thanks! :+1:

jeremiahpslewis:
Thank you!

PR #1159, opened 2022-01-26
[AutoBuild] Fix initial value of `products_info`

giordano:


src/AutoBuild.jl
```diff
@@ -1175,7 +1175,7 @@ function build_jll_package(src_name::String,
     dependencies = Dependency[d for d in dependencies if is_runtime_dependency(d)]
 
     platforms = keys(build_output_meta)
-    products_info = Dict{Product,Any}
+    products_info = Dict{Product,Any}()
```
giordano, line 1178: Actually, is this definition needed at all? I need to double check

PR #1168, opened 2022-02-07
Update dependencies for m1

jeremiahpslewis:
Update dependencies so that all support apple silicon

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,7 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.1"
+julia_version = "1.8.0-DEV.1436"
 manifest_format = "2.0"
+project_hash = "84719174863332abcfe95c1b5909f3d7c0ac3fbb"
 
 [[deps.ArgParse]]
 deps = ["Logging", "TextWrap"]
@@ -11,6 +12,7 @@ version = "1.1.4"
 
 [[deps.ArgTools]]
 uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+version = "1.1.1"
 
 [[deps.Artifacts]]
 uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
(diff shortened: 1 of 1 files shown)


Manifest.toml
```diff
@@ -1,7 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.1"
+julia_version = "1.8.0-DEV.1436"
```
giordano, line 3: Ci here runs on v1.7

giordano (commented):
You know right that this PR is basically doing nothing?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "BinaryBuilder"
 uuid = "12aac903-9f7c-5d81-afc2-d9565ea332ae"
 authors = ["Elliot Saba <staticfloat@gmail.com>"]
-version = "0.5.1"
+version = "0.5.2"
```
giordano, line 4: Changing the manifest has no effect besides CI, what's the point of a new release?

giordano (commented):
You can't just edit the manifest manually to change the Julia version, you have to re-resolve it with julia v1.7. Besides, as I've been saying this PR is basically doing nothing, I'm missing the point

jeremiahpslewis:
@giordano Can you correct my understanding, I'm modeling it off of https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1165, thought that once BB.jl has been updated, a PR to Yggdrasil like this one (https://github.com/JuliaPackaging/Yggdrasil/pull/4371) would ensure that these dependencies are used by the ci.

...but, now that I've written it out, it's clearly irrelevant for the CI seeing as users who install BinaryBuilder will pick up dependencies based on Project.toml and not Manifest.toml. Nonetheless, if there are unknown issues introduced by the new _jll builds (unrelated to M1 platform), we won't see them if we don't use it extensively, so not entirely a bad move?

giordano:
The point of #1165 was to

* test with newer version of BinaryBuilderBase.jl, since there was the update of the rootfs and I needed to make sure everything was working fine on this side. https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/215 was merged only tests were passing here. This PR is just updating some minor dependencies.
* release a new version, since there had been none in a while. This is literally the only change since 0.5.1, _nothing_ else changed, this definitely doesn't warrant a new release 

This is only updating

PR #1170, opened 2022-02-08
[AutoBuild] Add `Documentation` section to generated README

giordano:
The line about documentation currently appears in the `Bug Repors` section,
which looks odd.  Also, removed an extra newline.

CC: @sjkelly


sjkelly:
Good catch, thanks!

PR #1190, opened 2022-03-22
[AutoBuild] Make artifacts lazy by default if augmenting platform

### LINKED ISSUE #1179: Set lazy=true when using platform augmentation
https://github.com/JuliaPackaging/Yggdrasil/pull/4527#discussion_r816997233

We should do this as long as `compat` includes 1.6.

giordano:
CC: @vchuravy.  Verified it works in https://github.com/giordano/finufft_jll.jl/commit/e63f2118485986a074719c26436706ff170196e4

Fix #1179.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -43,6 +43,11 @@ function Base.show(io::IO, t::BuildTimer)
     end
 end
 
+# Helper function to get the minimum version supported by the given compat
+# specification, given as a string.
+minimum_compat(compat::String) =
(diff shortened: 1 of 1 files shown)


giordano:
Uhm, this wasn't quite enough for the registration process we use in Yggdrasil.  I suspect having a different default value for `lazy_artifacts` in https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/28af078295cdcd781f64d5d710237d5217bfbe57/src/AutoBuild.jl#L1067 is quite fishy.

PR #1194, opened 2022-03-23
[Auditor] Don't try to dlopen libraries for incompatible ISAs

### LINKED ISSUE #1193: Auditor tries to dlopen libraries ignoring the `march` tag
At the moment the auditor tries to dlopen the library under this condition: https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/39554446b20e8ff9427d9cbd2b163776a4744b2f/src/Auditor.jl#L140 but
```julia
julia> platform = HostPlatform()
Linux x86_64 {cxxstring_abi=cxx11, julia_version=1.7.2, libc=glibc, libgfortran_version=5.0.0, libstdcxx_version=3.4.29}

julia> platform["march"] = "avx512"
"avx512"

julia> platforms_match(platform, HostPlatform())
true
```
this means that `HostPlatform()` matches anything derived from `HostPlatform()` which differs only on the `march` tag, which is ignored.  In particular this means we'll try to dlopen an AVX512 library on the build machine, but for example on Yggdrasil we don't have an AVX512-equipped system, so we get an illegal instruction error.

We need a custom comparison strategy for the `march` key, similarly to what we do for the `os_version` in Julia `Base`: https://github.com/JuliaLang/julia/blob/62e0729dbc5f9d5d93d14dcd49457f02a0c6d3a7/base/binaryplatforms.jl#L287-L317

giordano:
Fix #1193.


giordano:
So, I verified that this fixes the attempt to dlopen the library for an incompatible ISA, but for some reasons I still don't grasp audit fails anyway to locate the products :confused:  I'll need to do more debugging.

giordano:
It turns out we try to dlopen libraries twice.... :facepalm:  We need to repeat the same exact thing in https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/55c795dbca6c74b076636bdcfdccbfeb34063eb7/src/Products.jl#L225

PR #1209, opened 2022-06-11
[AutoBuild] Filter out non-build deps before calling `setup_dependencies`

giordano:
Follow-up to https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/249.  CC: @vchuravy

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -785,7 +785,7 @@ function autobuild(dir::AbstractString,
             verbose=verbose,
         )
         setup_deps(f, prefix, dependencies, platform, verbose) =
-            setup_dependencies(prefix, Pkg.Types.PackageSpec[getpkg(d) for d in filter_platforms(dependencies, platform) if f(d)], platform; verbose)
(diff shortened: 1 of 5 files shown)


src/AutoBuild.jl
```diff
@@ -785,7 +785,7 @@ function autobuild(dir::AbstractString,
             verbose=verbose,
         )
         setup_deps(f, prefix, dependencies, platform, verbose) =
-            setup_dependencies(prefix, Pkg.Types.PackageSpec[getpkg(d) for d in filter_platforms(dependencies, platform) if f(d)], platform; verbose)
+            setup_dependencies(prefix, Pkg.Types.PackageSpec[getpkg(d) for d in filter_platforms(dependencies, platform) if f(d) && is_build_dependency(d)], platform; verbose)
```
giordano, line 788: This was the main missing bit: we weren't filtering out non-build dependencies, because before https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/249 all dependencies were also build-dependencies.

vchuravy:
If you want a package to try this one libclangext would be one.

giordano:
Let's test in production! 😁

PR #1220, opened 2022-08-16
Support for HTTP.jl 1.0 release

quinnj:


giordano:
Note that without changing the Manifest we aren't actually testing against v1: https://dev.azure.com/JuliaPackaging/BinaryBuilder.jl/_build/results?buildId=21326&view=logs&j=52b15b81-63c8-535b-eb97-b9bb2162f36f&t=069f5c50-6918-5819-047d-dae88da1825f&l=108.

PR #1222, opened 2022-08-23
[CI] Update Manifest.toml

giordano:
Follow up to #1220


ararslan (approved):
As an aside, does the manifest actually need to be checked in here? Most packages don't.

giordano:
> does the manifest actually need to be checked in here? Most packages don't.

This package was created by Elliot, he loves checking in the Manifest :stuck_out_tongue:

PR #1228, opened 2022-09-18
Throw EOFError when encountering Ctrl-D

### LINKED ISSUE #1227: `Ctrl-D` causes infinite loop of line prompt
https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/f20e6d0b9ddd66a097fa4a0beab70c2ea2f9b704/src/wizard/utils.jl#L48-L57

To reproduce:

- Run the wizard
- Go to step 2
- When the prompt asks you to enter a URL, press `Ctrl-D`
- Infinite loop starts

lucifer1004:
This PR fixes #1227.

### DIFF
--- a/src/wizard/utils.jl
+++ b/src/wizard/utils.jl
@@ -33,6 +33,9 @@ function line_prompt(name, msg; ins=stdin, outs=stdout, force_identifier=false, 
         else
             val = strip(read(_getpass(ins), String))
         end
+        if eof(ins)
+            throw(EOFError())
+        end
         if !isopen(ins)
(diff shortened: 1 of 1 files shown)


giordano:
It doesn't look like this worked very well

giordano:
That test has nothing to do with the wizard though.  Comment out https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/c3ab9105f9f4787eda1deee5586a58738d4c9f40/test/runtests.jl#L69-L72 to run only the wizard tests.

Issue #778, opened 2020-05-06
Generate multiple packages with the a single builder


giordano:
From time to time I look to the package managers of Linux distributions to see if we can pick up some interesting ideas.  One thing that I think would be really cool to have here is to be able to generate multiple JLL packages with a single builder: the result of a build doesn't go into a single tarball, but it might be split into many of them.

One fancy application is to be able to generated:
* `Libfoo_jll`: contains only `bin/` and `lib/`, this is the runtime part, what the Julia packages will use;
* `Libfoo_dev_jll`: contains `include/`, header files are generally useless for Julia packages and they mostly clutter `~/.julia/artifacts/` with dozens of small filess.  Ideally, this would be automatically installed, if available, when `Libfoo_jll` is used as dependency in a build;
* `Libfoo_dbg_jll`: contains the debug symbols of the shared library, that users can optionally install to get more useful debug information about crashes or errors.  Based on an idea by @keno.

Also, I think that `LLVM_full_jll` is currently "wrong": IMO it should simply be an empty metapackage binding all the other pieces.  Instead now it's a monster package containing the same data as its pieces, which means that if we use both `LLVM_full_jl` and `libLLVM_jll` in a build, they would step onto each other's toes.  Having a single builder that produces all other subpackages would probably make @vchuravy happy, too.

Keno:
maybe also a jll for the original source directory such that we can download it automatically in the debugger if necessary.

staticfloat:
I agree that this is desirable.  I'm not entirely sure that the right way to do it is to create multiple JLL packages; or at least, not necessarily the user-facing way.

Here are my thoughts:

* For some projects, we have the genuine desire to split a JLL into multiple independent packages; `Clang_jll` and `LLVM_jll` really don't have anything to do with eachother; sometimes you may want `Clang_jll` and not `LLVM_jll` and vice-versa.  The fact that they both stem from the same build process is more or less an implementation detail.  (Oh, and they both rely upon `LibLLVM_jll`, but that's fine). To save on build time/duplicated effort, it would be nice to be able to split a single `build_tarballs()`'s output into multiple, independent, JLL packages.

[...]

PR #1229, opened 2022-09-18
[AutoBuild] Split log files into a separate tarball

giordano:
Long term plan is still to do #778, but that's a lot of work, and we have little time at the moment.  Simply splitting the logs into a separate tarball, instead, is a simpler short-term solution which has some very important bonuses:

- the git tree hash of the main tarball should be reproducible (as long as you use the same toolchain)
- which means we can start testing reproducibility of builds
- rebuilding a package without changing its content won't create a new artifact: less strain on the Package Storage Server.

I still need to do tests about deployment on Yggdrasil though (***Edit***: see https://github.com/JuliaPackaging/Yggdrasil/pull/5526).

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,7 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.3"
+julia_version = "1.8.1"
 manifest_format = "2.0"
+project_hash = "f35810f1518b3c855e1d8ed4f1522241fc6dcb7a"
 
 [[deps.ArgParse]]
 deps = ["Logging", "TextWrap"]
@@ -11,6 +12,7 @@ version = "1.1.4"
 
 [[deps.ArgTools]]
 uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+version = "1.1.1"
 
 [[deps.Artifacts]]
 uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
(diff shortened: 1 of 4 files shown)


giordano:
Uhm, the failure
```
[ Info: Building for x86_64-linux-gnu-libgfortran5-cxx11-libstdcxx29-julia_version+1.7.2, x86_64-unknown-freebsd
JLLs - building: Error During Test at /data/staticfloat/bb_azp/rootfs/agent_1/_work/1/s/test/jll.jl:41
  Unable to locate LibraryProduct(["libfoo"], :libfoo, String[], false, Symbol[]) within /data/staticfloat/bb_azp/rootfs/agent_1/tmp/jl_26NC7I for x86_64-linux-gnu-libgfortran5-cxx11-libstdcxx29-julia_version+1.7.2
   [15] include(fname::String)
      @ Base.MainInclude ./client.jl:451
   [16] top-level scope
      @ /data/staticfloat/bb_azp/rootfs/agent_1/_work/1/s/test/runtests.jl:72
   [17] include(fname::String)
      @ Base.MainInclude ./client.jl:451
   [18] top-level scope
      @ none:6
   [19] eval
      @ ./boot.jl:373 [inlined]
   [20] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:268
   [21] _start()
      @ Base ./client.jl:495
```
looks unrelated to the changes to this package in this PR, but maybe some other upstream changes in BBB broke it (but we haven't seen any widespread breakage).  I'll have to dig into this.

staticfloat (approved):

PR #1233, opened 2022-09-30
[docs] Add section about reproducibility

giordano:
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/268, https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/270, and #1229 made it easier to investigate our reproducibility capabilities, and I think we are at a place where we can start talking about them.

Side note, in Yggdrasil we've had a [reproducibility scanning tool](https://github.com/JuliaPackaging/Yggdrasil/blob/0b127481fd158ef78b54283b39634831206da3e8/reproducibility_scanning_tools.jl) for over two years and the very few times I used it I always found reproducible builds, but it was a bit annoying to use as it had to extract the content of the tarballs: now instead we can directly look at the SHA256 sum, or the git tree hash.  But this also gives you an idea of how long our builds have been reproducible in practice (although verifying it was cumbersome)

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -67,10 +67,10 @@ If you rebuild multiple times the same package, with the same version of BinaryB
 
 There are however some caveats:
 
-* reproducibility can not be expected if not using the toolchain offered by `BinaryBuilder.jl`;
+* reproducibility can only be expected when using the toolchain offered by `BinaryBuilder.jl`;
 * there are [very specific cases](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1230) where the macOS C/C++ toolchain does not produce reproducible binaries.
   This happens when doing debug builds (`-g` flag) _and_ not building object files with deterministic names separately (e.g. if directly building and linking a program or a shared library from the source file, letting the compiler create the intermediate object files automatically with random names).
-  We have decided not to take action for this case because in practice most packages use build systems which compile intermediate object files with deterministic names (which is also the only way to take advante of `ccache`, which `BinaryBuilder.jl` uses extensively) and typically do not do debug builds, thus sidestepping the issue entirely.
+  We have decided not to take action for this case because in practice most packages use build systems which compile intermediate object files with deterministic names (which is also the only way to take advantage of `ccache`, which `BinaryBuilder.jl` uses extensively) and typically do not do debug builds, thus sidestepping the issue entirely.
 
 ## Videos and tutorials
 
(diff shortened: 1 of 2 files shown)


docs/src/index.md
```diff
@@ -53,6 +53,25 @@ If you already have access to the [GitHub Codespaces](https://github.com/feature
...
+
+While `BinaryBuilder.jl` does not guarantee to always have reproducible builds, it achieves this goal in most cases.
+Reproducibility in `BinaryBuilder.jl` includes also the generated tarballs: they are created with [`Tar.jl`](https://github.com/JuliaIO/Tar.jl), which takes [a few measures](https://github.com/JuliaIO/Tar.jl/blob/1de4f92dc1ba4de4b54ac5279ec1d84fb15948f6/README.md#reproducibility) to ensure reproducibility of tarballs with the same git tree hash.
+If you rebuild multiple times the same package, with the same version of BinaryBuilder, the generated tarball which contains the main products (i.e. not the log files which are known not to be reproducible) should always have the same git tree hash and SHA256 sum, information which are printed to screen at the end of the build process and stored in the `Artifacts.toml` file of the [JLL package](@ref JLL-packages).
+
+There are however some caveats:
+
+* reproducibility can not be expected if not using the toolchain offered by `BinaryBuilder.jl`;
```
ericphanson, line 70: the double negative can be hard to parse:
```suggestion
* reproducibility can only be expected when using the toolchain offered by `BinaryBuilder.jl`;
```

docs/src/index.md
```diff
@@ -53,6 +53,25 @@ If you already have access to the [GitHub Codespaces](https://github.com/feature
...
+* there are [very specific cases](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1230) where the macOS C/C++ toolchain does not produce reproducible binaries.
+  This happens when doing debug builds (`-g` flag) _and_ not building object files with deterministic names separately (e.g. if directly building and linking a program or a shared library from the source file, letting the compiler create the intermediate object files automatically with random names).
+  We have decided not to take action for this case because in practice most packages use build systems which compile intermediate object files with deterministic names (which is also the only way to take advante of `ccache`, which `BinaryBuilder.jl` uses extensively) and typically do not do debug builds, thus sidestepping the issue entirely.
```
ericphanson, line 73:
```suggestion
  We have decided not to take action for this case because in practice most packages use build systems which compile intermediate object files with deterministic names (which is also the only way to take advantage of `ccache`, which `BinaryBuilder.jl` uses extensively) and typically do not do debug builds, thus sidestepping the issue entirely.
```

PR #1240, opened 2022-10-26
[Auditor] Do not skip GCC libraries when updating linkage

giordano:
I feel like these libraries were skipped in the old days before CompilerSupportLibraries was a thing.  Ref: https://github.com/JuliaPackaging/Yggdrasil/issues/3703#issuecomment-1291225562.  CC: @vtjnash @apaz-cli


src/auditor/dynamic_linkage.jl
```diff
@@ -313,31 +311,6 @@ end
 # Determine whether a library is a "default" library or not, if it is we need
 # to map it to `@rpath/$libname` on OSX or `\$ORIGIN/$libname` on Linux/FreeBSD
 is_default_lib(lib, oh) = false
```
giordano, line 313: I honestly don't understand what's the purpose of this function.

test/auditing.jl
```diff
@@ -624,6 +624,52 @@ end
...
+            @test build_output_meta[platform][3] == Base.SHA1("06ddfbeb9914a534ed3f21795b5da5b536d33c16")
+
+            # Unpack it somewhere else
+            @test verify(tarball_path, tarball_hash)
+            testdir = joinpath(build_path, "testdir")
+            mkdir(testdir)
+            unpack(tarball_path, testdir)
+            # Make sure auditor set the rpath of `hello_world`, even if it links only to
+            # libgfortran.
+            @test Auditor._rpaths(joinpath(testdir, "bin", "hello_world_fortran")) == ["\$ORIGIN/../lib"]
```
giordano, line 670: Without the change, the RPATH would be empty because all required libraries would be skipped.

vtjnash:
Thanks!

src/auditor/dynamic_linkage.jl
staticfloat: This method is for when `oh` is not a `MachOHandle` or an `ELFHandle`; e.g. when it's a `COFFHandle`, or some other handle type that we add in the future to `ObjectFile`

staticfloat (approved):

staticfloat:
Looks like we need to update the locked gitshas in some other tests?

giordano:
> Looks like we need to update the locked gitshas in some other tests?

The failing tests were gfortran builds on Linux which changed the SHA, but not the Windows ones, which is all consistent with the changes in this PR.  Isn't it exciting? 😃

giordano:
This PR is highlighting that so far for the most part we were relying on some system GCC libraries (like `libgcc_s`) to be available, but if we expect that julia can run on a system without GCC that was a wrong assumption.  I had to correct a few auditor tests that were relying on it.  Instead I think we should always use `CompilerSupportLibraries` if stuff like `libgcc_s` is needed

src/auditor/dynamic_linkage.jl
giordano: The method yes, I understand, but not the function as a whole.  With this PR this is the only method of this function, so we can just remove it entirely

vtjnash:
Thanks for hunting all that down

PR #1255, opened 2023-01-07
Add buildkite logger

vchuravy:


src/Logging.jl
```diff
@@ -38,3 +38,40 @@ function enable_azure_logging()
...
+    #Make it a named tuple for easier working
+    log = LoggingExtras.handle_message_args(args...)
+
+    # Buildkite calls it `style` not level`, and `warning` not `warn`:
+    log[:level]
+    stylemap = Dict(Logging.Error => "error", Logging.Warn => "warning")
+    style = stylemap[log[:level]]
+
+    # TODO pretty rendering of properties
+    annotate(log.message; context="BinaryBuilder", style)
```
vchuravy, line 64: We should markdown format the message. At least add a `\n`

vchuravy: Not sure what the context should be.

PR #1267, opened 2023-03-08
Require at least RegistryTools 2.1.0

DilumAluthge:
We need at least RegistryTools 2.1.0 so that we can get https://github.com/JuliaRegistries/RegistryTools.jl/pull/85.


giordano:
You need to update the Manifest

DilumAluthge:
Done. I used Julia 1.7.2 when updating the manifest.

Command I used to update the manifest:

```
add --preserve=all RegistryTools@2.1 Registrator@1.8
```

giordano:
This is broken https://buildkite.com/julialang/yggdrasil/builds/1294#0186c3c4-a69f-4cef-b8f7-981600dc5f36, I'm going to revert this urgently

GunnarFarneback:
To state the obvious, RegistryTools 2.0 bumped the major version because it was breaking.

PR #1269, opened 2023-03-13
Upgrade to RegistryTools v2, and require at least RegistryTools 2.1.0

DilumAluthge:
Reverts JuliaPackaging/BinaryBuilder.jl#1268 (which reverted #1267)


giordano:
I mean, I reverted #1267 because it was broken.  Unless you fix the breaking changes, this PR is not going to be accepted in the present form.

To make it crystal clear, we don't test registration at all, so don't expect to be able to catch errors that way.  Of course, unless you volunteer to set up a mock registry for testing, which would be much welcome).

DilumAluthge:
https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1269/commits/d9f2fbf7b82211d6132a7ef9ee9e7904807a50b3 fixes the breaking changes. AFAICT that's the only place where RegistryTools is used.

Yeah, I don't know how to write a test for this, short of setting up a test registry, etc, like we do for RegistryCI.

DilumAluthge:
Bump @giordano @staticfloat 

The second commit here should fix the breaking changes. Obviously I have no way of testing that. The commit is based on my reading of the RegistryTools source code, as well as the comment in the RegistryTools README about upgrading to the latest RegistryTools release.

staticfloat (approved):
Tested on Julia v1.7

PR #1304, opened 2023-11-13
Don't request username in init_jll_package

andreasnoack:
The username might not be available if authenticating as a GitHub App. Also, when authenticating with a token, the username isn't necessary.


staticfloat (approved):

DilumAluthge:
It's still timing out even after I retried it.

DilumAluthge:
@staticfloat Any idea what's going on here?

staticfloat:
Looks like it just takes more than 120m to run the tests and download all the shards. :(

giordano:
I don't think it has to download anything.

PR #1305, opened 2023-11-24
Don't query GitHub username and use dummy credentials when committing

andreasnoack:
This is a followup to https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1304 since there were two more authentications that relied on querying the potentially non-existent GitHub username. 

The second commit might not be the right approach but it allowed me to build new binaries while authenticating as a GitHub app. The issue is that https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/ad70f55fb09aeb13324e4e69bf2c6df4b4e2fc4c/src/AutoBuild.jl#L1589 ends up querying username information from the local git repo when calling https://github.com/JuliaLang/julia/blob/187e8c2222878c68b2afc9295ab8dc61773bd7f2/stdlib/LibGit2/src/signature.jl#L66-L72.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1581,13 +1581,12 @@ end
 function push_jll_package(name, build_version;
                           code_dir = joinpath(Pkg.devdir(), "$(name)_jll"),
                           deploy_repo = "JuliaBinaryWrappers/$(name)_jll.jl",
-                          gh_auth = Wizard.github_auth(;allow_anonymous=false),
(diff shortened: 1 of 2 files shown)


andreasnoack:
After discussions with @DilumAluthge, I've dropped the second commit that used dummy credentials. Instead, the user should set the git default user because building the binaries.

DilumAluthge:
Thanks Elliot!

Issue #1311, opened 2024-01-02
Documentation for linking against LAPACK/BLAS


mattsignorelli:
Since this is probably a common thing, I feel like there should be a section in the documentation for properly linking against LAPACK/BLAS. For example, I developed a binary package and I had to message the Slack to learn which `_jll`'s to use. I saw other people in the Slack had similar issues, and all the answers are slightly different. I also ended up linking against the wrong BLAS (linked `LAPACK32_jll` to `ReferenceBLAS_jll`, because I didn't know that `ReferenceBLAS32_jll` existed). 

Just a short section in the documentation that basically says:

For LP64 LAPACK/BLAS, use `LAPACK32_jll` with either `OpenBLAS32_jll` or `ReferenceBLAS32_jll` 
for ILP64 LAPACK/BLAS, use `LAPACK_jll` with either `OpenBLAS_jll` or `ReferenceBLAS_jll`

I'd be happy to prepare the wording in a pull request for this

PR #1312, opened 2024-01-02
Add LAPACK/BLAS linking to FAQ

mattsignorelli:
Added a short FAQ in the documentation about how to properly link against LAPACK/BLAS in response to #1311


docs/src/FAQ.md
```diff
@@ -105,4 +105,8 @@ However, not all ecosystems follow SemVer. The following two cases are quite com
...
\ No newline at end of file
+While the SemVer (and Julia's) "conservative" approach to compatibility ensures there will never be runtime crashes due to installed incompatible libraries, you might still end up with systems that refuse to install in the first place (which the Julia ecosystem considers the lesser evil). E.g., package `A.jl` that depends on newer versions of `CLF` and package `B.jl` that depends on older versions can not be installed at the same time. This happens less often in ecosystems that follow semver, but might happen relatively often in an ecosystem that does not. Thus developers that rely on `jll` packages that do not follow semver should be proactive in updating their compat bonds (and are strongly encouraged to heavily use the `CompatHelper` bot).
+
+### How do I link against LAPACK/BLAS?
+
+For the ILP64 interface (64-bit integers) of LAPACK, use the binary dependencies `LAPACK_jll` and either `ReferenceBLAS_jll` ([Netlib Reference BLAS](https://www.netlib.org/blas/)) or `OpenBLAS_jll` ([OpenBLAS](https://www.openblas.net/)). For the LP64 interface (32-bit integers) of LAPACK, use `LAPACK32_jll` and either `ReferenceBLAS32_jll` or `OpenBLAS32_jll`.
```
giordano, line 112: Actually, ideally one would link to libblastrampoline.

mattsignorelli: Ok, so then I feel that somewhere in the documentation for BinaryBuilder, new Julia/BinaryBuilder users should be pointed to `libblastrampoline` for LAPACK/BLAS linking, and also more specific steps on how to use it and compile it properly on all platforms. Specifically, I feel the documentation/FAQs should explicitly state the modifications that someone needs to make to the `CMakeLists.txt`, and the build script itself too.

For example, using this [`build_tarballs.jl`](https://github.com/JuliaPackaging/Yggdrasil/pull/7737/files#) as a reference, on Windows there is some name changing that the build script needs to do to find `libblastrampoline`. A small blurb in the documentation stating this could save a lot of time/pain.

And in cases where modifying the `Makefile` or `CMakeLists.txt` to instead link to `libblasttrampoline` is painful/inconvenient/etc, the solution I put above is a nice fallback, at least. See the [`build_tarballs.jl`](https://github.com/JuliaPackaging/Yggdrasil/pull/7829/commits/b0dbbe6a6bb18cc8a20199fa390295d545757d46) for a library I've been working on, where the build script is completely trivial - just calling `cmake`, no name changing, and also no changes to the `CMakeLists.txt`.

Speaking from personal experience, one of my first tasks as new Julia user for my work was to wrap a relatively simple C library which has only LAPACK as a dependency. I feel the process was more painful than it needed to be, and a simple blurb such as the one above in the FAQ would have saved me a lot of time. Initially, even after I got it compiled, and all of the tests passed and merged into Yggdrasil, a user of my package-in-development, which uses the _jll package, found that `using` the package on Windows threw an error, even though it worked on MacOS and Linux. After some investigating I found this was all because of having `LAPACK32_jll` with `ReferenceBLAS_jll`, instead of `ReferenceBLAS32_jll`, which I wasn't aware existed. And so small blurb in the FAQ I feel could really help new users with LAPACK linking with minimal pain.

mattsignorelli:
How about something like this?

giordano:
It's much more complicated than that.  I'll open another PR with more info

PR #1315, opened 2024-02-15
allowing use of fine-grained tokens

### LINKED ISSUE #1281: Deploying to a private repo
Hey all,

I'm trying to find a way to deploy to a private repo. A solution [has been already published here](https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1101), involving password-less ssh, which I'm also considering.

I've tried a different approach in the meanwhile, using the GitHub token. The first issue I've found is that t[he Wizard only requests access to public repositories](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/673c93e0c98efee166a70d1781a589059ae1f654/src/wizard/github.jl#L47), but this seems to be undocumented.
The second issue is that there is an expectation that [an existing github token must be of 40 characters](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/673c93e0c98efee166a70d1781a589059ae1f654/src/wizard/github.jl#LL8C1-L9C146), what's the reason? When setting a non 40 chars token, the failure was also "silent", it does not print anything in the log. Furthermore, in my case the token consisted of 93 characters  and it worked fine.

Work-around to skip the check on the token's length:
```
using BinaryBuilder
using GitHub

github_token = xxx
BinaryBuilder.Wizard._github_auth[] = GitHub.authenticate(github_token)
push!(ARGS, "--verbose", "--deploy=user/private_repo")

# Run the script
include("build-tarballs.jl")
```

epolack:
Losing-up the check on token size to allow for new protocol.

Fix #1281.

### DIFF
--- a/src/wizard/github.jl
+++ b/src/wizard/github.jl
@@ -6,7 +6,7 @@ const _github_auth = Ref{GitHub.Authorization}()
 
 function github_auth(;allow_anonymous::Bool=true)
 
-    if (!isassigned(_github_auth) || !allow_anonymous && isa(_github_auth[], GitHub.AnonymousAuth)) && length(get(ENV, "GITHUB_TOKEN", "")) == 40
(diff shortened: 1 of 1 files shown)


giordano:
Thanks!

PR #1324, opened 2024-04-28
Prevent stdlib compat entries from breaking old Julia versions

DilumAluthge:
In old Julia versions, `Pkg.test` sets the version numbers of all stdlibs to `0.0.0`. So, if we don't allow `0.0.0`, then `Pkg.test` will throw a resolver error.

Example of this breaking: https://github.com/JuliaRegistries/RegistryCI.jl/pull/557 (e.g. [this CI log](https://github.com/JuliaRegistries/RegistryCI.jl/actions/runs/8869626132/job/24350502653?pr=557)).

See also:
1. https://github.com/JuliaBinaryWrappers/MozillaCACerts_jll.jl/blob/main/Project.toml
2. https://github.com/JuliaRegistries/General/blob/master/jll/M/MozillaCACerts_jll/Compat.toml
3. https://github.com/JuliaRegistries/General/blob/master/jll/M/MozillaCACerts_jll/Versions.toml


giordano:
It just occurred to me that when this goes live in Yggdrasil, this will make all new builds of packages with the same X.Y.Z version number fail in the registry because the compat bound will have changed.  I'm not particularly thrilled by having to manually fix all of them.

PR #1337, opened 2024-07-10
[AutoBuild] Make `julia_compat` a required keyword argument

giordano:
Completely untested, will probably need to adapt some tests, but opening the PR to save my quick progress.  Ref: https://github.com/JuliaPackaging/Yggdrasil/issues/9046#issuecomment-2220313199.  CC: @vchuravy

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -6,7 +6,7 @@ using RegistryTools
 import LibGit2
 import PkgLicenses
 
-const DEFAULT_JULIA_VERSION_SPEC = "1.0"
+const DEFAULT_JULIA_VERSION_SPEC = "nothing"
 const DEFAULT_JLLWRAPPERS_VERSION_SPEC = "1.2.0"
 const PKG_VERSIONS = Base.VERSION >= v"1.7-" ? Pkg.Versions : Pkg.Types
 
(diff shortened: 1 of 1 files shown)


src/AutoBuild.jl
```diff
@@ -191,6 +192,14 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
         error("Package name \"$(src_name)\" is not a valid identifier")
     end
 
+    if julia_compat == DEFAULT_JULIA_VERSION_SPEC
+        error("""
+              `julia_compat` must be set explicitly to a valid compat spec string.
+              If this is an existing recipe which didn't previously specify `julia_compat`, use the old default value of `"1.0"`,
+              otherwise for a new recipe you're encouraged to set `julia_compat="1.6"`.
```
imciner2, line 199: Should we make this more generic (e.g. say to the current LTS release)? That way we don't need to update the error messages for newer versions.

giordano: I think sticking to v1.6 is ok, we don't have any particular reason to restrict to newer versions. Maybe I could rephrase slightly to say something along the lines of "at least 1.6"

ViralBShah:
LTS may be too aggressive - and perhaps LTS-1 (1.6) is a good default.

PR #1344, opened 2024-09-12
Check for a branded ELF note when OS/ABI is NONE

ararslan:
Currently the auditor is only checking for the correct OS/ABI and producing a warning on FreeBSD when it doesn't match. This warning is incorrect on AArch64 though, as FreeBSD instead uses an ELF note to declare the OS/ABI. The change implemented here looks for such a note when the OS/ABI in the ELF header is NONE.

~~Note that there are no tests added... I'd be interested in any suggestions for how best to test this given that the functions I modified are not currently tested, at least directly (AFAICT).~~

### DIFF
--- a/src/auditor/dynamic_linkage.jl
+++ b/src/auditor/dynamic_linkage.jl
@@ -2,7 +2,7 @@ using ObjectFile.ELF
 
 function os_from_elf_note(oh::ELFHandle)
     for section in Sections(oh)
-        section_handle(section) == ELF.SHT_NOTE || continue
+        section_type(section) == ELF.SHT_NOTE || continue
         seek(oh, section_offset(section))
         name_length = read(oh, UInt32)
         iszero(name_length) && continue
--- a/src/auditor/dynamic_linkage.jl
+++ b/src/auditor/dynamic_linkage.jl
@@ -1,5 +1,30 @@
 using ObjectFile.ELF
 
+function os_from_elf_note(oh::ELFHandle)
+    for section in Sections(oh)
+        section_handle(section) == ELF.SHT_NOTE || continue
+        seek(oh, section_offset(section))
+        name_length = read(oh, UInt32)
+        iszero(name_length) && continue
+        descriptor_length = read(oh, UInt32)
+        note_type = read(oh, UInt32)
+        name = String(read(oh, name_length - 1))  # skip trailing NUL
+        if note_type == 1
+            # Technically it's part of the Linux specification that any executable should
+            # have an ELF note with type 1, name GNU, and descriptor length ≥4, but in
+            # practice I haven't observed that consistently, especially on musl. So for
+            # now, only bother checking FreeBSD, which uses an ELF note rather than OS/ABI
+            # to identify itself on AArch64 and RISC-V.
+            if name == "FreeBSD" && descriptor_length == 4
+                return name
+            end
+        end
+    end
+    return nothing
+end
+
+os_from_elf_note(::ObjectHandle) = nothing
+
 """
     platform_for_object(oh::ObjectHandle)
 
@@ -39,7 +64,9 @@ function platform_for_object(oh::ObjectHandle)
             end
         end
 
-        if oh.ei.osabi == ELF.ELFOSABI_LINUX || oh.ei.osabi == ELF.ELFOSABI_NONE
+        if oh.ei.osabi == ELF.ELFOSABI_NONE
+            return Platform(arch, os_from_elf_note(oh) == "FreeBSD" ? "freebsd" : "linux")
+        elseif oh.ei.osabi == ELF.ELFOSABI_LINUX
             return Platform(arch, "linux")
         elseif oh.ei.osabi == ELF.ELFOSABI_FREEBSD
             return Platform(arch, "freebsd")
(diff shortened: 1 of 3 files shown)


giordano:
I haven't looked at the implementation yet, but I want to quickly answer the question

> Note that there are no tests added... I'd be interested in any suggestions for how best to test this given that the functions I modified are not currently tested, at least directly (AFAICT).

Yeah, the problem is that we don't know (or forgot?  who knows) how libraries with the wrong OS/ABI field were generated on FreeBSD. An alternative should be to compile a simple dummy library (something like `echo '' | cc -x c - -shared -fPIC -o libdummy.${dlext}`) and then `patchelf --set-os-abi` it

src/auditor/extra_checks.jl
```diff
@@ -5,7 +5,19 @@ using Dates: DateTime, datetime2unix
 
 function check_os_abi(oh::ObjectHandle, p::AbstractPlatform, rest...; verbose::Bool = false, kwargs...)
     if Sys.isfreebsd(p)
-        if oh.ei.osabi != ELF.ELFOSABI_FREEBSD
+        # On AArch64 and RISC-V, FreeBSD uses an ELF note section to identify itself rather
+        # than OS/ABI in the ELF header. In that case, the OS/ABI will be generic Unix (NONE).
+        # See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=252490 and
+        # https://github.com/freebsd/freebsd-src/blob/main/lib/csu/common/crtbrand.S
```
giordano, line 11:
```suggestion
        # https://github.com/freebsd/freebsd-src/blob/3c95262007ef934c9e98b87460a48889bf42c1b9/lib/csu/common/crtbrand.S
```

src/auditor/dynamic_linkage.jl
```diff
@@ -39,7 +64,9 @@ function platform_for_object(oh::ObjectHandle)
             end
         end
 
-        if oh.ei.osabi == ELF.ELFOSABI_LINUX || oh.ei.osabi == ELF.ELFOSABI_NONE
+        if oh.ei.osabi == ELF.ELFOSABI_NONE
```
giordano, line 67: Should we just delete `platform_for_object` since it appears to be unused?

ararslan: Sure, I can if you'd like me to.

ararslan:
Ah rip, the patchelf we ship is old and doesn't have `--set-os-abi`

giordano:
Sigh, never mind the test then, we can't upgrade patchelf because newer version is broken (https://github.com/JuliaPackaging/Yggdrasil/pull/7728#issuecomment-1830964109, caused by https://github.com/NixOS/patchelf/issues/492)

test/auditing.jl
```diff
@@ -788,57 +789,113 @@ end
...
+                    v"4.2.0",
+                    # No sources
+                    FileSource[],
+                    # Build a library with a mismatched OS/ABI in the ELF header
+                    raw"""
+                    mkdir -p "${libdir}"
+                    echo 'int wrong() { return 0; }' | cc -shared -fPIC -o "${libdir}/libwrong.${dlext}" -x c -
+                    echo 'int right() { return 0; }' | cc -shared -fPIC -o "${libdir}/libright.${dlext}" -x c -
+                    # NetBSD runs anywhere, which implies that anything that runs is for NetBSD, right?
+                    patchelf --set-os-abi 2 "${libdir}/libwrong.${dlext}"
```
ararslan, line 862:
```suggestion
                    raw"""
                    apk update
                    apk add binutils
                    mkdir -p "${libdir}"
                    echo 'int wrong() { return 0; }' | cc -shared -fPIC -o "${libdir}/libwrong.${dlext}" -x c -
                    echo 'int right() { return 0; }' | cc -shared -fPIC -o "${libdir}/libright.${dlext}" -x c -
                    # NetBSD runs anywhere, which implies that anything that runs is for NetBSD, right?
                    elfedit --output-osabi=NetBSD "${libdir}/libwrong.${dlext}"
```
I wonder if this will work

ararslan:
Finally, 168 days after this PR was first opened, it actually works, is tested, and passes CI.

src/auditor/dynamic_linkage.jl
ararslan: Actually, unless you feel strongly, I'd prefer to do that separately from this PR since it's an unrelated change.

src/auditor/dynamic_linkage.jl
```diff
@@ -1,5 +1,30 @@
...
+            if name == "FreeBSD" && descriptor_length == 4
+                return name
+            end
+        end
+    end
+    return nothing
+end
+
+os_from_elf_note(::ObjectHandle) = nothing
+
```
giordano, line 27: Good to know a blank line was not covered by tests!

image

src/auditor/dynamic_linkage.jl
giordano: Fine by me.

src/auditor/extra_checks.jl
giordano: Bump :slightly_smiling_face:

src/auditor/dynamic_linkage.jl
ararslan:
```julia
@test isempty(readlines("src/auditor/dynamic_linkage.jl")[27])
```

src/auditor/extra_checks.jl
ararslan: Ah thanks for the bump, I missed this comment somehow

ararslan:
Bump :)

giordano (approved):

PR #1363, opened 2025-02-08
[AutoBuild] Replaced platform triplet string comparison with platform comparison in rebuild_jll_package

stemann:
Should remedy issues seen with registration of ONNXRuntime_CUDA (using the `cuda_platform` platform tag, e.g., `cuda_platform="jetson"`) in https://github.com/JuliaPackaging/Yggdrasil/pull/10426, and investigated in https://github.com/JuliaPackaging/Yggdrasil/pull/10476 .

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1171,8 +1171,25 @@ function rebuild_jll_package(name::String, build_version::VersionNumber, sources
     build_output_meta = Dict()
 
     # For each platform, we have two tarballs: the main with the full product,
-    # and the logs-only one.  This function filters out the logs one.
-    filter_main_tarball(f, platform) = occursin(".$(triplet(platform)).tar", f) && !occursin("-logs.", f)
+    # and the logs-only one.  This function filters out the logs one, and
+    # finds the tarball matching `platform`.
+    function filter_main_tarball(tarball_filename, platform)
+        if occursin("-logs.", tarball_filename)
+            return false
+        end
+        tarball_filename_match = match(r"^(?<name>[\w_]+)\.v(?<version>\d+\.\d+\.\d+)\.(?<platform_triplet>([^-]+-?)+).tar", tarball_filename)
+        if isnothing(tarball_filename_match)
+            @warn "Tarball filename does not match expected pattern: $(tarball_filename)"
+            return false
+        end
+        try
+            tarball_filename_platform = parse(Platform, tarball_filename_match[:platform_triplet])
+            return tarball_filename_platform == platform
+        catch
+            @warn "Failed to parse tarball filename: $(tarball_filename)"
+            return false
+        end
+    end
 
     # Then generate a JLL package for each platform
     downloaded_files = readdir(download_dir)
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1161,6 +1161,27 @@ function rebuild_jll_package(obj::Dict;
     )
 end
 
+# For each platform, we have two tarballs: the main with the full product,
+# and the logs-only one.  This function filters out the logs one, and
+# finds the tarball matching `platform`.
+function filter_main_tarball(tarball_filename, platform)
+    if occursin("-logs.", tarball_filename)
+        return false
+    end
+    tarball_filename_match = match(r"^(?<name>[\w_]+)\.v(?<version>\d+\.\d+\.\d+)\.(?<platform_triplet>([^-]+-?)+).tar", tarball_filename)
+    if isnothing(tarball_filename_match)
+        @warn "Tarball filename does not match expected pattern: $(tarball_filename)"
+        return false
+    end
+    try
+        tarball_filename_platform = parse(Platform, tarball_filename_match[:platform_triplet])
+        return tarball_filename_platform == platform
+    catch
+        @warn "Failed to parse tarball filename: $(tarball_filename)"
+        return false
+    end
+end
+
 function rebuild_jll_package(name::String, build_version::VersionNumber, sources::Vector,
                              platforms::Vector, products::Vector, dependencies::Vector,
                              download_dir::String, upload_prefix::String;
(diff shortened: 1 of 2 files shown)


stemann:
Might be an idea to look into a test or two…

src/AutoBuild.jl
```diff
@@ -1171,8 +1171,25 @@ function rebuild_jll_package(name::String, build_version::VersionNumber, sources
...
+            return false
+        end
+        tarball_filename_match = match(r"^(?<name>[\w_]+)\.v(?<version>\d+\.\d+\.\d+)\.(?<platform_triplet>([^-]+-?)+).tar", tarball_filename)
+        if isnothing(tarball_filename_match)
+            @warn "Tarball filename does not match expected pattern: $(tarball_filename)"
+            return false
+        end
+        try
+            tarball_filename_platform = parse(Platform, tarball_filename_match[:platform_triplet])
+            return tarball_filename_platform == platform
```
giordano, line 1187: One thing I'm not 100% sure is whether this comparison is fine. Platform equality is finicky, I usually prefer using platform_match, but probably in this case we _really_ want to check exact equality specifically, so this may be good.

giordano (commented):
I haven't looked at the code very closely yet, but at a high level this looks very promising, thanks for doing this.

> Might be an idea to look into a test or two…

Yes, the [registration pipeline](https://github.com/JuliaPackaging/Yggdrasil/blob/2dbe3b7b29aff2272567071ada82024883d46102/.ci/register_package.jl) is largely untested here, because it's also complicated to reproduce end-to-end without a mock registry and such. However this is a small self-contained function which should be easy to unit test, so please do go ahead and add tests!

src/AutoBuild.jl
stemann: Good points - both of them. I’ll check up on `platform_match`.

giordano:
```
julia> using Base.BinaryPlatforms

julia> p1 = Platform("x86_64", "linux"; tag1="foo", tag3="blah")
Linux x86_64 {libc=glibc, tag1=foo, tag3=blah}

julia> p2 = Platform("x86_64", "linux"; tag2="bar", tag4="glah")
Linux x86_64 {libc=glibc, tag2=bar, tag4=glah}

julia> platforms_match(p1, p2)
true

julia> p3 = Platform("x86_64", "linux"; tag3="blah", tag1="foo")
Linux x86_64 {libc=glibc, tag1=foo, tag3=blah}

julia> p4 = Platform("x86_64", "linux"; tag4="glah", tag2="bar")
Linux x86_64 {libc=glibc, tag2=bar, tag4=glah}

julia> platforms_match(p1, p3)
true

julia> p1 == p3
true

julia> p2 == p4
true
```
Probably equality is better than platform_match here, but make sure with appropriate tests that it works correctly also in complex cases.

stemann: I have added some tests.

I would argue that the functionality of the `Platform` constructor should not be tested here (p1 vs p3, and p2 vs p4), but the difference between `Platform` equality and `platforms_match` should, e.g. where `platforms_match` is true and equality is false,
```
julia> p5 = Platform("x86_64", "linux")
Linux x86_64 {libc=glibc}

julia> p6 = Platform("x86_64", "linux"; cxxstring_abi="cxx11")
Linux x86_64 {cxxstring_abi=cxx11, libc=glibc}

julia> platforms_match(p5, p6)
true

julia> p5 == p6
false
```

stemann:
It should be noted that I did not get concrete evidence that this is indeed the root cause for the issues in https://github.com/JuliaPackaging/Yggdrasil/pull/10426 etc., but it seems like a (very) likely cause. The issue was primarily that `triplet(platform)` put `cuda_platform` before `cuda`:

With some quasi-empty tarballs in "./foo":
```julia
platforms = expand_cxxstring_abis([
    Platform("aarch64", "linux"; cuda="10.2", cuda_platform="jetson"),
    Platform("x86_64", "linux"; cuda="11.3"),
    Platform("x86_64", "windows"; cuda="11.3")
]; skip=!Sys.islinux)
# 5-element Vector{Platform}:
#  Linux aarch64 {cuda=10.2, cuda_platform=jetson, cxxstring_abi=cxx03, libc=glibc}
#  Linux aarch64 {cuda=10.2, cuda_platform=jetson, cxxstring_abi=cxx11, libc=glibc}
#  Linux x86_64 {cuda=11.3, cxxstring_abi=cxx03, libc=glibc}
#  Linux x86_64 {cuda=11.3, cxxstring_abi=cxx11, libc=glibc}
#  Windows x86_64 {cuda=11.3}

readdir(joinpath(pwd(), "foo"))
# ONNXRuntime_CUDA.v1.10.0.aarch64-linux-gnu-cxx03-cuda_platform+jetson-cuda+10.2.tar.gz
# ONNXRuntime_CUDA.v1.10.0.aarch64-linux-gnu-cxx11-cuda_platform+jetson-cuda+10.2.tar.gz
# ONNXRuntime_CUDA.v1.10.0.x86_64-linux-gnu-cxx03-cuda+11.3.tar.gz
# ONNXRuntime_CUDA.v1.10.0.x86_64-linux-gnu-cxx11-cuda+11.3.tar.gz
# ONNXRuntime_CUDA.v1.10.0.x86_64-w64-mingw32-cuda+11.3.tar.gz

triplet(first(platforms))
# "aarch64-linux-gnu-cxx03-cuda_platform+jetson-cuda+10.2"

BinaryBuilder.rebuild_jll_package("foo", v"0.1.0", [], platforms, [], [], joinpath(pwd(), "foo"), "foo"; verbose=true)
# Runs past finding platform for each file in "./foo"
```
I have also not discovered why the triplets sometimes had the `cuda` and `cuda_platform` tags in another order. ... perhaps an issue in Base.BinaryPlatforms...(?)

giordano:
Is this ready?

> I have also not discovered why the triplets sometimes had the `cuda` and `cuda_platform` tags in another order.

`Dict`s are not sorted, so that's totally expected.

stemann:
OK.

Yes, I think it is ready.

PR #1364, opened 2025-02-09
[Auditor] Make the auditor multithreaded

giordano:
Completely untested, but saving my progress for now.  Let's see how far we get with this.


giordano:
Good news: building locally https://github.com/JuliaPackaging/Yggdrasil/pull/10479 (with hot ccache 🔥) goes from
```
[ Info: Timings: setup: 28.0s, build: 2m 14.22s, audit: 21m 23.27s, packaging: 12.44s
```
to
```
[ Info: Timings: setup: 27.72s, build: 2m 15.6s, audit: 8m 57.83s, packaging: 12.35s
```
when using 16 threads.  Clearly scaling isn't ideal, but we have to put locks around all uses of the sanbox because otherwise it's a mess, but a 2x speedup is more than welcome.

Bad news: we need to put more locks around all logging macros because they aren't thread-safe and printing is messy, but hopefully they won't slow things down massively.

Note to self: we may be able to replace calls to `patchelf` at https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/edecf34fb8c7e8f7791b5e79116e701eb7a9274b/src/auditor/soname_matching.jl#L74 and https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/edecf34fb8c7e8f7791b5e79116e701eb7a9274b/src/auditor/dynamic_linkage.jl#L354 with the JLL, which removes the need to use locks in a couple of places.

giordano:
With `Threads.@spawn` instead of `Threads.@threads` (note: I haven't pushed it yet) I get marginally better timings, thanks to slightly more efficient load balancing:
```
[ Info: Timings: setup: 24.97s, build: 2m 7.83s, audit: 7m 27.03s, packaging: 12.16s
```
If CI is happy, this should be in a good shape.

src/auditor/dynamic_linkage.jl
```diff
@@ -1,4 +1,5 @@
 using ObjectFile.ELF
+using Patchelf_jll: patchelf
```
giordano, line 2: I see BB2 is already using patchelf, which is great: https://github.com/JuliaPackaging/BinaryBuilder2.jl/blob/54202516e430264dc6060f763c55a7f0fbbd542a/BinaryBuilderAuditor.jl/src/Scanning.jl#L1.  But at a quick glance I don't see multi-threading enabled yet.

giordano:
CI is happy now, let's see how it goes in production

PR #1365, opened 2025-02-10
[Auditor] Use executables from `Binutils_jll` when possible

giordano:


src/auditor/instruction_set.jl
```diff
@@ -36,12 +37,17 @@ function instruction_mnemonics(path::AbstractString, platform::AbstractPlatform)
     output = IOBuffer()
 
     # Run objdump to disassemble the input binary
-    if Sys.isbsd(platform)
-        objdump_cmd = "llvm-objdump -d $(basename(path))"
+    objdump_args = `-d $(basename(path))`
+    if Binutils_jll.is_available() && (Sys.islinux(platform) || Sys.isfreebsd(platform))
+        run(ignorestatus(pipeline(`$(Binutils_jll.objdump()) $(objdump_args)`; stdout=output, stderr=devnull)))
```
giordano, line 42: Running `objdump` can take a long time, depending on the size of binary file.  If this works, removing a lock acquisition here can be very beneficial for parallelism in presence of several object files to analyze.

giordano: It doesn't work, sigh.

PR #1370, opened 2025-02-25
Fix `filter_main_tarball`

lgoettgens:
This should fix the issue encountered in https://github.com/JuliaPackaging/Yggdrasil/pull/10611#issuecomment-2681599438.

There were actually two issues that are fixed by this PR:
- `filter_main_tarball` gets a path and not a filename as the first arg.
- the triplet parsing fails for `any`

I don't think there is a way to test it apart from somehow getting this version into the yggdrasil toolchain and re-running CI on https://github.com/JuliaPackaging/Yggdrasil/commit/3751eecbc3120c92219aaf056bf8d1bf2511bc9c again. (Which would be great to have anyway to get the GAP_lib_jll release from there into the registry.)

cc @fingolfin @giordano

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1168,7 +1168,7 @@ function filter_main_tarball(tarball_filename, platform)
     if occursin("-logs.", tarball_filename)
         return false
     end
-    tarball_filename_match = match(r"^(?<name>[\w_]+)\.v(?<version>\d+\.\d+\.\d+)\.(?<platform_triplet>([^-]+-?)+).tar", tarball_filename)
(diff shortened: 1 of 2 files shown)


giordano:
Thanks for fixing this!

giordano:
> and re-running CI on https://github.com/JuliaPackaging/Yggdrasil/commit/3751eecbc3120c92219aaf056bf8d1bf2511bc9c again.

I'm afraid that's not possible, the version of BinaryBuilder is taken from the checked in registry in each revision (which I just updated in https://github.com/JuliaPackaging/Yggdrasil/pull/10617), you'll have to trigger again a new build.

PR #1371, opened 2025-02-27
Update build_tips.md

asinghvi17:
Adds brief descriptions of how to do "bare" (no build system) C and C++ builds - just to show where things should go.

### DIFF
--- a/docs/src/build_tips.md
+++ b/docs/src/build_tips.md
@@ -127,6 +127,34 @@ Example of packages using Rust:
 
 	The Rust toolchain currently used does not work with the `i686-w64-mingw32` (32-bit Windows) platform.
 
+ ## C builds
+
+ If your library has no build system like Make, CMake, Meson, or Autoconf, you may need to use the C compiler directly.  The C compiler is stored in the `CC` environment variable, and you can direct output to `libdir` (shared libraries) and `bindir` (executables).
+
+ ```sh
+# this assumes you are operating out of a Git source named `hello`
+# adjust your `cd` appropriately
+cd $WORKSPACE/srcdir/hello 
+mkdir -p ${libdir}                                         # make sure the libdir is instantiated
+${CC} -shared -o ${libdir}/libhello.${dlext} -fPIC hello.c # compile the library, save to `libdir`
+```
+
+This is simply compiling a single shared library, `libhello`, from `hello.c`.
+
+## C++ builds
+
+Similarly to C builds, sometimes your C++ libraries will not have a build system associated with them.  The C++ compiler is stored in the `CXX` environment variable.
+
+```sh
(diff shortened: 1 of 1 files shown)


fingolfin (approved):
Seems fine to me

Out of curiosity, what is the immediate motivation for this, did you have to package something for Yggdrasil were this was necessary?

docs/src/build_tips.md
```diff
@@ -127,6 +127,34 @@ Example of packages using Rust:
...
+## C++ builds
+
+Similarly to C builds, sometimes your C++ libraries will not have a build system associated with them.  The C++ compiler is stored in the `CXX` environment variable.
+
+```sh
+# this assumes you are operating out of a Git source named `hello`
+# adjust your `cd` appropriately
+cd $WORKSPACE/srcdir/hello
+mkdir -p ${libdir}
+$CXX -shared -std=c++11 -O3 -fPIC -o ${libdir}/libhello.${dlext} src/hello.cpp # you may want to edit the `std` flag, for exampLE
```
giordano, line 153:
```suggestion
$CXX -shared -std=c++11 -O3 -fPIC -o ${libdir}/libhello.${dlext} src/hello.cpp # you may want to edit the `std` flag, for example
```

docs/src/build_tips.md
```diff
@@ -127,6 +127,34 @@ Example of packages using Rust:
 
 	The Rust toolchain currently used does not work with the `i686-w64-mingw32` (32-bit Windows) platform.
 
+ ## C builds
+
+ If your library has no build system like Make, CMake, Meson, or Autoconf, you may need to use the C compiler directly.  The C compiler is stored in the `CC` environment variable, and you can direct output to `libdir` (shared libraries) and `bindir` (executables).
+
```
giordano, line 133: I don't want to give the impression that this are the exact commands one shall run, it's rather to give an idea (I can't say how many times people take these instructions almost verbatim rather than as indication)
```suggestion
As a high-level example:

```

docs/src/build_tips.md
```diff
@@ -127,6 +127,34 @@ Example of packages using Rust:
...
+mkdir -p ${libdir}                                         # make sure the libdir is instantiated
+${CC} -shared -o ${libdir}/libhello.${dlext} -fPIC hello.c # compile the library, save to `libdir`
+```
+
+This is simply compiling a single shared library, `libhello`, from `hello.c`.
+
+## C++ builds
+
+Similarly to C builds, sometimes your C++ libraries will not have a build system associated with them.  The C++ compiler is stored in the `CXX` environment variable.
+
```
giordano, line 147:
```suggestion
As a high-level example:

```

asinghvi17:
Yeah, I was building the tg library which is completely self contained, no build tools at all

Issue #1241, opened 2022-11-02
Artifact download URL when using --deploy=local is wrong


aviks:
When running `build_tarballs.jl` with a `--deploy=local` option, the resulting `Artifact.toml` has the artifact download url set to something like this: `https://github.com/local/releases/download/MyPkg-v0.2.1+1/MyPkg.v0.2.1.aarch64-apple-darwin.tar.gz`

Now the right value here is *unkown* so it's not clear what should happen. Maybe some more affordances to let the user know that this value should be changed? `https://github.com/CHANGE/ME/releases/downloads....` .  The current value is plausible enough to be dangerous, IMO.

PR #1378, opened 2025-04-21
fix artifacts download url when `--deploy=local`

t-bltg:
Fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1241, which I also hit while working on https://github.com/JuliaPackaging/Yggdrasil/pull/11052.

I also took the liberty of removing the exported symbol `print_artifacts_toml` since it appeared nowhere, should we do the same for [`build`](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/2a5e7117be9ade099cd1092e7a444bdcffc37500/src/AutoBuild.jl#L1) ?

Added test.

Locally, the artifact is correctly downloaded and used after a local build:
`julia build_tarballs.jl x86_64-linux-gnu --verbose --deploy=local`, and tested in the `GR` test suite with the resulting `GR_jll` dev'ed.


t-bltg:
@giordano for a review, maybe ?

giordano (approved):

PR #1380, opened 2025-05-06
Remove `tzcnt` from audited BMI1 instructions

topolarity:
By happy-accident, older CPU's decode this instruction in backward-compatible way for non-zero inputs [^1], so it can be used even without feature support. In fact, compilers will emit this instruction in some circumstances even when BMI1 is not technically available on your architecture.

I hit this on https://github.com/JuliaPackaging/Yggdrasil/pull/11124

[^1]: https://stackoverflow.com/questions/61422827/does-x64-support-imply-bmi1-support


giordano:
It'd be better to update the script https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/2a5e7117be9ade099cd1092e7a444bdcffc37500/contrib/generate_instructions_list.jl which generates that file, otherwise this change risks of being overwritten.

giordano (approved):
Thanks!

PR #1383, opened 2025-05-13
Remove `.pkg` if no longer augmenting

vchuravy:
@maleadt removed augmentation recently from pocl_jll 

https://github.com/JuliaBinaryWrappers/pocl_jll.jl/commit/cfcdb23a62c76e835704c047753a0a69dae9e041

But the `.pkg` dir remained confusing things down the road.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1561,7 +1561,7 @@ function build_jll_package(src_name::String,
               """)
     else
         # If no augmentation exists, ensure that the pkg_dir get's removed again.
-        rm(pkg_dir, recursive=true), force=true
+        Base.rm(pkg_dir, recursive=true, force=true)
     end
 
     # Generate target-demuxing main source file.
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -1560,7 +1560,7 @@ function build_jll_package(src_name::String,
               TOML.print(stdout, artifacts)
               """)
     else
-        # If no augmentation exists, ensure that the pkg_dir get's removed again.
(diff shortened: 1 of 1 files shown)


vchuravy:
x-ref: https://github.com/JuliaPackaging/Yggdrasil/pull/11185#issuecomment-2877680027

giordano:
> This may point to a larger issue, we should perhaps remove the content of the git repository before regenerating the files.

Uhm, that'd break the case where we call build_tarballs multiple times.

src/AutoBuild.jl
```diff
@@ -1559,6 +1559,9 @@ function build_jll_package(src_name::String,
               # Output the result to `stdout` as a TOML dictionary
               TOML.print(stdout, artifacts)
               """)
+    else
+        # If no augmentation exists, ensure that the pkg_dir get's removed again.
+        rm(pkg_dir, recursive=true), force=true
```
vchuravy, line 1564:
```suggestion
        Base.rm(pkg_dir, recursive=true, force=true)
```

src/AutoBuild.jl
```diff
@@ -1559,6 +1559,9 @@ function build_jll_package(src_name::String,
               # Output the result to `stdout` as a TOML dictionary
               TOML.print(stdout, artifacts)
               """)
+    else
+        # If no augmentation exists, ensure that the pkg_dir get's removed again.
```
giordano, line 1563:
```suggestion
        # If no augmentation exists, ensure that the pkg_dir gets removed again.
```

giordano (approved):

PR #1389, opened 2025-07-05
fix `libgfortran` version in auditor test

t-bltg:
The official `1.7.3` version of `julia` for the `x86_64-linux-gnu` triplet ships with `libgfortran.so.4`:
```julia
julia> VERSION
v"1.7.3"
julia> using Libdl
julia> foreach(println, Libdl.dllist())
linux-vdso.so.1
/lib/x86_64-linux-gnu/libdl.so.2
/lib/x86_64-linux-gnu/libpthread.so.0
/lib/x86_64-linux-gnu/libc.so.6
[...]/bin/../lib/libjulia.so.1
/lib64/ld-linux-x86-64.so.2
[...]/bin/../lib/julia/libgcc_s.so.1
[...]/bin/../lib/julia/libopenlibm.so
[...]/bin/../lib/julia/libjulia-internal.so.1
[...]/bin/../lib/julia/libunwind.so.8
[...]/bin/../lib/julia/libLLVM-12jl.so
/lib/x86_64-linux-gnu/librt.so.1
[...]/bin/../lib/julia/libatomic.so.1
[...]/bin/../lib/julia/libstdc++.so.6
/lib/x86_64-linux-gnu/libm.so.6
[...]/bin/../lib/julia/libz.so.1
[...]/lib/julia/sys.so
[...]/bin/../lib/julia/libpcre2-8.so
[...]/bin/../lib/julia/libgmp.so
[...]/bin/../lib/julia/libmpfr.so
[...]/bin/../lib/julia/libopenblas64_.so
[...]/bin/../lib/julia/libgfortran.so.4  # <== here !
[...]/bin/../lib/julia/libquadmath.so.0
[...]/bin/../lib/julia/libblastrampoline.so
[...]/bin/../lib/julia/libcholmod.so
[...]/bin/../lib/julia/libamd.so.2
[...]/bin/../lib/julia/libcolamd.so.2
[...]/bin/../lib/julia/libsuitesparseconfig.so.5
[...]/bin/../lib/julia/libccolamd.so.2
[...]/bin/../lib/julia/libcamd.so.2
[...]/bin/../lib/julia/libnghttp2.so.14
[...]/bin/../lib/julia/libcurl.so
[...]/bin/../lib/julia/libssh2.so.1
[...]/bin/../lib/julia/libmbedtls.so.13
[...]/bin/../lib/julia/libmbedx509.so.1
[...]/bin/../lib/julia/libmbedcrypto.so.5
```

[...]

PR #1391, opened 2025-07-05
fix testsuite on `1.8.5` , `1.9.4`, `1.10.10`, `1.11.5`

t-bltg:
Fixes test suite on recent `julia` versions (tested on `1.8.5`, `1.9.4`, `1.10.10` and `1.11.5`).

This passes locally on all versions (at least on `x86_64-linux-gnu`):
```
$ JULIA_REGISTRYCI_AUTOMERGE=true julia --project=. -e 'using Pkg; Pkg.test()'
```

Needs https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1389.

cc @giordano

### DIFF
--- a/test/building.jl
+++ b/test/building.jl
@@ -437,7 +437,8 @@ end
         # Test that having a LibraryProduct for AnyPlatform raises an error.
         # Note that the error is raised during audit, which is run in threads,
         # so we raise an extremely generic `TaskFailedException`.
-        @test_throws TaskFailedException autobuild(
+        exc = VERSION < v"1.9" ? TaskFailedException : CompositeException
+        @test_throws exc autobuild(
             build_path,
             "libfoo",
             v"1.0.0",
--- a/test/wizard.jl
+++ b/test/wizard.jl
@@ -4,7 +4,7 @@ import Pkg: PackageSpec
 
 import BinaryBuilder.BinaryBuilderBase: available_gcc_builds, available_llvm_builds, getversion
 
-const debug = Ref(true)
+const debug = Ref(false)
 
 function with_wizard_output(f::Function, state, step_func::Function)
     # Create fake terminal to communicate with BinaryBuilder over
--- a/test/wizard.jl
+++ b/test/wizard.jl
@@ -4,6 +4,8 @@ import Pkg: PackageSpec
 
 import BinaryBuilder.BinaryBuilderBase: available_gcc_builds, available_llvm_builds, getversion
 
+const debug = Ref(true)
+
 function with_wizard_output(f::Function, state, step_func::Function)
     # Create fake terminal to communicate with BinaryBuilder over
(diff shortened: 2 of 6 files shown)


test/wizard.jl
```diff
@@ -86,38 +88,63 @@ for i in available_ports
...
-        if isempty(new_c)
-            # We need to wait for more data, sleep for a bit
-            sleep(0.01)
-            continue
-        end
-
-        buffer = [buffer[2:end]; new_c]
-        if !any(buffer .!= needle)
-            return all_buffer
+@static if true
```
giordano, line 91: Can you elaborate (ideally with a comment in the code) what's going on here?  Reading an `if true` is weird at best, without an explanation.

test/wizard.jl
```diff
@@ -86,38 +88,63 @@ for i in available_ports
...
+        # N.B.: This is a terrible way to do this and works around the fact that our `IOBuffer`
+        # does not block. It works fine here, but do not copy this to other places.
+        needle = codeunits(needle)
+        buffer = zeros(UInt8, length(needle))
+        all_buffer = UInt8[]
+        while isopen(io)
+            new_c = read(io, 1)
+            if isempty(new_c)
+                # We need to wait for more data, sleep for a bit
+                sleep(0.001)
```
giordano, line 102: Why reducing the wait?

t-bltg: I've restored the original `sleep` calls.

test/building.jl
```diff
@@ -437,7 +437,8 @@ end
         # Test that having a LibraryProduct for AnyPlatform raises an error.
         # Note that the error is raised during audit, which is run in threads,
         # so we raise an extremely generic `TaskFailedException`.
-        @test_throws TaskFailedException autobuild(
+        exc = VERSION < v"1.9" ? TaskFailedException : CompositeException
```
t-bltg, line 440: Fix 1 for new versions.

test/wizard.jl
```diff
@@ -110,14 +112,37 @@ function readuntil_sift(io::IO, needle)
...
+
+function succcess_path_call_response(ins, outs)
+    output = readuntil_sift(outs, "Build complete")
+    if contains(String(output), "Warning:")
+        close(ins)
+        return false
+    end
+    call_response(ins, outs, "Would you like to edit this script now?", "N")
+    # MultiSelectMenu (https://docs.julialang.org/en/v1/stdlib/REPL/#MultiSelectMenu)
+    needle = if VERSION < v"1.9"
```
t-bltg, line 137: Fix 2 for new versions.

t-bltg:
Sorry, I've clarified :point_up_2: the places that are actual fixes for new versions in order to ease the review.

The other changes were added in order to ease development and debugging by toggling `debug`.

PR #1394, opened 2025-07-12
remove dead code (`detect_libstdcxx_version` and `check_libstdcxx_version`) ?

t-bltg:
It looks like `BinaryBuilder.Auditor.detect_libstdcxx_version` and `BinaryBuilder.Auditor.check_libstdcxx_version` is dead code and not used anywhere in `BinaryBuilder` without any traces of usage anywhere else in the `julia` ecosystem:

https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/master/src/auditor/compiler_abi.jl#L107-L165

So should they be removed ?

Hoisted out of https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1392.

### DIFF
--- a/src/auditor/compiler_abi.jl
+++ b/src/auditor/compiler_abi.jl
@@ -1,6 +1,5 @@
-import Base.BinaryPlatforms: detect_libstdcxx_version, detect_cxxstring_abi
-using ObjectFile
 using Binutils_jll: Binutils_jll
+using ObjectFile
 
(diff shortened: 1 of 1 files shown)


giordano:
I never touched this code because I always assumed this was meant to be used at some point, since Julia itself also has mechanisms to detect the libstdc++ version (on Linux do `Base.BinaryPlatform.HostPlatform()`).

t-bltg:
Ok, I wanted to add tests covering these lines, but they were failing, so I'm not sure what to do here.

PR #1400, opened 2025-07-23
Fix using shipped stdlib depot

IanButterworth:
Without this BB has to precompile stdlibs

@giordano

### DIFF
--- a/.ci/agent_startup.conf
+++ b/.ci/agent_startup.conf
@@ -6,7 +6,7 @@ StartLimitBurst       = 0
 [Service]
 EnvironmentFile       = ${SRC_DIR}/.env
 Environment           = AGENT_IDX=${AGENT_IDX}
-Environment           = JULIA_DEPOT_PATH=${AGENT_DIR}/depot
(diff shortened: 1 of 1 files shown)


giordano:
Uhm, this is only relevant for the CI setup here which is still running on v1.7, no? What you're trying to fix?

giordano:
That's in Yggdrasil, presumably at https://github.com/JuliaPackaging/Yggdrasil/blob/49837563d4e5e87132a651644181c6396bd7e32d/.buildkite/build.sh#L9, not here

IanButterworth:
I fixed that here and it didn't solve the problem https://github.com/JuliaPackaging/Yggdrasil/pull/11683/files#diff-877fafdf17962f8430a77bec7d7a680291a6ebeee5929c0cb0f815eed85c9c54R10

IanButterworth:
It's logically sound, but it didn't fix the issue I was seeing. It likely helps somewhere avoid re-precompiling stdlibs, just not sure where.

PR #1410, opened 2025-11-28
Fix patchelf concurrency issues. Show subprocess output in errors

IanButterworth:
## Summary

This PR fixes two issues in the auditor:

1. **Improved error reporting**: When patchelf or other subprocess commands fail, the error now includes the subprocess output, making debugging much easier.

2. **Fixed race condition in patchelf operations**: The auditor uses nested `@threads` loops—iterating over files in parallel, and within each file, iterating over libraries in parallel. When multiple threads try to relink different libraries in the same binary, they call `patchelf` on the same file concurrently, corrupting the ELF headers and causing "Missing ELF header" errors.

## Why this wasn't an issue before

In Julia ≤1.7, nested `@threads` loops effectively serialized because the default `:static` scheduling pinned iterations to threads and didn't support nesting. Starting in Julia 1.8, `@threads` defaults to `:dynamic` scheduling ([JuliaLang/julia#43919](https://github.com/JuliaLang/julia/pull/43919), [#44136](https://github.com/JuliaLang/julia/pull/44136)), which enables true nested parallelism—exposing the latent race condition.

i.e. in 1.7 the inner `@threads` becomes a standard for loop.
julia +1.7 -t6
```
julia> Threads.@threads for i in 1:3
           @info "outer i$i - $(Threads.threadid())"
           Threads.@threads for j in 1:3
               @info "inner j$j i$i - $(Threads.threadid())"
           end
       end
[ Info: outer i1 - 1
[ Info: outer i2 - 2
[ Info: outer i3 - 3
[ Info: inner j1 i3 - 3
[ Info: inner j2 i3 - 3
[ Info: inner j3 i3 - 3
[ Info: inner j1 i2 - 2
[ Info: inner j2 i2 - 2
[ Info: inner j3 i2 - 2
[ Info: inner j1 i1 - 1
[ Info: inner j2 i1 - 1
[ Info: inner j3 i1 - 1
```

julia +1.12 -t6
```
julia> Threads.@threads for i in 1:3
           @info "outer i$i - $(Threads.threadid())"
           Threads.@threads for j in 1:3
               @info "inner j$j i$i - $(Threads.threadid())"
           end
       end
[ Info: outer i3 - 4
[ Info: outer i1 - 6
[ Info: outer i2 - 5
[ Info: inner j1 i1 - 2
[ Info: inner j3 i3 - 6
[ Info: inner j2 i1 - 5
[ Info: inner j1 i2 - 3
[ Info: inner j2 i3 - 2
[ Info: inner j1 i3 - 4
[ Info: inner j3 i1 - 4
[ Info: inner j2 i2 - 6
[ Info: inner j3 i2 - 7
```

## Problem

The inner loop in `check_dynamic_linkage` runs in parallel:
```julia
Threads.@threads for libname in collect(keys(libs))
```
Each iteration may call patchelf (via update_linkage, relink_to_rpath, or ensure_soname) on the same binary file (path(oh)). Without synchronization, concurrent patchelf writes corrupt the file.

## Solution
Add per-file locking via with_patchelf_lock(path) to serialize patchelf operations on the same binary
Capture subprocess output and include it in error messages when commands fail

### DIFF
--- a/src/auditor/dynamic_linkage.jl
+++ b/src/auditor/dynamic_linkage.jl
@@ -525,11 +525,11 @@ function _update_linkage_elf(prefix::Prefix, platform::AbstractPlatform, path::A
         rpaths = unique(vcat(current_rpaths, rp))
         # I don't like strings ending in '/.', like '$ORIGIN/.'.  I don't think
         # it semantically makes a difference, but why not be correct AND beautiful?
-        chomp_slashdot = p -> begin
-            if length(p) > 2 && p[end-1:end] == "/."
-                return p[1:end-2]
+        chomp_slashdot = path -> begin
+            if length(path) > 2 && path[end-1:end] == "/."
+                return path[1:end-2]
             end
-            return p
+            return path
         end
         rpaths = chomp_slashdot.(rpaths)
         # Remove paths starting with `/workspace`: they will not work outside of the


src/auditor/soname_matching.jl
```diff
@@ -47,6 +47,19 @@ function ensure_soname(prefix::Prefix, path::AbstractString, platform::AbstractP
...
+    end
+
+    # For Linux/FreeBSD, wrap the entire read-modify-verify cycle in the file lock
+    return with_patchelf_lock(path) do
+        _ensure_soname_elf(prefix, path, platform; verbose, autofix, subdir)
+    end
+end
+
+function _ensure_soname_macho(prefix::Prefix, path::AbstractString, platform::AbstractPlatform;
+                              verbose::Bool = false, autofix::Bool = false, subdir::AbstractString="")
```
giordano: The fact that this function takes `autofix` as argument is code smell: we shouldn't call this function to start with if `autofix`is false.

src/auditor/soname_matching.jl
```diff
@@ -47,6 +47,19 @@ function ensure_soname(prefix::Prefix, path::AbstractString, platform::AbstractP
         return true
     end
 
+    # macOS uses install_name_tool (check first since Sys.isbsd is true for macOS too)
```
giordano: Fun, also weird comments are duplicated

src/auditor/dynamic_linkage.jl
```diff
@@ -448,57 +450,32 @@ function update_linkage(prefix::Prefix, platform::AbstractPlatform, path::Abstra
         return
     end
 
+    # macOS uses install_name_tool (check first since Sys.isbsd is true for macOS too)
```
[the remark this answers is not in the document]
IanButterworth: Claude initially didn't realize `isbsd` covered macos. This is just a correction comment that I should've dropped

PR #1412, opened 2025-11-29
[CI] Run tests on GitHub Actions

giordano:
The general notion is that the tests of this package would take a long time without persistent caching, and would also require a large amount of storage available.  Let's test how bad things are nowadays.

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -1,0 +1,136 @@
+name: Tests
+
+on:
+  push:
+    branches: "master"
+    tags: ["*"]
+    paths:
+      - '.github/workflows/Tests.yml'
+      - '*.toml'
+      - 'src/**'
+      - 'test/**'
+  pull_request:
+    paths:
+      - '.github/workflows/Tests.yml'
+      - '*.toml'
+      - 'src/**'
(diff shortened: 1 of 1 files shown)


giordano:
```
  Unable to automatically download/install artifact 'GCCBootstrap-x86_64-linux-musl.v9.1.0.x86_64-linux-musl.unpacked' from sources listed in '/home/runner/.julia/packages/BinaryBuilderBase/yt9V4/Artifacts.toml'.
  Sources attempted:
  - https://pkg.julialang.org/artifact/13a1bcd3a29ff96c26935ddcea9aeb149ad49451
      Error: AssertionError: size == padded_size == 0
  This error may be a symptom of insufficient disk space on the device the tarball is being written to.
  
  - https://github.com/JuliaPackaging/Yggdrasil/releases/download/GCCBootstrap-v9.1.0+3/GCCBootstrap-x86_64-linux-musl.v9.1.0.x86_64-linux-musl.unpacked.tar.gz
      Error: AssertionError: size == padded_size == 0
  This error may be a symptom of insufficient disk space on the device the tarball is being written to.
```
despite having 49 GB available at the beginning, sigh.

PR #1420, opened 2026-03-13
Document how to use local dependencies for testing

eschnett:


giordano (commented):
Looks overall ok, but it would be better to merge this with https://docs.binarybuilder.org/stable/FAQ/#Can-I-publish-a-JLL-package-locally-without-going-through-Yggdrasil?-1 (staying in this page sounds good to me though, and keep the FAQ leaner)

docs/src/building.md
```diff
@@ -475,9 +475,61 @@ You can also provide the target triplet `Base.BinaryPlatforms.host_triplet()`
...
+It is possible to use a locally-deployed JLL package (built with
+`--deploy=local`) as dependency for another JLL package that is also
+built locally. This is useful for testing dependencies, e.g. to check
+whether a change to one package `A_jll` will break a dependency
+`B_jll`.
+
+To build JLL `A` and use it as dependency when building JLL `B`, do
+the following:
+
+1. Build JLL `A` with the option `--deploy-local`. This will (as
```
lgoettgens, line 497:
```suggestion
1. Build JLL `A` with the option `--deploy=local`. This will (as
```

PR #1440, opened 2026-07-08
Get tests passing / CI green again

maleadt:
Locally, I was seeing a bunch of failures that seemingly didn't happen on CI. This PR try to hardens against those failures in order to get CI to pass locally.

On CI there was a hash mismatch, fixed here.

Also bump to 1.12, which is what Yggdrasil uses. The rootfs doesn't have it though, so downloading the release for now.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -43,37 +43,58 @@ jobs:
       set -e
       $(JULIA) -e 'using Pkg; Pkg.Registry.update(); Pkg.instantiate()'
       $(JULIA) -e 'using BinaryBuilder; BinaryBuilder.versioninfo()'
-      $(JULIA) -e 'using Pkg; Pkg.status(; mode=PKGMODE_MANIFEST)'
+      $(JULIA) -e 'using Pkg; Pkg.status(; mode=Pkg.PKGMODE_MANIFEST)'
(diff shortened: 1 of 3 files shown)


maleadt:
```
/etc/profile.d/0_bb_utils.sh: line 35: set: write error: No space left on device
/etc/profile.d/0_bb_utils.sh: line 41: echo: write error: No space left on device
AnyPlatform: Test Failed at /data/staticfloat/bb_azp/rootfs/agent_3/_work/1/s/test/building.jl:460
  Expression: autobuild(build_path, "libfoo", v"1.0.0", [DirectorySource(build_tests_dir)], libfoo_cmake_script, [AnyPlatform()], libfoo_products, Dependency[])
    Expected: CompositeException
      Thrown: SystemError
      SystemError: close: No space left on device
```

giordano:
Can we clear up the depot once and the old installation of Julia v1.7, if it's still there?  I see there's a call to `Pkg.gc()`, but that's probably not doing much right now.  This is kinda the problem of the tests in this repo: they're very storage-hungry, and redownloading all artifacts every time would take a long time

giordano:
For Cargo, we may need to implement https://github.com/JuliaPackaging/Yggdrasil/pull/10590#issuecomment-2683448300

PR #1446, opened 2026-08-21
Speed up JLL registration

maleadt:
Avoid recomputing build metadata when registering large JLL packages.

The main improvement carries metadata already computed by `autobuild`—the tarball SHA-256, artifact tree hash, and product locations—in a versioned sidecar. `rebuild_jll_package` verifies the tarball digest and metadata shape before using it, and falls back to inspecting the tarball when metadata is missing or invalid.

The fallback path is also faster:

- index tarballs by canonical platform triplet instead of repeatedly scanning the download directory;
- inspect independent platform tarballs concurrently, with bounded concurrency;
- use the registries already installed in the depot to determine the registration PR title instead of cloning the full General history.

On 18 Clang_unified tarballs, the JLL reconstruction phase changed from:

- master: 59.7 seconds
- parallel inspection: 13.5 seconds
- build metadata reuse: 3.8 seconds

This work prompted by an excruciatingly slow registration of Clang_unified_jll on Yggdrasil.

Assisted by: Claude Fable 5, GPT 5.6 Sol


test/basic.jl
```diff
@@ -256,3 +260,111 @@ end
     dict = build_project_dict("Clang", v"9.0.1+2", dependencies)
     @test dict["compat"]["libLLVM_jll"] == "8.3"
 end
+
+@testset "Tarball lookup" begin
+    # `main_tarball_platform` itself is covered in jll.jl; this is about the table
+    using BinaryBuilder: tarball_lookup_table, tarball_platform_key
```
giordano: Not really a blocker, I just find slightly unusual seeing `using`s inside testset.

src/AutoBuild.jl
```diff
@@ -1219,107 +1213,302 @@ function rebuild_jll_package(obj::Dict;
...
+"""
+    build_meta_path(tarball_path) -> String
+
+Path of the `.meta.json` sidecar describing `tarball_path`.
+
+The name does not match `*.tar.*`, so artifact uploaders do not treat it as a release
+asset.
+"""
+function build_meta_path(tarball_path::AbstractString)
+    return string(replace(tarball_path, r"\.tar\.[A-Za-z0-9]+$" => ""), ".meta.json")
```
giordano: I think this is equivalent to
```suggestion
    return replace(tarball_path, r"\.tar\.[A-Za-z0-9]+$" => ".meta.json")
```
?

src/AutoBuild.jl
```diff
@@ -1079,6 +1068,9 @@ function autobuild(dir::AbstractString,
             products_info,
         )
 
+        # Let a separate registration job reuse the metadata computed while packaging.
+        write_build_meta(tarball_path, tarball_hash, git_hash, products_info)
```
giordano, line 1072: Is the idea that we upload the metadata files as artifacts, and then download them during registration, so that we can reuse the information?

maleadt: Yes, as Buildkite artifacts. Do you want me to put up the Yggdrasil PR already? I was going to wait until this is merged, but it may be easier to review seeing both ends.

giordano: No, I think it's fine to wait for this to be merged, I just wanted to understand the high-level flow.

giordano (approved):

PR #1447, opened 2026-08-26
Fail loudly when the JLL push or registry update did not take effect

maleadt:
LibGit2.push does not report reference updates the server refused or failed to apply: the objects are uploaded, the branch stays put, and no error is raised. Registration then built a release and a registry PR on top of the previous wrappers (aws_c_common_jll 1.0.0, 2026-08-26).

- push_jll_package fetches the remote main after pushing and errors if it is not at the new commit; it now returns that commit.
- upload_to_github_releases takes a commit and passes it to ghr as -c, so a freshly created release tag targets the wrappers built for it rather than whatever the default branch currently points at.
- Pkg only logs a failed registry update; update_registry_checked turns that into an error, since a stale registry silently yields wrong wrapper build numbers and registration PR titles.

### DIFF
--- a/src/AutoBuild.jl
+++ b/src/AutoBuild.jl
@@ -406,6 +406,7 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
     end
 
     products_dir = joinpath(pwd(), "products")
+    jll_commit = nothing
     if deploy_jll
         if verbose
             @info("Committing and pushing $(namejll(src_name)).jl wrapper code version $(build_version)...")
@@ -434,7 +435,7 @@ function build_tarballs(ARGS, src_name, src_version, sources, script,
                                 julia_compat, extra_kwargs...)
         end
         if deploy_jll_repo != "local"
-            push_jll_package(src_name, build_version; code_dir=code_dir, deploy_repo=deploy_jll_repo)
+            jll_commit = push_jll_package(src_name, build_version; code_dir=code_dir, deploy_repo=deploy_jll_repo)
         end
         if register
             if verbose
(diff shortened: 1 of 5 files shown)


src/AutoBuild.jl
```diff
@@ -528,16 +530,25 @@ function get_compilers_versions(; compilers = [:c])
...
+    commitish = commit === nothing ? [] : ["-c", string(commit)]
     for attempt in 1:attempts
         try
             # Note: in some cases we may want to reduce/avoid concurrency to avoid exceeding
             # secondary rate limits:
             # https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#avoid-concurrent-requests
             # https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits
             concurrency = get(ENV, "BINARYBUILDER_GHR_CONCURRENCY", string(Sys.CPU_THREADS))
-            run(`$(ghr()) -u $(dirname(repo)) -r $(basename(repo)) -t $(gh_auth.token) -p $(concurrency) $(tag) $(path)`)
+            run(`$(ghr()) -u $(dirname(repo)) -r $(basename(repo)) -t $(gh_auth.token) -p $(concurrency) $(commitish) $(tag) $(path)`)
```
giordano, line 551: I presume this would fail if the commit isn't there when creating the tag?

giordano (approved):

src/AutoBuild.jl
maleadt: Yes, that's the reasoning.

PR #1448, opened 2026-08-27
Never set `maintainer_can_modify=false`

giordano:


src/wizard/github.jl
```diff
@@ -132,6 +132,10 @@ function create_or_update_pull_request(repo, params; auth=github_auth())
                 "base" => params["base"],
                 "head" => string(split(repo, "/")[1], ":", params["head"]),
             ))
+            # It's undocumented, but `maintainer_can_modify=true` appears to be always
+            # invalid in update PR requests, causing the error
+            #     Fork collab can only be enabled on cross-repo pull requests
+            params["maintainer_can_modify"] = false
```
Octogonapus: It's not documented but my read here is that `true` is valid only on cross-repo PRs. If this code is always used for same-repo PRs (within General) then this seems correct.

Octogonapus: Though I would err on the side of passing this in https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/76b5f0f18ee4ee47da12ceb2fe257d2a18488360/src/AutoBuild.jl#L699

Octogonapus: Official advice from github is actually to omit `maintainer_can_modify` entirely from same-repo PRs

PR #1450, opened 2026-09-03
Initialize JLL products in the order they are declared

maleadt:
The generated wrappers sorted products alphabetically, so a recipe could not
control the order in which its libraries are dlopened. That order matters when
a library depends on another product of the same package but has no rpath to
find it: the system loader then resolves the NEEDED entry itself, and may pick
up an unrelated copy from ld.so.conf. This is what happens with the CUDA 10.2
runtime on Jetson, where libcublas (sorted first) pulls in the JetPack
libcublasLt instead of the one from the artifact.

For most JLLs the order is irrelevant, because the audit rewrites the linkage
of every library to an $ORIGIN-relative rpath, so intra-package dependencies
are always found within the artifact. Only recipes that skip the audit and
ship unmodified vendor binaries are affected.

Products not in the recipe list (e.g. when rebuilding from old metadata) keep
the previous alphabetical order.

Test added in `test/jll.jl`; verified by hand that the CUDA_Runtime_jll 10.2 wrapper now comes out with `libcublasLt` before `libcublas` (as ordered by JuliaPackaging/Yggdrasil#14666, which was ineffective because of the sorting).


giordano:
I'm mildly sure that the problem is that when we go through the meta.json serialisation/deserialisation we lose the ordering information, I don't think this PR addresses that path.

maleadt:
I'm also considering the alternative of turning on the auditor for CUDA dependencies.

maleadt:
I checked the meta.json path: the ordering survives it. `get_meta_json` serializes `"products" => products` as a JSON array, `merge_json_objects` appends and `unique`s (both order-preserving), and `cleanup_merged_object!` reconstructs the products with a comprehension, so `obj["products"]` reaches `rebuild_jll_package` in recipe order, and this PR passes that vector through to `build_jll_package`. What does lose ordering is `products_info` (a `Dict{Product,Any}`, also the `"products"` mapping in the sidecar `build_meta`), but that is only used for lookup here, not for ordering.

Added a test that round-trips through `get_meta_json` → JSON → `merge_json_objects` → `cleanup_merged_object!` and checks the order.

maleadt:
> I'm mildly sure that the problem is that when we go through the meta.json serialisation/deserialisation we lose the ordering information, I don't think this PR addresses that path.

Dixit Claude:

> I checked the meta.json path: the ordering survives it. `get_meta_json` serializes `"products" => products` as a JSON array, `merge_json_objects` appends and `unique`s (both order-preserving), and `cleanup_merged_object!` reconstructs the products with a comprehension, so `obj["products"]` reaches `rebuild_jll_package` in recipe order, and this PR passes that vector through to `build_jll_package`.

giordano:
Ah, that may be true in JSON 1 which uses a custom datatype, but [before v1](https://juliaio.github.io/JSON.jl/stable/migrate/#Reading-JSON) the deserialisation was in a plain Dict, so the order was anyone's guess (reason why there were lots of sort calls to sort alphabetically)

maleadt:
But these are arrays, not dicts? I'm not too familiar with JSON though. In any case, I had Claude add a test on the order, so this should catch any issue.

giordano:
Probably some things changed over the last 6 years since #802 and #805, but the problem at the time was that the order of the products in the wrappers was randomly changing every time, because the information was in a dictionary in the JSON file and its order was undefined when we reading back from it, so the bandaid solution was to sort them after reading the JSON file.

But if the order now is deterministic, then all is good, I guess.

PR #1453, opened 2026-09-15
Keep the libuv event loop alive during builds

maleadt:
Julia 1.12 can lose the wakeup of the thread that services the libuv event loop: when a worker thread arms IO (Pkg downloads artifacts from a `Threads.@spawn` task) while thread 0 is parked, nobody runs uv_run again and the download stalls forever with data waiting in the socket. On Yggdrasil this shows up as jobs stuck at "Installing N artifacts" until the 5-hour timeout.

Arm a periodic Timer for the lifetime of build_tarballs when running with more than one thread. An active handle keeps the loop alive, so thread 0 blocks inside uv_run rather than parking and registers new watchers on its next iteration. Remove once the runtime bug is fixed.

Works around JuliaLang/julia#63173.


giordano:
_sigh_. But if it does the job I guess it's good.

maleadt:
I'm going to try and catch this live on `amdci7` first to see if I can debug some more; so don't merge this yet.

